PR #2, opened 2021-06-25
Datalog prototyping

aferr:
Check in some code that was used to prototype dataflow/IFC for Raksha


markww:
This hasn't been touched since July, closing to clean up PR view.

PR #3, opened 2021-06-29
Some exploration of taint analysis with Souffle

bgogul:
This is a WIP and may contain bugs. Sending it earlier to get feedback on my current thinking. Take a look at `rule.dl`, which defines the analysis.

### DIFF
--- a/examples/rules.dl
+++ b/examples/rules.dl
@@ -1,0 +1,48 @@
+// Is this symbol an access path to some data?
+.decl isAccessPath(x: symbol)
+// Is this symbol a label.
+.decl isLabel(x: symbol)
+// A data flow edge
+.decl edge(src: symbol, tgt: symbol)
+ // A data flow path
+.decl path(src: symbol, tgt: symbol)
+// Definitely has this label.
+.decl hasLabel(v: symbol, label: symbol)
+// Definitely has this label.
+.decl claimHasLabel(v: symbol, label: symbol)
+// Definitely does not have this label.
+// This is dual of `hasLabel`. We need a better name.
+.decl mayNotHaveLabel(v: symbol, label: symbol)
+// May have the label
+.decl mayHaveLabel(v: symbol, label: symbol)
+// Label l is preserved along the edge.
+.decl preservesClaim(src: symbol, tgt: symbol, label: symbol)
+
+// Some inference rules
+
+// Symbols used in edges are access paths
+isAccessPath(x) :- edge(x, _). 
+isAccessPath(y) :- edge(_, y). 
+
+// Symbols used in hasLabel or mayHaveLabel are labels
+isLabel(l) :- claimHasLabel(_, l). 
+isLabel(l) :- mayHaveLabel(_, l). 
+
+// Transitive paths
+path(from, to) :- edge(from, to).
+path(from, to) :- edge(from, intermediate), path(intermediate, to).
+
+mayHaveLabel(tgt, l) :- claimHasLabel(tgt, l).
+mayHaveLabel(tgt, l) :- edge(src, tgt), mayHaveLabel(src, l).
+
+claimHasLabel(tgt, l) :- edge(src, tgt), preservesClaim(src, tgt, l), claimHasLabel(src, l).
+   
+mayNotHaveLabel(tgt, l) :- isAccessPath(tgt), isLabel(l), !claimHasLabel(tgt, l).
+mayNotHaveLabel(tgt, l) :- edge(src, tgt), mayNotHaveLabel(src, l),  !claimHasLabel(tgt, l).
+
+hasLabel(x, l) :- mayHaveLabel(x, l), !mayNotHaveLabel(x, l).
+
+.output isAccessPath
+.output hasLabel
+.output mayNotHaveLabel
+.output mayHaveLabel
--- a/examples/combined.dl
+++ b/examples/combined.dl
@@ -1,0 +1,43 @@
+#include "rules.dl"
+
+// particle P1
+//   foo: writes Foo
+//   claim foo is tag1 and is tag2
+
+// Particle P2
+//   foo: writes Foo
+//   claim foo is tag2 and is tag3
+
+// particle P3
+//   bar: reads Foo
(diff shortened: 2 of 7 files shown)


examples/rules.dl
```diff
@@ -0,0 +1,48 @@
...
+.decl preservesClaim(src: symbol, tgt: symbol, label: symbol)
+
+// Some inference rules
+
+// Symbols used in edges are access paths
+isAccessPath(x) :- edge(x, _). 
+isAccessPath(y) :- edge(_, y). 
+
+// Symbols used in hasLabel or mayHaveLabel are labels
+isLabel(l) :- claimHasLabel(_, l).
```
markww, line 28: Bit of a nit, but can we expand "l" to "label" throughout? Mistook this for the number '1' and was somewhat confused for a second.

examples/rules.dl
```diff
@@ -0,0 +1,48 @@
+// Is this symbol an access path to some data?
+.decl isAccessPath(x: symbol)
+// Is this symbol a label.
+.decl isLabel(x: symbol)
+// A data flow edge
+.decl edge(src: symbol, tgt: symbol)
+ // A data flow path
+.decl path(src: symbol, tgt: symbol)
+// Definitely has this label.
+.decl hasLabel(v: symbol, label: symbol)
```
markww, line 10: From what I understand, it seems that we are working with two (maybe three) different concepts each currently represented with the type `symbol`:

1. Vertices (such as "P1.output", "Egress.input", etc)
2. Labels (such as "tag1", "tag2", etc)
3. (Possibly) Symbols (not to be confused with Soufle's primitive `symbol`) which is Vertex | Label

If the above is correct, we might save ourselves some confusion as this file grows if we make some base types:

```
.type Vertex <: symbol
.type Label <: symbol
.type Symbol = Vertex | Label
```

and use them instead of `symbol` in the appropriate places.

examples/rules.dl
```diff
@@ -0,0 +1,48 @@
...
+// Definitely has this label.
+.decl claimHasLabel(v: symbol, label: symbol)
+// Definitely does not have this label.
+// This is dual of `hasLabel`. We need a better name.
```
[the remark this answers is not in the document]
markww: If this is necessary (I'm guessing as a strong override to the weaker `mayHaveLabel`?), I'd like to propose `cannotHaveLabel` instead.

examples/rules.dl
```diff
@@ -0,0 +1,48 @@
...
+// Transitive paths
+path(from, to) :- edge(from, to).
+path(from, to) :- edge(from, intermediate), path(intermediate, to).
+
+mayHaveLabel(tgt, l) :- claimHasLabel(tgt, l).
+mayHaveLabel(tgt, l) :- edge(src, tgt), mayHaveLabel(src, l).
+
+claimHasLabel(tgt, l) :- edge(src, tgt), preservesClaim(src, tgt, l), claimHasLabel(src, l).
+   
+mayNotHaveLabel(tgt, l) :- isAccessPath(tgt), isLabel(l), !claimHasLabel(tgt, l).
```
markww, line 40: I am new to Datalog, but doesn't this line mean that "a target may not have a label if it participates in an edge and does not claim to have a label"? Wouldn't that basically mean that all vertices that do not claim to have a label may not have that label? I think perhaps this could be solved by replacing `!claimHasLabel(tgt, l)` with `!mayHaveLabel(tgt, l)`.

examples/rules.dl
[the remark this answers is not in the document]
bgogul: `cannotHaveLabel` seems like a good name instead of `mayNotHaveLabel`.

BTW, please read this rules with a grain of salt. I am not positive that the rules for `mayNotHaveLabel` and `hasLabel` are completely correct. Datalog has [stratification restrictions](https://stackoverflow.com/questions/12379775/datalog-stratification) when using negation. I was introducing some of these relations willy nilly to get souffle to do something.

examples/rules.dl
bgogul: Of course. I am about to send  a separate  PR that defines the base relations that we need for data flow analysis. I will keep this in mind.

examples/rules.dl
bgogul: See my comment about stratification restrictions. This was one of the work arounds I added. To reiterate, I am not positive that these rules are fully correct. (`mayHaveLabel` is correct though.)

examples/rules.dl
aferr: If we want to make this to map to IFC concepts, it might be useful to reserve the term "Label" for the sets and give another name to the elements of the sets.

aferr: As another option: I sometimes use "ell" as a reference to LaTeX's `\ell` for label variables. This way it's also distinct from the name of the type.

examples/rules.dl
```diff
@@ -0,0 +1,48 @@
...
+// Is this symbol a label.
+.decl isLabel(x: symbol)
+// A data flow edge
+.decl edge(src: symbol, tgt: symbol)
+ // A data flow path
+.decl path(src: symbol, tgt: symbol)
+// Definitely has this label.
+.decl hasLabel(v: symbol, label: symbol)
+// Definitely has this label.
+.decl claimHasLabel(v: symbol, label: symbol)
```
aferr, line 12: (Because you mentioned that this is a good time to figure out terminology)

I think here a "label" is an element of a set, but to be consistent with IFC terminology, the labels should be the sets and we need another term for the elements of a set. Essentially, the IFC "flowsTo" relation is "is subset or equal to" on the sets (e.g., `{Alice}` is a label that says the data has only info owned by `Alice` can flowTo a location labeled `{Alice, Bob}`). 

Some other terms we could use for the elements of the set could be "tag" ([Flume](https://pdos.csail.mit.edu/papers/flume-sosp07.pdf) used the term tag for these -- Flume also used a label model where labels are drawn from the lattice over sets with subsetEq as the order) or "labelElement", or maybe "dataIdentifier".

aferr (approved):

markww:
Hasn't been touched since July, closing.

PR #4, opened 2021-06-29
Add souffle as a third_party library.

bgogul:
Setting up souffle as a third_party repo so that it can be used in the project.


aferr (approved):
Looks good! One question: Conrad mentioned in another PR that we should have copyright info at the top of every file. Do we need that for these as well?

aferr:
Err, I see now that you wrote WIP in the title. Let me know if you want me to undo the approval and wait.

bgogul:
> Looks good! One question: Conrad mentioned in another PR that we should have copyright info at the top of every file. Do we need that for these as well?

Yes, we will need to add copyright info to all files. I added them right now. I also added an example exercising Souffle's C++ interface.

bgogul:
@markww just FYI.

PR #9, opened 2021-07-02
Add some basic datalog relations for taint tracking.

bgogul:
This is a good opportunity for us to establish the common vocabulary on which to build further. We will need to add some tests, but we don't have the necessary components to write tests yet.

### DIFF
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -1,0 +1,77 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+
+// We will use the access path abtraction to refer to data during analysis.
+// Consider the following data schema (not a valid syntax):
+//  schema A {
+//    a: Integer
+//    b: schema B {
+//      bf: String
+//    }
+//  }
+//
+// Suppose that data with schema A is stored in a database that is accessible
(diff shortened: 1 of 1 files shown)


markww (approved):
LGTM

src/analysis/souffle/taint.dl
```diff
@@ -0,0 +1,77 @@
...
+//
+// Suppose that data with schema A is stored in a database that is accessible
+// through a handle `h`. `h.a`, `h.b`, and `h.b.bf` are all examples of access
+// paths. For the time being, we will use a `symbol` to represent access paths.
+// However, in the long term, it would make sense to define an ADT for this.
+//
+.type AccessPath <: symbol
+
+// A symbol representing a taint (or information flow) label. These would be
+// strings of the form "secret", "public", "timestampInMs", etc. Just like
```
aferr, line 34: FYI with the label model we are using (based on the lattice over sets using subsetEq as the order) "public" (the most public label) should actually be the empty set and "secret" should be the universe (the set containing all possible elements).

(Here I'm just copypasting my comment from the other PR since perhaps this one is where it's more relevant)

I think here a "label" is an element of a set, but to be consistent with IFC terminology, the labels should be the sets and we need another term for the elements of a set. Essentially, the IFC "flowsTo" relation is "is subset or equal to" on the sets (e.g., {Alice} is a label that says the data has only info owned by Alice can flowTo a location labeled {Alice, Bob}).

Some other terms we could use for the elements of the set could be "tag" (Flume used the term tag for these -- Flume also used a label model where labels are drawn from the lattice over sets with subsetEq as the order) or "labelElement", or maybe "dataIdentifier".

src/analysis/souffle/taint.dl
```diff
@@ -0,0 +1,77 @@
+//-----------------------------------------------------------------------------
```
markww, line 1: Does this file need an #ifndef guard for inclusion?

src/analysis/souffle/taint.dl
bgogul: Changed this to Tag.

src/analysis/souffle/taint.dl
bgogul: That is a good question. However, I am not sure if Souffle has such guards. I guess we will find out soon enough. 😄

aferr: It does. [Here is an example](https://github.com/google-research/raksha/blob/cb01ac2ace4cf94148c0567a5a506d1321bc1845/datalog-prototyping/ifc_plus_auth_logic_prototype/labelChecking.dl)

PR #16, opened 2021-07-08
First draft of simplified Manifest fact structures.

markww:
This commit creates the Datalog data structures and relationships to be used in specifying input facts to the dataflow analysis. In many cases, these are simplified, mostly by omitting type information. The goal here is to be able to represent a simple manifest containing a few particle specs that only read and write handles, claim a label on some handle connection, and check it later.

This commit assumes that the manifest facts we provide to the dataflow analysis will have a structure that is matched closely to the original manifest protobuf. Eventually, we will be creating a tool that translates manifest protobufs to be analyzed into facts. Having a structure that matches the original protobuf as closely as possible in Datalog will allow this translation tool to be simple and insulate it from knowledge about the inner workings of the Datalog dataflow analysis.

This is in first-draft state; I am not certain that this even compiles at this point. Nonetheless, it is probably worth putting it up so others can see what I'm thinking at this point.

### DIFF
--- a/src/analysis/souffle/list_macros.dl
+++ b/src/analysis/souffle/list_macros.dl
@@ -1,0 +1,7 @@
+#ifndef LIST_MACROS_H__
+#define LIST_MACROS_H__
+
+#define LIST_TYPE(EltType) EltType##List
+#define DEFINE_LIST_TYPE(EltType) .type LIST_TYPE(EltType) [ item: EltType, next: LIST_TYPE(EltType) ]
+
+#endif // #ifndef LIST_MACROS_H__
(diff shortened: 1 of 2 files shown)


markww:
Gogul and I discussed this, we need to send a more abstracted version of this to Datalog. Closing this.

src/analysis/souffle/list_macros.dl
```diff
@@ -0,0 +1,7 @@
+#ifndef LIST_MACROS_H__
+#define LIST_MACROS_H__
+
+#define LIST_TYPE(EltType) EltType##List
+#define DEFINE_LIST_TYPE(EltType) .type LIST_TYPE(EltType) [ item: EltType, next: LIST_TYPE(EltType) ]
```
aferr, line 5: This is nice. I didn't know you could also do type definitions this way in Souffle.

markww: I'm not going to claim this is exactly right; I didn't compile it. But since Souffle runs the C Preprocessor, it should be the case that C Preprocessor-type tricks should work similarly to how they do in a C/C++ file, similar to how they work in .S (preprocessing-needed) assembly files.

PR #18, opened 2021-07-09
Make taint.dl more header-like, add a very simple tag checker analysis in Datalog

markww:
On the header aspect:

In particular, add a #ifndef guard and remove the .output directive. Although Souffle datalog doesn't seem to have a standard convention for this, we should consider creating our own convention for distinguishing between files that act like C header files and files that act like compilation units.

I think Datalog header files should be guarded by include guards and should contain no .input nor .output directives. The reason for this is that, as headers, they should be includable into multiple different forms of Datalog analysis which may want to report different things. The .input and .output directives tie a file to gathering particular types of evidence from the outside and reporting certain types of evidence, respectively.

Although it may require some editor configuration, I think it might be valuable to name these kinds of files .dlh and .dlc (for header and compilation unit). Want feedback on that though, so haven't made the filename change here yet.

On the simple dataflow analysis aspect: see the comment for that commit.


aferr:
Re: Splitting files into .dlh and .dlc

It sounds potentially appealing.

## Split 1
So is the idea also that the `.dlh` files will have `.decl`'s but not rules, and all the rules go in the `.dlc` files? If we do that split, does it work out such that if we want to use something defined in `foo.dlh` and `foo.dlc` we only have to include `foo.dlh` but not `foo.dlc`? I would have guessed that we'd need to still include both, but this way you can use different versions of `foo.dlc` for different cases while keeping `foo.dlh` for all of them.

## Split 2
Or is the split that both declarations and rules go into `foo.dlh` and then `foo.dlc` just has `.input` and `.output` directives? (I think this might be closer to what you were thinking). This seems a bit more useful than Split 1, but the naming convention makes me think it works like Split 1. Maybe we could take this idea but make it `foo.dl` instead of `foo.dlh` and `foo_io.dl` instead of `foo.dlc`? (I'm not too strongly opinionated about naming for this, just giving more input).

aferr (approved):
LGTM. I don't know much about the .arcs parts in the tests, though.

markww:
@aferr , I was thinking something more like Split 2 in your comment. I thought myself very clever to use a naming convention that would poke people to use their C/C++ instincts for reasoning about inclusion, but you have made me realize that there's a distinction that I hadn't accounted for. While the Datalog files use the C Preprocessor for inclusion, the *way* that the C Preprocessor is used is very different because Datalog does not perform linking and does not have bodyless declarations that can be matched with implementations in other compilation units. As such, the `.dlh` and `.dlc` convention is misleading in a manner that I did not anticipate.

In that light, I agree that a different naming convention would be appropriate. Maybe something like `.dli` for includable datalog and `.dl` for executable datalog script.

src/analysis/souffle/taint.dl
```diff
@@ -28,6 +28,10 @@
 // paths. For the time being, we will use a `symbol` to represent access paths.
 // However, in the long term, it would make sense to define an ADT for this.
 //
+
```
bgogul, line 31: @markww it might be worthwhile sending this out as separate PR and merge it in, while you work on the analysis.

bgogul (commented):
It might be worthwhile sending out the header-like change as a separate PR, while you work on the tests.

markww:
I've split off the header change into #29 . @bgogul had some comments about the test approach that I'm going to address in a new branch, and am abandoning this one.

PR #19, opened 2021-07-10
Adds a souffle_cc_library rule.

bgogul:
This allows us to create a cc_library that corresponds to the c++ interface of a `.dl` file. See `cpp_interface.cc` for an example of how this can be used.


markww (approved):
LGTM

PR #29, opened 2021-07-14
Convert taint.dl to be more header-like.

markww:
In particular, add a #ifndef guard and remove the .output directive. Although Souffle datalog doesn't seem to have a standard convention for this, we should consider creating our own convention for distinguishing between files that act like C header files and files that act like compilation units.

I think Datalog header files should be guarded by include guards and should contain no .input nor .output directives. The reason for this is that, as headers, they should be includable into multiple different forms of Datalog analysis which may want to report different things. The .input and .output directives tie a file to gathering particular types of evidence from the outside and reporting certain types of evidence, respectively.

Although it may require some editor configuration, I think it might be valuable to name these kinds of files .dlh and .dlc (for header and compilation unit). Want feedback on that though, so haven't made the filename change here yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+bazel-bin
+bazel-out
+bazel-raksha
+bazel-testlogs
+**/*.swp
--- a/build_defs/BUILD
+++ b/build_defs/BUILD
@@ -1,1 +1,1 @@
-licenses("[notice]")
+licenses(["notice"])
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -28,6 +28,10 @@
 // paths. For the time being, we will use a `symbol` to represent access paths.
 // However, in the long term, it would make sense to define an ADT for this.
 //
+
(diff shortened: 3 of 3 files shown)


markww:
If this looks familiar, it is because it is the header change from PR #18 .

bgogul (approved):
> I think it might be valuable to name these kinds of files .dlh and .dlc (for header and compilation unit).

Let us not do the renaming now. As we build more souffle code and gain experience we can  consider the name change.

markww:
> > I think it might be valuable to name these kinds of files .dlh and .dlc (for header and compilation unit).
> 
> Let us not do the renaming now. As we build more souffle code and gain experience we can consider the name change.

That is fine with me. On this point, @aferr pointed out that the names `.dlh` and `.dlc`, in particular, imply a C-like pattern of organization, which we can't actually match faithfully due to the lack of a linker. I think, if we were to adopt a convention, I'd instead recommend. `.dli` to represent includable Datalog files.

But yes, let's see what our actual experience is before deciding on such a convention.

PR #35, opened 2021-07-20
Turn off warnings when compiling Souffle-generated C++.

markww:
Prior to this commit, the souffle_cc_library macro would generate
a number of warnings when compiling the C++ file generated by Souffle.
These warnings are uninteresting to us, as we aren't going to change
non-critical issues in Souffle generated code. Having these warnings in
the build log could obscure real issues. Use -w to suppress them
instead.


aferr (approved):
lgtm

PR #36, opened 2021-07-20
Switch testPasses -> testFails; allow multiple testFails rules.

markww:
Initially, I created tests using a relation called testPasses for two
reasons:

1. To allow tests to state what they want rather than the opposite of
   what they want.
2. To have a natural check that we included a test predicate at all.

@bgogul pointed out that we might want to add multiple rules stating the
requirements of the test and fail if any one of those rules failed. This
was not possible to do with testPasses. This commit switches to using
testFails, which naturally allows for expressing multiple requirements
in multiple rules.

This, however, opens us up to the possibility that someone might forget
to put a testFails at the bottom of a test entirely. I could not find
a good way to fix this in datalog, so as a very crude guardrail, I added a
shell test that just greps each file for testFails.

### DIFF
--- a/src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver.cc
+++ b/src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver.cc
@@ -17,7 +17,7 @@
 // This file is a simple driver to run tests expressed as a datalog script that
 // contains facts describing the dataflow graph (such as edges and tag claims)
 // and our expectations (such as which tags will be present or absent). It
-// returns 0 if the test has contents in the testPasses relation, 1 otherwise.
+// returns 1 if the test had contents in the testFails relation, 0 otherwise.
 // This allows the datalog tests to be run from a bazel cc_test rule and avoids
 // managing external facts files or diffing against an expected output file.
 //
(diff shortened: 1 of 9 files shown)


src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver.cc
```diff
@@ -26,28 +26,44 @@
...
+  souffle::Relation *test_failures = prog->getRelation("testFails");
+
+  assert(test_failures != nullptr);
 
-  int const test_passes_size = test_passes->size();
-  assert((test_passes_size == 0) || (test_passes_size == 1));
-  std::cout << "Test " << test_name;
-  if (test_passes_size == 1) {
-    std::cout << " passes." << std::endl;
+  bool const test_has_failures = test_failures->size() > 0;
```
bgogul, line 41: I might be missing some detail, but why didn't the following work? 

```
  isTest(x) :- testPasses(x).
  testFails(x) :- isTest(x), !testPasses(x).
```
In fact_test_driver.cc, check the following:
  - `isTest` (or `testPasses`) is not empty, which ensures that there was at least one `testPasses` rule.
  - `testFails` is empty, which ensures that none of the tests fail.

markww: This is the test that made me realize that that approach wouldn't work:

```
// Make sure that a test where one testPasses predicate passes while the other fails fails.
#include "taint.dl"
#include "fact_test_helper.dl"

testPasses("always_pass_rule") :- 1 = 1.
testPasses("always_fail_rule") :- 1 = 2.
```

With the above rules, this always passes. We would hope that the code you pasted would make the two statements of `testPasses` act like they were `AND`ed together rather than `OR`ed. But here is what happens:

1.  The first rule, which always passes, populates `testPasses` and thus also `isTest`.
2. The second rule, which always fails, fails to populate `testPasses` and thus also does not populate `isTest`. As such, because it is not in `isTest`, it cannot populate `testFails`.

Note that `x` can be in `isTest` only if `x` is in `testPasses`. That means that we can substitute `testPasses` for `isTest` everywhere it occurs, making the recipe for `testFails` the following:

`testFails(x) :- testPasses(x), !testPasses(x).`

This is a contradiction, and thus `testFails` is guaranteed to be empty. In that circumstance, the test driver will only fail if `isTest` was empty, which would occur if no `testPasses` rule were satisfied.

I get messed up by the Datalog syntax all the time; my C/C++ instincts tell me that `Relation(x)` means "call Relation with x" when in fact it means "x is a member of Relation". This is *especially* bad when a relation has "is" in the title. The relation `isTest` is misleadingly named; it is not a predicate function that returns `true` if `x` is a test and `false` otherwise; it is a set which `x` may or not be a member of. I propose that we avoid using `is`-prefixed relation names to prevent laying this trap for ourselves in the future.

bgogul: Ah, makes sense now. Will take a look at the PR in detail in a bit.

markww: It occurs to me that we have a desire for a feature that Datalog doesn't really have. When you provide a rule, Datalog either populates the relation in the rule or it does not populate it. What we really want is a rule which specifies two relations: a rule and its complement. Imagine that we used a bit of C macro magic to define `TEST_RULE` as follows:

```
#define TEST_RULE(test_aspect, recipe) \
  testPasses(test_aspect) :- recipe . \
  testFails(test_aspect) :- !(recipe)
```

 to rewrite the test as follows:

```
TEST_RULE("always_pass_rule", 1 = 1).
TEST_RULE("always_fail_rule", 1 = 2).
```

Now the first rule will end up in `testPasses` and the second rule will end up in `testFails`. We could even simplify it further to make it as follows:

Now we can define `allTests` (formerly `isTest`) as follows:

```
allTests(testAspect) :- testPasses(testAspect).
allTests(testAspect) :- testFails(testAspect).
```

And it would be guaranteed to capture all test aspects. Now we can have the driver assert that `test_passes->size() == all_tests->size() && (all_tests->size() > 0)`

I think the above could be the best possible solution, but it would require the ugliness of macros. Let me know what you think.

markww: I went ahead and put this idea into PR #37 , an alternative approach to this PR.

markww:
This is no longer necessary now that PR #37 is in.

PR #37, opened 2021-07-21
Rewrite the test infrastructure to rely on a new TEST_RULE macro.

markww:
Add a new TEST_RULE macro that adds an aspect to testPasses if the
attached condition succeeds, testFails otherwise. This allows us to
place a test into one of the two buckets after it evaluates instead of
being omitted entirely under some condition.

### DIFF
--- a/src/analysis/souffle/fact_test_helper.dl
+++ b/src/analysis/souffle/fact_test_helper.dl
@@ -5,9 +5,25 @@
 
 .type TestAspectName <: symbol
 
+// Any tests in this rule have passed. Should usually be populated by TEST_RULE rather than
+// directly used.
+.decl testPasses(testAspectName: TestAspectName)
+
 // If a the testFails relation has any contents, then the test has failed. If not, it has passed.
-// Tests should create rules indicating what should not be true of the test for it to pass.
+// Tests should create rules indicating what should not be true of the test for it to pass. Should
+// usually be populated by TEST_RULE rather than directly used.
 .decl testFails(testAspectName: TestAspectName)
 .output testFails(IO=stdout, delimiter=",")
+
+.decl allTests(testAspectName: TestAspectName)
+.output allTests(IO=stdout, delimiter=",")
+
+allTests(testAspectName) :- testPasses(testAspectName).
+allTests(testAspectName) :- testFails(testAspectName).
+
+#define TEST_RULE(test_aspect_name, rule) \
+  testPasses(test_aspect_name) :- rule. \
+  testFails(test_aspect_name) :- !(rule)
+
 
 #endif // SRC_ANALYSIS_SOUFFLE_ANALYZE_TAG_CHECKS_TEST_HELPER_DL_
--- a/src/analysis/souffle/fact_test_helper.dl
+++ b/src/analysis/souffle/fact_test_helper.dl
@@ -5,12 +5,12 @@
 
 .type TestAspectName <: symbol
 
-// Any tests in this rule have passed. Should usually be populated by TEST_RULE rather than
+// Any tests in this rule have passed. Should usually be populated by TEST_CASE rather than
 // directly used. If it is directly used, the test must be "declared" using an allTests fact on the
 // aspect.
 .decl testPasses(testAspectName: TestAspectName)
 
-// All test aspects tested by a particular test. Should usually be populated by TEST_RULE rather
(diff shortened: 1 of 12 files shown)


src/analysis/souffle/fact_test_helper.dl
```diff
@@ -3,15 +3,27 @@
...
+.decl allTests(testAspectName: TestAspectName)
+.output allTests(IO=stdout, delimiter=",")
+
+allTests(testAspectName) :- testPasses(testAspectName).
+allTests(testAspectName) :- testFails(testAspectName).
+
+#define TEST_RULE(test_aspect_name, rule) \
+  testPasses(test_aspect_name) :- rule. \
+  testFails(test_aspect_name) :- !(rule)
+
```
bgogul, line 27: How about tweaking the macro a little bit. 

```suggestion
#define TEST_RULE(test_aspect_name, rule) \
  isTest(test_aspsect_name). \
  testPasses(test_aspect_name)
```

Then you can use the macro as follows:
```
TEST_RULE("should_not_be_tagged_with_trusted") :-  !mayHaveTag("R.P2.bar", "trusted").
```

instead of 
```
TEST_RULE("should_not_be_tagged_with_trusted", !mayHaveTag("R.P2.bar", "trusted")).
```

Now, `isTest` is to true for all the declared tests. Then we can go add the following rule:
```
testFails(x) :- isTest(x), !testPasses(x).
```
Further, we can simply do the following in `fact_rule_checker` to detect test failures:
  - `isTest` is not empty.
  - `testFails` is empty.

bgogul (commented):
Macros seems like a reasonable compromise for tests. I left a suggestion that would make this macro look like an inference rule. PTAL and let me know.

src/analysis/souffle/fact_test_helper.dl
markww: This is excellent, I like it a lot. This also allows `TEST_RULE` rules to use `,` as `AND`. Going to implement this now.

markww: Done, PTAL.

src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver.cc
```diff
@@ -26,28 +26,54 @@
...
+    std::cout
+      << "Test "
+      << test_name
+      << " does not have any test conditions."
+      << std::endl;
+    prog->printAll();
+    return 1;
+  } else if (test_has_failures) {
+    std::cout << "Test " << test_name << " failed." << std::endl;
+    prog->printAll();
```
bgogul, line 55: Does this print the entire program? It would be more useful to just print the failing test names. We can do it in a separate PR.

bgogul (approved):

src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver.cc
markww: It doesn't print the entire program, it prints all of the output rules in the program. At this time, this includes only the test names and the failing tests, both of which are important information.

src/analysis/souffle/fact_test_helper.dl
```diff
@@ -3,15 +3,36 @@
...
+// TEST_RULE is constructed so that it can take the place of a rule head. It "declares" a test
+// aspect via the allTests fact and sets up a testPasses head for the aspect in question. Example
+// usage:
+//
+// TEST_RULE("my_test_aspect") :- MyRelation(x), !MyOtherRelation(x).
+//
+// All recipes with a TEST_RULE head shall be seen as a condition that must be met for the test to
+// pass.
+#define TEST_RULE(test_aspect_name) \
```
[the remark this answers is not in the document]
markww: Good call, I will do that.

src/analysis/souffle/fact_test_helper.dl
```diff
@@ -3,15 +3,36 @@
...
+// aspect.
+.decl testPasses(testAspectName: TestAspectName)
+
+// All test aspects tested by a particular test. Should usually be populated by TEST_RULE rather
+// than directly used.
+.decl allTests(testAspectName: TestAspectName)
+.output allTests(IO=stdout, delimiter=",")
+
+// If a the testFails relation has any contents, then the test has failed. Should not be populated
+// directly.
```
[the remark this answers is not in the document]
markww: Good catch, I'll get this.

PR #39, opened 2021-07-21
Suppress warnings when building Souffle.

markww:
In a previous commit, I added options to suppress warnings while
building files generated by Souffle. This commit further suppresses
warnings in code that we will not directly fix by suppressing warnings
when building Souffle itself.

### DIFF
--- a/third_party/souffle/BUILD
+++ b/third_party/souffle/BUILD
@@ -35,7 +35,7 @@ genrule(
         "src/parser/parser.hh",
         "src/parser/stack.hh",
     ],
-    cmd = "M4=$(M4) $(BISON) -d --output=$(RULEDIR)/src/parser/parser.cc $<",
+    cmd = "M4=$(M4) $(BISON) -d --warnings=none --output=$(RULEDIR)/src/parser/parser.cc $<",
     toolchains = [
         "@rules_bison//bison:current_bison_toolchain",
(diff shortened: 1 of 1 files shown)


bgogul (approved):

third_party/souffle/BUILD
```diff
@@ -93,7 +96,12 @@ cc_library(
 cc_binary(
     name = "souffle",
     srcs = ["src/main.cpp"],
-    copts = ["-std=c++17"],
+    copts = [
+        "-std=c++17",
+        # We don't care about non-critical issues in the third-party Souffle
```
bgogul, line 101: Just a nit: we do care as these warnings may be indicative of some underlying problem. However, I see your point about these wall of warnings obfuscating other warnings raised in our code.

bgogul (approved):

third_party/souffle/BUILD
markww: OK. Should I make the comment a little more nuanced in a follow-on PR? Something like "We don't want warnings in 3rd party packages to obscure issues in our own code that we are more likely to need to fix."?

bgogul: Not a big deal, but your suggestion sounds better.

markww: Softened the language as suggested.

PR #40, opened 2021-07-21
Rename trusted tags

markww:
The two tests that we have ported over from Arcs, ok_claim_propagates
and fail_different_tag, have tag names that imply integrity-style
behavior (such as trusted) when their behavior in the tests is more
consistent with confidentiality-style behavior. This change changes
those tests to use names that imply confidentiality behavior instead.

This change is based off of PR #37 and should be submitted after it.

PR #41, opened 2021-07-22
Rename trusted tags2

markww:


markww:
This is basically just a copy of PR #40 to avoid an ugly merge.

PR #42, opened 2021-07-22
Port the fail_check_on_subpaths Arcs test as a facts test.

markww:


src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
```diff
@@ -0,0 +1,34 @@
...
+// propagate like confidentiality tags.
+// #Ingress: P1
+// #FAIL: hc:P2.foo is userSelection
+//particle P1
+//  foo: writes Foo {a: Text, b: Number}
+//  claim foo.a is userSelection
+//  claim foo.b is screenContent
+//particle P2
+//  foo: reads Foo {a: Text, b: Number}
+//  check foo is userSelection
```
bgogul, line 12: This is the test case where the tags for `foo` are computed from the tags for the fields `foo.a` and `foo.b`. In our case, foo should have the confidentiality tags `{userSelection, screenContent}`. 

The check is trying to establish that `foo` *definitely* has the label `userSelection`.  Since `foo.b` does not have the `userSelection` tag, this check is supposed to fail.  Without a `definitelyHasTag` relation, we cannot validate this check, but we can verify that the correct tags are inferred.

src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
```diff
@@ -0,0 +1,34 @@
...
+//recipe R
+//  P1
+//    foo: writes h
+//  P2
+//    foo: reads h
+#include "taint.dl"
+#include "fact_test_helper.dl"
+
+claimHasTag("R.P1.foo.a", "userSelection").
```
bgogul, line 21: Both of these should induce a new relation `claimHasTag("R.P1.foo", "userSelection")`. We should add a new inference rule to taint.dl.

bgogul (changes requested):

src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
markww: Should they claim `R.P1.foo` has userSelection, or should they introduce an edge from the member to the parent like so:

```
edge("R.P1.foo.a", "R.P1.foo").
edge("R.P1.foo.b", "R.P1.foo").
```

If I understand correctly what you're saying here, you're saying that the privacy tags should propagate up to the parent because tags on members should propagate to their containing types?

Or, when you say "add a new inference rule", are you saying that we should have something that derives that automatically like so:

```
edge(member, base) :- isAccessPath(member), isAccessPath(base), contains(cat(base, "."), member).
```
?

src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
markww: I see, so I'll add the checks you mentioned for the privacy aspect now and add a comment at the top saying that we need to add the integrity aspect later.

src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
bgogul: Yes, that is what I mean. I was thinking about updating the `mayHaveTag` relation, but your approach of updating the `edge` relation is the correct one.

bgogul: >  Both of these should induce a new relation claimHasTag("R.P1.foo", "userSelection").

BTW, I meant `mayHaveTag` instead of `claimHasTag`.

markww: I'm not sure the above is the right way to do this. Consider the following case:

```
particle P1
  foo: writes Foo { a: Text }
  claim foo.a is userSelection
particle P2
  foo: reads Foo { b: Number }
  check foo.b is not userSelection
recipe R
  P1
    foo: writes h
  P2
    foo: reads h
```

Now, in the above, we have clear information that particle `P1` performed some operation that put `userSelection` on `Foo.a`. That is written into the handle `h` which is read by `P2`, which reads `Foo.b`. While both of the particles speak of `Foo`, they each speak of disjoint fields of `Foo`.

If we draw an edge from members of `Foo` to the base `Foo`, then the `userSelection` taint will spread to `Foo.b`, which seems wrong, as `P1` did not write `Foo.b`.

I wonder if we should draw edges to base objects at all, or if we should instead consider only the leaf fields when drawing edges.

bgogul: Indeed that will make the analysis quite imprecise. I just remembered that, in Arcs, we only consider data flow edges between leaf fields. (One caveat is that we don't support recursive data structures). However, we do support queries on parents, which is when we combine the tags on the leaves.  See [AccessPathLabels.kt](https://github.com/PolymerLabs/arcs/blob/e7a79c2c885bd090f45c8fb9f8898ba8367f4907/java/arcs/core/analysis/AccessPathLabels.kt#L79).

That is probably why I was suggesting new inference rules for the `mayHaveTag` relation.

bgogul: Sorry for putting you off the track!

markww: No prob. Ok, that makes sense. Here's an alternative approach that might be a bit more elegant in graph representation but would have the semantics you describe: Imagine that each `check` is given its own node in the dataflow graph. So, using the example I gave above, imagine someone had the following in their Arcs manifest:

```
check foo.b is not userSelection
```

Now, imagine this is expressed as a `checkDoesNotHaveTag` fact as follows:

```
.decl checkDoesNotHaveTag(accessPath: AccessPath, tag: Tag, checkNum: number)

checkDoesNotHaveTag("R.P2.foo.b", "userSelection", $).
```

We would provide the following new rules for creating a new node in the graph specifically for accumulating tags for the check:

```
// Convenience macro to prevent things from getting too confusing.
#DEFINE CHECK_NODE_PATH(access_path, check_num) cat(cat(cat("check", to_string(checkNum)), "."), accessPath)

// Always draw an edge from the exact access path being checked to the check node.
edge(accessPath, CHECK_NODE_PATH(accessPath, checkNum)) :-
  checkDoesNotHaveTag(accessPath, _, checkNum).

// Draw an edge from all subfields of the checked accessPath to the check node.
edge(memberAccessPath, CHECK_NODE_PATH(accessPath, checkNum)) :-
  checkDoesNotHaveTag(accessPath, _, checkNum), contains(cat(accessPath, "."), memberAccessPath).

// An example test case you might put on the check node to fulfill its duties
TEST_CASE("does not have userSelection") :-
  !mayHaveTag(CHECK_NODE_PATH(accessPath, checkNum), "userSelection").
```
 
In the case above, with `check foo.b is not userSelection`, the node will only have an edge drawn from `R.P2.foo.b` to the check node. However, if we were to `check foo`, we would draw an edge from all access paths prefixed with foo to the check node (which would also be exactly `foo.b`). Because this node is a dead end, however, these edges would not cause imprecise over-tainting for later checks.

bgogul: Hmm.. I am confused by the need for `checkDoesNotHaveTag`. I was thinking more like the following:
```
mayHaveTag(base, tag) :- 
   isAccessPath(member), isAccessPath(base), 
   contains(cat(base, "."), member), 
   mayHaveTag(member, tag).
```

Won't that work?

markww: You're right, that works just fine as long as we have the invariant that edges go only from leaf to leaf. My mind was still stuck thinking that we could propagate taint accumulated higher on the `accessPath` down the dataflow graph. I'll use your suggestion here.

src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
markww: Done.

src/analysis/souffle/tests/arcs_fact_tests/fail_check_on_subpaths.dl
markww: Done.

markww:
I think I've addressed all of your comments @bgogul , PTAL. Thanks!

src/analysis/souffle/taint.dl
```diff
@@ -78,4 +78,13 @@ path(from, to) :- edge(from, intermediate), path(intermediate, to).
 mayHaveTag(tgt, tag) :- claimHasTag(tgt, tag).
 mayHaveTag(tgt, tag) :- edge(src, tgt), mayHaveTag(src, tag).
```
bgogul, line 79: You might have to sync your branch to get Andrew's changes.

bgogul (approved):

src/analysis/souffle/taint.dl
markww: Just did this, thank you for the reminder.

PR #44, opened 2021-07-22
Support Downgrades Operating on Labels

aferr:
Previously downgrades were done using the relation
saysDowngrades(s: Principal, p: Path, t: tag) to remove just one tag.
Often, to avoid downgrading to a label that is too weak, this would
need to be followed up with a separate `claimHasTag` to add new tags
back in. This commit adds in the ability to downgrade to a particular
label with saysDowngradesLabel(s: Principal, p: Path, ell: Label).
This makes it a bit more convenient to do the downgrade followed by
upgrade sequence because the new label can have extra tags.

(When integrity tags are added eventually upgrades/downgrades work the
opposite way. Adding integrity tags (downgrades) would need to be controlled
whereas removing integrity tags (upgrades)  would not).


aferr:
FYI this just adds one small commit on top of [PR #33](https://github.com/google-research/raksha/pull/33). I can redo the PR after that one is checked in if that's easier. Otherwise it might be useful to just look at the last commit.

src/analysis/souffle/taint.dl
```diff
@@ -77,13 +77,26 @@
 // it).
 .decl ownsTag(owner: Principal, tag: Tag)
 
+// A statement from authorization logic that a path should be downgraded to a
+// particular label. This allows for multiple downgrades as well as "upgrades"
+// (increases to labels) to happen with the same relation.
+// NOTE: this is not in auth_logic_interface.dl because this relies on IFCLabel 
+.decl saysDowngradesLabel(speaker: Principal, path: AccessPath, newLabel: IFCLabel)
```
markww, line 84: Does it make sense to perform downgrades on the integrity aspect and the privacy aspect simultaneously? I can see having a general notion of "downgrades" which removes in one case and adds in the other getting highly confusing. It might make sense to have a separate `saysDowngradesPrivacyLabel` and `saysDowngradesIntegrityLabel`.

Similarly, should labels have a mix of integrity and privacy tags? Should integrity and privacy tags be part of the same `Tag` base type domain? Perhaps it should be the case that there are two separate base types:

```
.type IntegrityTag <: symbol
.type PrivacyTag <: symbol
```

And `claimHasTag` adds a tag in each of those domains.

This may make sense for another reason: Gogul and I were discussing how certain operations that make sense on one form of tag are incoherent on another. It makes sense to ask whether an `accessPath` may or may not have a privacy tag, but it makes little sense to ask whether it must have a privacy tag. Privacy tags are added on the slightest suspicion of flow to an `accessPath`, so it's not clear if any `accessPath` other than the original claim "must" have them. A similar argument can be made for the relationship between may and integrity tags.

I guess this question spirals out beyond this commit to an effort Gogul and I talked about where I would start splitting the notion of the tags in some way. I guess the time for that may be already here. Perhaps this is what I should work on next.

markww: See PR #45

aferr:
Closing this because labels are gone.

src/analysis/souffle/taint.dl
aferr: This PR was closed, so I did not respond initially. Sorry about that. Usually "downgrades" is the general term for making a label (which might have both confidentiality and integrity components) lower in the lattice. 

If we do use FLAM or Flume style labels, we will have separate confidentiality and integrity components, and probably separate relations for downgrading each of these components. Downgrading confidentiality is usually called declassification and downgrading integrity is usually called endorsement. We have been talking about using the downgrading rules for robust declassification and transparent endorsement -- both of those are described [in this paper](https://www.cs.cornell.edu/~ethan/papers/nmifc.pdf) (though there are also papers that give a gentler introduction to robust declassification in earlier work. I can send you them if that becomes a useful thing to do). 

I didn't separate them with this (now closed) PR at the time just because we only had confidentiality so far.

PR #45, opened 2021-07-22
Split the Tag domain into Privacy, Integrity, and Claimed tags.

markww:
As discussions on Raksha have evolved, it has become clear that there is
no coherent notion of a simple, unqualified "tag". Privacy tags spread
like taint in a worst-case scenario, integrity tags disappear unless
properly handled, and the initial tag the user claims immediately splits
into each of the above kinds. It makes sense to ask whether a Privacy
tag "may" be present, but not whether it "must" be present; it does not
make sense to ask whether an Integrity tag "may" be present, but is
sensible to ask whether it "must" be present.

To address this reality, this commit introduces new base types for
Privacy, Integrity, and Claimed tags and places the appropriate type on
the appropriate "presence" queries.

### DIFF
--- a/src/analysis/souffle/tags.dl
+++ b/src/analysis/souffle/tags.dl
@@ -33,7 +33,7 @@
 .type IntegrityTag <: symbol
 
 // This is a tag that a user or some principal claims is present upon some accessPath. This
-// introduces *both* a privacy and integrity tag upon that accessPath, which will each
+// introduces *both* a secrecy and integrity tag upon that accessPath, which will each
 // independently spread through the graph from that point according to their different behavior
 // patterns.
 .type ClaimedTag <: symbol
--- a/src/analysis/souffle/tags.dl
+++ b/src/analysis/souffle/tags.dl
@@ -1,0 +1,41 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+
+// Definition of the base tag types that are used by this dataflow analysis.
+
+#ifndef SRC_ANALYSIS_SOUFFLE_TAGS_DL_
+#define SRC_ANALYSIS_SOUFFLE_TAGS_DL_
+
+// A symbol representing a taint tag. SecrecyTags indicate that the data upon which they are
+// annotated potentially contain information that should be handled with some sort of special care
+// and should not be exposed to some outside group. The conservative approximation for SecrecyTags
+// is, when you are in doubt whether a particular datum should have this tag, annotate it. This
+// causes these tags to spread like taint.
+.type SecrecyTag <: symbol
+
+// A symbol representing a guarantee tag. IntegrityTags indicate that the data upon which they are
+// annotated have some critical property that handlers must take care to preserve. The conservative
+// approximation for IntegrityTags is, when you are in doubt whether a particular datum should have
+// this tag, do not annotate it. This causes these tags to disappear unless carefully handled.
+.type IntegrityTag <: symbol
+
+// This is a tag that a user or some principal claims is present upon some accessPath. This
+// introduces *both* a privacy and integrity tag upon that accessPath, which will each
+// independently spread through the graph from that point according to their different behavior
+// patterns.
+.type ClaimedTag <: symbol
+
+#endif // SRC_ANALYSIS_SOUFFLE_TAGS_DL_
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -40,7 +40,7 @@
 // (Useful for scoping when negation is involved.)
 .decl isAccessPath(accessPath: AccessPath)
 
-// The set of all privacy tags.
+// The set of all secrecy tags.
 // (Useful for scoping when negation is involved.)
 .decl allSecrecyTags(secrecyTag: SecrecyTag)
 
@@ -68,7 +68,7 @@
 // If true, `accessPath` *definitely* has the integrity property indicated by tag at runtime.
 .decl mustHaveIntegrityTag(accessPath: AccessPath, integrityTag: IntegrityTag)
 
-// There is a claim that `accessPath` definitely has `tag`. This introduces *both* a privacy and
+// There is a claim that `accessPath` definitely has `tag`. This introduces *both* a secrecy and
 // integrity tag.
 .decl claimHasTag(accessPath: AccessPath, claimedTag: ClaimedTag)
 
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -34,12 +34,12 @@
 
 .type AccessPath <: symbol
 
-// A symbol representing a taint tag. PrivacyTags indicate that the data upon which they are
+// A symbol representing a taint tag. SecrecyTags indicate that the data upon which they are
 // annotated potentially contain information that should be handled with some sort of special care
-// and should not be exposed to some outside group. The conservative approximation for PrivacyTags
+// and should not be exposed to some outside group. The conservative approximation for SecrecyTags
 // is, when you are in doubt whether a particular datum should have this tag, annotate it. This
 // causes these tags to spread like taint.
-.type PrivacyTag <: symbol
+.type SecrecyTag <: symbol
 
 // A symbol representing a guarantee tag. IntegrityTags indicate that the data upon which they are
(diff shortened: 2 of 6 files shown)


src/analysis/souffle/taint.dl
```diff
@@ -34,48 +34,82 @@
...
+allIntegrityTags(as(claimedTag, IntegrityTag)) :- claimHasTag(_, claimedTag).
+allIntegrityTags(integrityTag) :- mustHaveIntegrityTag(_, integrityTag).
 
 // Transitive paths
 path(from, to) :- edge(from, to).
 path(from, to) :- edge(from, intermediate), path(intermediate, to).
 
-mayHaveTag(tgt, tag) :- claimHasTag(tgt, tag).
-mayHaveTag(tgt, tag) :- edge(src, tgt), mayHaveTag(src, tag).
+mayHavePrivacyTag(tgt, as(claimedTag, PrivacyTag)) :- claimHasTag(tgt, claimedTag).
```
bgogul, line 108: what does `as(claimedTag, PrivacyTag)` do?

bgogul (changes requested):
Can you split the tags of to a separate file? The `taint.dl` is combining too many concepts in one place.

src/analysis/souffle/taint.dl
markww: `as` is a cast from one base type with a particular backing primitive type into another base type with the same backing primitive type. In this case, I'm saying "take the value of `claimedTag` and treat it as a `PrivacyTag` so long as it also is backed by `symbol`."

src/analysis/souffle/taint.dl
```diff
@@ -34,48 +34,82 @@
...
-// A symbol representing a taint (or information flow) tag. These would be
-// strings of the form "secret", "public", "timestampInMs", etc. Just like
-// `AccessPath`, it would make sense to define an ADT in the long term.
-.type Tag <: symbol
+// A symbol representing a taint tag. PrivacyTags indicate that the data upon which they are
+// annotated potentially contain information that should be handled with some sort of special care
+// and should not be exposed to some outside group. The conservative approximation for PrivacyTags
+// is, when you are in doubt whether a particular datum should have this tag, annotate it. This
+// causes these tags to spread like taint.
+.type PrivacyTag <: symbol
```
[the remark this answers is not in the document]
markww: yes, certainly.

src/analysis/souffle/taint.dl
```diff
@@ -34,48 +34,82 @@
...
 .decl path(src: AccessPath, tgt: AccessPath)
 
-// Declares that `accessPath` may have be tainted with `tag`.
-// If false, `accessPath` *definitely* does not `tag` taint at runtime.
-// If true, `accessPath` may have the `tag` taint at runtime.
-.decl mayHaveTag(accessPath: AccessPath, tag: Tag)
+// Declares that `accessPath` may have be tainted with `privacyTag`.
+// If false, `accessPath` *definitely* does not `privacyTag` taint at runtime.
+// If true, `accessPath` may have the `privacyTag` taint at runtime.
+.decl mayHavePrivacyTag(accessPath: AccessPath, privacyTag: PrivacyTag)
```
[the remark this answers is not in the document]
markww: Will do.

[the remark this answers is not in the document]
markww: Done.

src/analysis/souffle/taint.dl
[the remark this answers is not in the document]
markww: Done.

markww:
I think I addressed all of your comments Gogul, PTAL, thanks!

src/analysis/souffle/tests/arcs_fact_tests/ok_claim_propagates.dl
```diff
@@ -32,5 +32,9 @@ edge("R.P2.bar", "R.P2.foo").
...
 edge("R.h2.Foo", "R.P3.bar").
 
-// This test passes if R.P3.bar has the trusted tag.
-TEST_CASE("ok_claim_propagates") :- mayHaveTag("R.P3.bar", "userSelection").
+TEST_CASE("initial_has_secrecy") :- mayHaveSecrecyTag("R.P1.foo", "userSelection").
+TEST_CASE("initial_has_integrity") :- mustHaveIntegrityTag("R.P1.foo", "userSelection").
+// This test passes if R.P3.bar has the userSelection tag.
+TEST_CASE("ok_claim_propagates_secrecy") :- mayHaveSecrecyTag("R.P3.bar", "userSelection").
+TEST_CASE("ok_claim_not_propagates_integrity") :-
+  allIntegrityTags("userSelection"), !mustHaveIntegrityTag("R.P3.bar", "userSelection").
```
bgogul, line 40: is `allIntegrityTags("userSelection")` at all?

src/analysis/souffle/taint.dl
```diff
@@ -32,50 +32,67 @@
...
-// `AccessPath`, it would make sense to define an ADT in the long term.
-.type Tag <: symbol
+.type AccessPath <: symbol
 
 // Is this symbol an access path to some data?
 // (Useful for scoping when negation is involved.)
 .decl isAccessPath(accessPath: AccessPath)
 
-// Is this a Tag?
+// The set of all secrecy tags.
```
bgogul, line 43: It might make sense to move all of these to `tags.dl` as well. However, let us wait until Andrew merges #33

bgogul (approved):

src/analysis/souffle/tags.dl
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef SRC_ANALYSIS_SOUFFLE_TAGS_DL_
+#define SRC_ANALYSIS_SOUFFLE_TAGS_DL_
+
+// A symbol representing a taint tag. SecrecyTags indicate that the data upon which they are
+// annotated potentially contain information that should be handled with some sort of special care
+// and should not be exposed to some outside group. The conservative approximation for SecrecyTags
+// is, when you are in doubt whether a particular datum should have this tag, annotate it. This
+// causes these tags to spread like taint.
+.type SecrecyTag <: symbol
```
bgogul, line 27: Actually, I don't think we should have separate types for `Secrecy` and `Integrity`. We just have a `Tag` type that would have an "integrity` and `secrecy` components that are captured via relations: `mayHaveSecrecyTag`, `mustHaveIntegrityTag`, etc. A same label like `product_id` could be used for capturing integrity and secrecy.

bgogul (changes requested):

src/analysis/souffle/tags.dl
aferr: Agreed. There should be one type of tag and we just need two different sets of relations for each. Both the confidentiality and integrity components of labels are sets, but they use different lattices (one is ordered by subset and the other is ordered by superset -- as a result the meet/join are also flipped). With the way IFC is usually used, a pretty common label would be to have the same set for both confidentiality and integrity (e.g. C = {Alice}, I = {Alice} ), and keeping tags in the same type allows this common label.

src/analysis/souffle/tags.dl
```diff
@@ -0,0 +1,41 @@
+//-----------------------------------------------------------------------------
```
aferr, line 1: @markww If you want to work on supporting confidentiality and integrity, it could be useful for you to take a look at the [Flume](https://pdos.csail.mit.edu/papers/flume-sosp07.pdf) paper. The label model we would use here is pretty similar to that one because it is based on pairs of sets. (The flowsto, meet, and join will work about the same for us as they do in that paper. Downgrading using capabilities and other aspects of their label model are not things we will pick up).

Another label model we have talked about often is [FLAM](https://www.cs.cornell.edu/andru/papers/flam/flam-csf15.pdf), which is also quite nice -- but I think it's more useful to look at Flume first because what we will implement will really be closer to Flume. Flume and FLAM actually have a lot in common (as do all IFC label models essentially) once you notice FLAM's normal form has a confidentiality and integrity component (and that all labels can be reduced to normal form).

We can't actually represent FLAM in datalog because checking flowsto in FLAM is reducible to the SAT problem because for both of its components it uses full (rather than semi-) lattices. Flume *is* implementable in datalog because its label components use just semi-lattices. 

(It's possible that if we had our own datalog solver we could approximate FLAM by using the hypercube label model, but this is slow with datalog's bottom-up solver --- let's save that for the distant future).

src/analysis/souffle/tags.dl
aferr, line 27: The term confidentiality is used more often than secrecy in IFC literature (although the Flume paper happens to be one that uses secrecy :) )

src/analysis/souffle/tags.dl
```diff
@@ -0,0 +1,41 @@
...
+// annotated have some critical property that handlers must take care to preserve. The conservative
+// approximation for IntegrityTags is, when you are in doubt whether a particular datum should have
+// this tag, do not annotate it. This causes these tags to disappear unless carefully handled.
+.type IntegrityTag <: symbol
+
+// This is a tag that a user or some principal claims is present upon some accessPath. This
+// introduces *both* a secrecy and integrity tag upon that accessPath, which will each
+// independently spread through the graph from that point according to their different behavior
+// patterns.
+.type ClaimedTag <: symbol
```
aferr, line 39: What's this for? Is it meant to distinguish between the tags that were explicitly given by some programmer from the tags that are inferred?

markww:
This PR has gotten quite stale, and it sounds like the core change it proposes is unpopular. Closing.

PR #46, opened 2021-07-22
Ensure that each TEST_CASE must be provided a unique name.

markww:
Before this, if someone provided two TEST_CASEs with the same name on
accident, and one failed, the test would still pass. This would occur
because the duplicated name would be added idempotently to allTests
twice, but to testSucceeds once, resulting in the condition for
testFails for that case not triggering. Adding a guaranteed-unique
number to each TEST_CASE as well with the $ functor allows us to catch
this case and error on it in the handler.


bgogul (approved):
Nice!

PR #49, opened 2021-07-23
Imports the Arcs manifest parser into third_party/arcs

bgogul:
This PR also does the following:
 - Adds a `arcs_manifest_proto` skylark rule to convert a `.arcs` file into a `.binarybp` proto file. 
 - Moves the .proto files from `src/ir` to`third_party/arcs/proto` to preserve arcs provenance.
 - Adds a test for the parser to `.github/workflows/main.yml`


third_party/arcs/proto/manifest.proto
```diff
@@ -111,11 +109,9 @@ message TypeProto {
     SingletonTypeProto singleton = 7;
     CountTypeProto count = 8;
     ListTypeProto list = 9;
-    NullableTypeProto nullable = 13;
   }
+  bool optional = 10;
```
markww, line 113: Will taking over `proto` field number 10 for `optional` instead of whatever it was used for in the past cause incompatibilities with old Arcs `proto`s we might analyze? Maybe there aren't enough of them for it to matter, but it might be safer to keep the `reserved 10` and give the new `optional` a new number? Similar concerns for Enum member deletions.

markww: Or is the diff misleading me? Is this the `proto` file as it appears in the original Arcs repo?

third_party/arcs/examples/BUILD
```diff
@@ -0,0 +1,11 @@
+load("//build_defs:arcs.bzl", "arcs_manifest_proto")
+
+arcs_manifest_proto(
```
markww, line 3: It looks like we have this example Arcs manifest file tree and a recipe to build it, but I don't think it's hooked up to `@raksha//:all_tests` anywhere. As such, the CI isn't doing anything to check the health of the parser. Could you add a test that does that, even if it's a simple smoke test that just parses these manifests, checks that a proto file is produced, and checks that the exit code of the parser is 0?

third_party/arcs/proto/manifest.proto
bgogul: I noticed this just after sending the PR for review. It seems like the webpack bundle was created off of an earlier commit in the Arcs repo. I am trying to build one off of HEAD. I will hold off merging this until I resolve this one way or another. 

In any case, I don't expect us to use protos created outside of our repo. So, it shouldn't be a big deal even if we use a slightly older version of the proto.

third_party/arcs/examples/BUILD
bgogul: Yep, will do. Could you point me to how you set up the CI for the facts tests?

markww: Sure. Github Actions uses the source file `.github/workflows/main.yml` to figure out what to do. As you can see from that config file, it has a single workflow named `CI` which, upon any `push` or `pull_request` to the `main` branch, runs the single job `test` in `jobs`. `test` just checks out the repo, uses `apt-get` to grab a bunch of Ubuntu packages that Souffle needs, and runs `bazel test :all_tests` on the checked-out repo.

If you just edit the build so that `bazel test :all_tests` eventually kicks off a test of the `manifest2proto` parser, that should do it. If it needs any additional Ubuntu packages, you can add that to the `apt-get` invocation in that `main.yml` file.

If you want to do something more than the above and need to further edit the `main.yml` file, editing it within the GitHub web UI provides some helpful hints from GitHub Marketplace about little code snippets you can use to cobble some actions together.

third_party/arcs/proto/manifest.proto
markww: OK, whichever way this resolves, this sounds fine to me.

third_party/arcs/examples/BUILD
bgogul: I ended up adding another step in the .yml file.

third_party/arcs/proto/manifest.proto
bgogul: Let us get this in. We can update the protos and the webpack in a subsequent PR.

markww: SGTM.

third_party/arcs/examples/BUILD
markww: Works for me.

markww (approved):

PR #51, opened 2021-07-26
Low Graph IR in C++ and Souffle Translation

aferr:
This PR makes the C++ side of the low graph IR and a translation into datalog. The protobufs still need to be translated into this low graph IR (potentially by first being translated into another higher-level IR).


src/graph_ir/low_graph_ir.cc
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+
+#include "low_graph_ir.h"
+#include <string>
+#include <vector>
+#include <sstream>
+
+using namespace std;
```
markww, line 22: Usually not recommended, the `std` namespace contains a huge amount of stuff. Better to just explicitly qualify things from `std` with `std::`.

src/graph_ir/low_graph_ir.cc
```diff
@@ -0,0 +1,53 @@
...
+
+string LowGraphAccessPath::PrettyPrint() {
+    // to copy string
```
markww, line 25: If you just want to return a copy, why assign to a local instead of just returning it through `std::string`?

src/graph_ir/low_graph_ir.cc
```diff
@@ -0,0 +1,53 @@
...
+    string pretty = full_path_name;
+    return pretty;
+}
+
+string LowGraphNode::PrettyPrintAPVec(vector<LowGraphAccessPath> vec) {
+    stringstream ss;
```
markww, line 31: It looks like you have a pattern where, in many pretty print functions, you create a local `stringstream`, populate it with a formatted string, then use `stringstream::str` to create that string. Then, at the callsite, you often perform the same trick again.

It might be more concise to take as a parameter a `std::ostream &` reference rather than constructing all of these stringstreams.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+
+
+#include <string>
+#include <vector>
+#include <set>
+
+#ifndef LOW_GRAPH_IR_
+#define LOW_GRAPH_IR_
+
+using namespace std;
```
markww, line 35: Definitely do not use `using namespace std;` in a header.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+ * input/output access path lists. It can be translated directly into datalog
+ * relations that similarly represent particles with flattened lists of full 
+ * access path names.
+ *
+ * There is likely a higher-level IR from which this compiles, and the higher 
+ * level IR can be generated by deserializing protobufs.
+ */
+
+
+#include <string>
```
markww, line 28: It won't make any practical difference for standard headers that are certainly doing the right thing, but put your `#include`s in the `#ifndef` guard. By placing them here, you're assuming that these headers are themselves `#ifndef`ed correctly. If one of these headers omitted or messed up its #ifndef` guard, you might introduce a very weird and hard-to-debug error by having these dropped in in a case where the rest of this body was omitted due to the `#ifndef` guard.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+       vector<LowGraphAccessPath> outputs;
+       LowGraphNode(string _name, vector<LowGraphAccessPath> _inputs,
+               vector<LowGraphAccessPath> _outputs) :
+           name(_name), inputs(_inputs), outputs(_outputs) {};
+       string PrettyPrint();
+    public:
+        static string PrettyPrintAPVec(vector<LowGraphAccessPath>);
+};
+
+typedef vector<LowGraphNode> LowGraph;
```
markww, line 58: I'd say it'd be better to just make a simple class that had a single `std::vector<LowGraphNode>` field called `LowGraph`. You get better type safety that way and you can more easily attach methods if you desire them later.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+#define LOW_GRAPH_IR_
+
+using namespace std;
+
+class LowGraphAccessPath {
+    public:
+        string full_path_name;
+        LowGraphAccessPath(string _full_path_name) : 
+            full_path_name(_full_path_name) {};
+        string PrettyPrint();
```
markww, line 42: For these pretty printing functions, you should probably have them declare their `this` as `const` like so:

```
std::string PrettyPrint() const;
```

Just to safeguard against accidentally modifying while doing something that should be a purely-read action.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+};
+
+class LowGraphNode {
+    public:
+       string name;
```
markww, line 47: Most of the members I see in this header do not change after construction and thus should be `const`.

src/graph_ir/low_graph_souffle_emitter.cc
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+#include "low_graph_souffle_emitter.h"
+
+using namespace std;
+
+string LowGraphSouffleEmitter::EmitPathList(
+        vector<LowGraphAccessPath> pathList) {
+    string ret = "$Nil()";
+    for (auto path: pathList) {
+        ret = "$Cons(" + path.full_path_name + ", " + ret + ")";
```
markww, line 27: This is a place where you do not want to use string concatenation. While the `+` operator on strings is fine in a leaf method when the number of concatenations is statically known, doing it `N` times for some input parameter `N` results in `N^2` performance. Use `ostringstream` instead. Similar throught this file.

src/graph_ir/low_graph_souffle_emitter.cc
```diff
@@ -0,0 +1,48 @@
...
+
+string LowGraphSouffleEmitter::EmitNode(LowGraphNode node) {
+    return "bindFlatParticle(" + node.name + ", " +
+        EmitNodeBody(node) + ").";
+}
+
+string LowGraphSouffleEmitter::EmitSouffleFromLowGraph(LowGraph low_graph) {
+    string ret = "";
+    for (auto node : low_graph) {
+        ret = ret += EmitNode(node) + "\n\n";
```
markww, line 45: I do not know what the behavior of `ret = ret +=` is, but I suspect it to be unspecified and compiler-dependent. Replace this with `ostringstream` for perofrmance and correctness reasons.

src/graph_ir/tests/test_low_graph_souffle_emitter.cc
```diff
@@ -0,0 +1,55 @@
...
+    // Particle 2: ImageSelector
+    //     reads:
+    //         image_detection_boxes
+    //         image_selection   // image selection action from user
+    //     writes:
+    //         selected_image_id
+
+    // Particle 1
+    vector<LowGraphAccessPath> image_detector_inputs = {
+        *(new LowGraphAccessPath("sensor_data_packet.camera_feed")),
```
markww, line 27: In general, this is poor form, as it guarantees a memory leak. In this test, it won't have any serious consequence, as `exit` will be your garbage collector, but it still doesn't look great. I suspect that this would be better expressed by just creating the test data using https://en.cppreference.com/w/cpp/language/list_initialization .

markww (changes requested):

src/graph_ir/low_graph_ir.cc
markww: Also, instead of returning this string, consider just printing it by taking an `ostream &` as a parameter.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+#include <set>
+
+#ifndef LOW_GRAPH_IR_
+#define LOW_GRAPH_IR_
+
+using namespace std;
+
+class LowGraphAccessPath {
+    public:
+        string full_path_name;
```
markww, line 39: In general, you should make fields private and mediate access to them through methods.

src/graph_ir/tests/test_low_graph_souffle_emitter.cc
```diff
@@ -0,0 +1,55 @@
+#include "low_graph_ir.h"
+#include "low_graph_souffle_emitter.h"
+
+#include <vector>
+#include <string>
+#include <iostream>
+
+using namespace std;
+
+LowGraph MakeTestGraph() {
```
markww, line 10: Instead of creating a function that returns a global it constructs, why not just create a number of `static const` globals? If you used a constant expression as the initializer for a `const` global, you would be freed from thinking about copys, initialization, or really any control flow in building your test graph.

src/graph_ir/tests/test_low_graph_ir.cc
```diff
@@ -0,0 +1,19 @@
+#include "low_graph_ir.h"
```
markww, line 1: You have these files that have `test` in the title, but they don't seem to be hooked up to any test rules. Please do so so they can be automatically run.

markww (changes requested):

src/graph_ir/low_graph_ir.cc
bgogul: This is better as follows:

```
const std::string& LowGraphAccessPath::PrettyPrint() { return full_path_name; }
```

The copy will happen as needed at the call site.

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+
+#include <string>
+#include <vector>
+#include <set>
+
+#ifndef LOW_GRAPH_IR_
+#define LOW_GRAPH_IR_
+
```
bgogul, line 34: `LOW_GRAPH_IR_H_`. See https://google.github.io/styleguide/cppguide.html#The__define_Guard

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+using namespace std;
+
+class LowGraphAccessPath {
```
bgogul, line 37: Why don't we call it `AccessPath`?

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+    public:
+        string full_path_name;
+        LowGraphAccessPath(string _full_path_name) : 
+            full_path_name(_full_path_name) {};
+        string PrettyPrint();
+};
+
+class LowGraphNode {
```
bgogul, line 45: Why don't we call it `Node` or `ProcessingNode`?

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,60 @@
...
+    public:
```
bgogul, line 46: Why are these public data members?

bgogul (changes requested):
@aferr I started reviewing the code, but started wondering if it is too early to write this translator. Here are my specific concerns/questions:
  - Do we have enough confidence that we have the right interfaces/relations in our datalog programs?
  - LowGraphIR seems to closely map the `FlatModule.dl`, but I think it is useful to define something higher-level like `ProcessorNode` (or `Node` or `Particle`) and then generate the relations in `FlatModule.dl`. 
  - How are you planning to use this now? How will this be integrated into our workflow eventually?

src/graph_ir/low_graph_ir.h
aferr: done

src/graph_ir/low_graph_ir.h
aferr: done.

src/graph_ir/low_graph_ir.h
aferr: Similar reasoning as the other LowGraph... name. Leaving this open to see what you think.

(Could also call this LowGraphParticle ? )

src/graph_ir/low_graph_ir.h
aferr: fixed.

src/graph_ir/low_graph_ir.h
aferr: wasn't sure if changing would be needed elsewhere, so made it permissive since this is a research prototype and I imagine we're rushing towards "runs at all".

Made them const now anyway.

src/graph_ir/low_graph_ir.h
aferr: It seems like we can just do that once we need it, so I'll skip it for now? Re-open this if you think it's important to do this.

src/graph_ir/low_graph_souffle_emitter.cc
aferr: This code needs to put text at the beginning of the previous version of the string, which I think requires a copy per iteration even if a stringstream is used. Am I mistaken? 

For the rest, I'm now using stringstreams.

src/graph_ir/tests/test_low_graph_ir.cc
aferr: Do we have cpp test rules yet? I thought we only had datalog test rules.

I also don't know how to do this. Can I do it in a follow-up PR so that I can get this in and then look it up after ?

src/graph_ir/tests/test_low_graph_souffle_emitter.cc
aferr: ok fixed.

src/graph_ir/tests/test_low_graph_souffle_emitter.cc
aferr: done.

aferr:
> @aferr I started reviewing the code, but started wondering if it is too early to write this translator. Here are my specific concerns/questions:
> 
> * Do we have enough confidence that we have the right interfaces/relations in our datalog programs?
> * LowGraphIR seems to closely map the `FlatModule.dl`, but I think it is useful to define something higher-level like `ProcessorNode` (or `Node` or `Particle`) and then generate the relations in `FlatModule.dl`.
> * How are you planning to use this now? How will this be integrated into our workflow eventually?

I figured we would have two layers of IRs one would be a higher-level one that is more similar to the particle graph part of arcs manifests (and includes nested structs as inputs/outputs), and this low-level IR that can be translated directly into datalog relations. We could write code that translates from the high-level IR into this low-level one.

src/graph_ir/low_graph_souffle_emitter.cc
markww: You raise a good point, and that makes it so that `stringstream` is not as easy to drop in. Rather than repeatedly nesting, I'd say just construct the prefix and the suffix separately as follows:

std::stringstream result;
for (auto path: pathList) {
  result << "$Cons(" << path.full_path_name << ", ";
}
result << "$Nil()";
for (int i = 0; i < pathList.size(); ++i) {
  result << ")";
}

This will run in `O(N)` time, where `N` is the number of path elements.

src/graph_ir/low_graph_ir.h
markww: Sure, that is fine.

src/graph_ir/tests/test_low_graph_ir.cc
```diff
@@ -0,0 +1,21 @@
...
+
+#include <vector>
+#include <string>
+#include <iostream>
+
+namespace raksha_graph_ir {
+
+int main() {
+    LowGraphAccessPath* p_a = new LowGraphAccessPath("a");
```
markww, line 10: If you use `new` to construct something, it is best to `delete` it later. But in this case, where all of these items are used locally in the same scope, it's better to not get the heap involved at all. Just declare these objects in automatic storage like so:

```C++
LowGraphAccessPath p_a("a");
```

src/graph_ir/low_graph_ir.h
bgogul: It will be easier to maintain if we have a single `AccessPath` class that gives us a flattened view as needed. It is fine to starting with a flattened representation and transparently adding the nesting and hierarchy later. For example,
```
class AccessPath {
  AccessPath(const std::string& full_path_name) : full_path_name_(full_path_name) {}
  const std::string& full_path_name() const { return full_path_name_; }
  ... 
}
```

Later, we can do the following:

```
class AccessPath {
  AccessPath(const Root& root, const std::set<Selector>& selectors) ...
  const std::string& full_path_name() const { 
     // either return a pre-computed full_path or compute from the internal state.
     return full_path_name_; 
  }
  ... 
}
```

src/graph_ir/low_graph_ir.h
bgogul: See related comment for `AccessPath`

src/graph_ir/low_graph_ir.h
```diff
@@ -0,0 +1,68 @@
...
+ * level IR can be generated by deserializing protobufs.
+ */
+
+#ifndef LOW_GRAPH_IR_H_
+#define LOW_GRAPH_IR_H_
+
+#include <string>
+#include <vector>
+
+namespace raksha_graph_ir {
```
bgogul, line 33: Nested namespaces are better here:

```
namespace raksha {
namespace ir {
```

src/graph_ir/tests/test_low_graph_ir.cc
bgogul: I just imported gtest into our repo. See https://github.com/google-research/raksha/blob/7ed3da8347f513249c8508f63ba265640ad1f063/src/analysis/souffle/cpp_interface_test.cc#L25 for an example.

bgogul (commented):
> I figured we would have two layers of IRs one would be a higher-level one that is more similar to the particle graph part of arcs manifests (and includes nested structs as inputs/outputs), and this low-level IR that can be translated directly into datalog relations. We could write code that translates from the high-level IR into this low-level one.

Even if we want to have two layers of IRs, it should be possible to share the data structures across the different layers of IR. See my comments in the PR. 

Defining C++ data structures would be useful, but I am still wondering if now is the right time. Let us chat a bit more offline and sort this out.

src/graph_ir/low_graph_ir.h
aferr: OK, doing it the way you suggested (having one set of classes for all layers of graph IRs) does seem reasonable. I thought there was some value in having distinct types for different layers so that type checking would prevent a pass that was written for one layer (e.g. flattened) from being used on another layer (e.g. not flattened), but this doesn't seem too important and could get in the way. Let's have one set of classes as you said.

PR #52, opened 2021-07-26
Tighten up member rule by requiring base to start at index 0 of member.

markww:
Originally, I used "contains" to check whether base was the base
of member when propagating taint from members to base paths. That isn't
the full story though: we want to propagate taint only when the base is
a prefix of the member starting at index 0. This distinction seems to
not matter when edges only start at leaf fields, but better to be
precise.

### DIFF
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -58,6 +58,17 @@ isAccessPath(path) :- mayHaveTag(path, _).
 isTag(tag) :- claimHasTag(_, tag).
 isTag(tag) :- mayHaveTag(_, tag).
 
+// A pair of (base, member) is in this set if the base access path is a non-trivial subpath of the
+// member access path.
+.decl isMemberOf(base: AccessPath, member: AccessPath)
+
+// The member is a member of the base if it starts with the base plus '.'.
+isMemberOf(base, member) :-
+  isAccessPath(base),
+  isAccessPath(member),
+  strlen(base) + 1 < strlen(member),
+  cat(base, ".") = substr(member, 0, strlen(base) + 1).
+
 // An access path may have a tag if some subpath to a member field has that tag. This allows
 // checking for some inner node in the access path whether any leaf node of that node might have
 // a particular tag.
@@ -65,6 +76,6 @@ mayHaveTag(base, tag) :-
   isAccessPath(base),
   isAccessPath(member),
   mayHaveTag(member, tag),
-  contains(cat(base, "."), member).
+  isMemberOf(base, member).
 
 #endif // SRC_ANALYSIS_SOUFFLE_TAINT_DL_
--- a/src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
+++ b/src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
@@ -1,0 +1,55 @@
+#include "taint.dl"
+#include "fact_test_helper.dl"
+
+claimHasTag("R.foo.a", "userSelection").
+edge("R.foo.a", "R.bar.a").
+edge("R.foo.b", "R.bar.b").
+edge("R.foobaz.a", "R.bar.c").
+
+// Make R, R.foo, and R.foobaz checked access paths, but not foo.bar
+isAccessPath("R").
+isAccessPath("R.foo").
+isAccessPath("R.foobaz").
+
+// We use this to ensure that we select exactly the (base, member) pairs we expect to see in
+// isMemberOf.
+.decl expectedIsMemberOf(base: AccessPath, member: AccessPath)
+
+expectedIsMemberOf("R", "R.foo").
+expectedIsMemberOf("R", "R.foobaz").
+
+expectedIsMemberOf("R", "R.foo.a").
+expectedIsMemberOf("R", "R.foo.b").
+
+expectedIsMemberOf("R.foo", "R.foo.a").
+expectedIsMemberOf("R.foo", "R.foo.b").
+
+expectedIsMemberOf("R", "R.bar.a").
+expectedIsMemberOf("R", "R.bar.b").
+expectedIsMemberOf("R", "R.bar.c").
+
+expectedIsMemberOf("R", "R.foobaz.a").
+expectedIsMemberOf("R.foobaz", "R.foobaz.a").
+
+// Used to report expectations that went unmatched.
+.decl unmatched(base: AccessPath, member: AccessPath, from: symbol)
+
+// Useful to make these output relations if things fail.
+.output isAccessPath(IO=stdout, delimiter=",")
+.output unmatched(IO=stdout, delimiter=",")
+.output isMemberOf(IO=stdout, delimiter=",")
+
+unmatched(base, member, "expected") :-
+  expectedIsMemberOf(base, member), !isMemberOf(base, member).
+
+unmatched(base, member, "actual") :-
+  !expectedIsMemberOf(base, member), isMemberOf(base, member).
+
+TEST_CASE("unmatched_is_empty") :- count : { unmatched(_, _, _) } = 0.
+
+TEST_CASE("size_of_expected_and_actual_match_and_not_zero") :-
+  count_expected = count : { expectedIsMemberOf(_, _) },
+  count_actual = count : { isMemberOf(_, _) },
+  count_expected != 0,
+  count_actual != 0,
+  count_expected = count_actual.
--- a/src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
+++ b/src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
@@ -1,6 +1,8 @@
 #include "taint.dl"
 #include "fact_test_helper.dl"
 
+// Note: these claims and edges are just used to introduce access paths in a fashion that is
+// similar to what we might see in the wild.
 claimHasTag("R.foo.a", "userSelection").
 edge("R.foo.a", "R.bar.a").
 edge("R.foo.b", "R.bar.b").
--- a/src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
+++ b/src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
@@ -1,14 +1,18 @@
 #include "taint.dl"
 #include "fact_test_helper.dl"
 
-// Note: these claims and edges are just used to introduce access paths in a fashion that is
-// similar to what we might see in the wild.
-claimHasTag("R.foo.a", "userSelection").
-edge("R.foo.a", "R.bar.a").
-edge("R.foo.b", "R.bar.b").
-edge("R.foobaz.a", "R.bar.c").
+// Leaf AccessPaths: These are AccessPaths to leaf fields, similar to what we expect might be
+// introduced by a claim or an edge.
+isAccessPath("R.foo.a").
+isAccessPath("R.foo.b").
 
-// Make R, R.foo, and R.foobaz checked access paths, but not foo.bar
+isAccessPath("R.bar.a").
+isAccessPath("R.bar.b").
+isAccessPath("R.bar.c").
+
+isAccessPath("R.foobaz.a").
+
+// Inner node AccessPaths. Make R, R.foo, and R.foobaz checked access paths, but not foo.bar.
 isAccessPath("R").
 isAccessPath("R.foo").
 isAccessPath("R.foobaz").
--- a/src/analysis/souffle/fact_test_helper.dl
+++ b/src/analysis/souffle/fact_test_helper.dl
@@ -51,5 +51,14 @@ testFails(testAspectName) :- allTests(testAspectName), !testPasses(testAspectNam
   allTestsAndCaseNum(test_aspect_name, $) :- 1 = 1. \
   testPasses(test_aspect_name)
 
+#define CHECK_TAG_PRESENT(access_path, tag) \
+  isAccessPath(access_path). \
+  TEST_CASE(cat(cat(cat("is_", tag), "_present_in_"), access_path)) :- \
+    isAccessPath(access_path), mayHaveTag(access_path, tag)
+
+#define CHECK_TAG_NOT_PRESENT(access_path, tag) \
+  isAccessPath(access_path). \
+  TEST_CASE(cat(cat(cat("is_", tag), "_not_present_in_"), access_path)) :- \
+    isAccessPath(access_path), !mayHaveTag(access_path, tag)
 
 #endif // SRC_ANALYSIS_SOUFFLE_ANALYZE_TAG_CHECKS_TEST_HELPER_DL_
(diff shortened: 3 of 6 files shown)


src/analysis/souffle/taint.dl
```diff
@@ -130,6 +130,7 @@ mayHaveTag(base, tag) :-
   isAccessPath(base),
   isAccessPath(member),
   mayHaveTag(member, tag),
-  contains(cat(base, "."), member).
+  // The member is a member of the base if it starts with the base plus '.'.
+  substr(member, 0, strlen(base) + 1) = cat(base, ".").
```
bgogul, line 134: Can we define a datalog relation like `isBaseOf` (or some relevant name) and write some tests?

bgogul (changes requested):

src/analysis/souffle/taint.dl
markww: I've done something similar with `isMemberOf`.

src/analysis/souffle/taint.dl
```diff
@@ -110,26 +110,68 @@ mayHaveTag(tgt, tag) :-
...
+  isAccessPathIndex(index),
+  index < strlen(accessPath),
+  substr(accessPath, index, 1) = ".".
+
+// Any access path that was directly mentioned is an access path.
+isAccessPath(accessPath) :- isDirectlyMentionedAccessPath(accessPath).
+
+// Any prefix of a directly mentioned access path that ends in a dot is an access path (not
+// including the dot).
+isAccessPath(substr(accessPath, 0, boundaryIdx)) :-
```
bgogul, line 147: Datalog computes all the relations eagerly. Therefore, this rule forces datalog solver to compute tags for all partial prefixes. The advantage of defining `isAccessPath` from `edge`, `claimHasTag`, `mayHaveTag`, etc., is that we will only compute tags for the access paths that we care about. 

Is this new inference rule needed for correctness of the `isMemberOf` rule?

bgogul (changes requested):

src/analysis/souffle/taint.dl
markww: I think, unfortunately, that either this elaboration or an alternative mechanism is needed to accomplish the `isMemberOf` rule. See the `isMemberOf` rule below:

```
isMemberOf(base, member) :-
  isAccessPath(base),
  isAccessPath(member),
  strlen(base) + 1 < strlen(member),
  cat(base, ".") = substr(member, 0, strlen(base) + 1).
  ```

In order to ground the variables used in the rule, I have to check that they are within `isAccessPath`. Here are the rules that populate `isAccessPath` other than those that consider partial prefixes:

```
// Access paths directly mentioned by other facts.
.decl isDirectlyMentionedAccessPath(accessPath: AccessPath)

// Symbols used in edges are access paths
isDirectlyMentionedAccessPath(x) :- edge(x, _).
isDirectlyMentionedAccessPath(y) :- edge(_, y).

isDirectlyMentionedAccessPath(path) :- claimHasTag(path, _).
isDirectlyMentionedAccessPath(path) :- mayHaveTag(path, _).

// Any access path that was directly mentioned is an access path.
isAccessPath(accessPath) :- isDirectlyMentionedAccessPath(accessPath).
```

If `member` is a leaf, it will be directly mentioned by an `edge` or `claim`, so that is not a problem. If `base` is an inner node, however, (as we expect it to be in all cases that are not trivially false) it will not be mentioned by an `edge`. It will also not be the subject of a `claimHasTag` (even if it were, this would be a trivially-true case, and not very interesting). The only way that an inner node can become part of `isAccessPath` is through `mayHaveTag`.

However, the one rule that could place it in `mayHaveTag` is the rule that propagates tags from members to bases. Therefore, there is a circular dependency unless we get this partial path into `isAccessPath` in a manner that is more directly connected to base facts.

If we do not want to derive all possible partial prefixes, I think the best way to do this is to introduce a new kind of fact that indicates what paths are of interest to checks. For instance, we could use the following:

```
.decl checkedAccessPath(accessPath: AccessPath)

isAccessPath(accessPath) :- checkedAccessPath(accessPath).
```

and introduce a `checkedAccessPath` for all inner nodes that are the subject of a check. A real system, of course, would probably just introduce one of these for any check node, but we might as well save ourselves some typing.

This is similar to the idea I had initially of introducing nodes for `checkHasTag` and friends, but unlike those nodes, this kicks the question of "how do we represent complex predicates on tags" down the road. We simply indicate that an access path is checked and allow that to populate the universe of `AccessPath`s.

What do you think?

Comment (bgogul) [reply src/analysis/souffle/taint.dl:147]: > If we do not want to derive all possible partial prefixes, I think the best way to do this is to introduce a new kind of fact that indicates what paths are of interest to checks. For instance, we could use the following:
> ```
> .decl checkedAccessPath(accessPath: AccessPath)
> isAccessPath(accessPath) :- checkedAccessPath(accessPath).
> ```
> and introduce a checkedAccessPath for all inner nodes that are the subject of a check. A real system, of course,  would probably just introduce one of these for any check node, but we might as well save ourselves some typing

I was thinking of something along these lines. If we go down this route, will it be sound if accidentally miss a  `isAccessPath` predicate for an access path?

Comment (markww) [reply src/analysis/souffle/taint.dl:147]: When we get to the point where we're automatically generating these rules, I think we're helped out by the fact that `check`s are structures explicitly given to us by our IR; it will be hard to forget to generate these check facts (or some successor to them containing a predicate). While we're hand-writing tests, though, I think it is valid to be concerned about forgetting to drop in these checks for non-leaf nodes. I can't immediately think of a good way to force us to not miss them, but I'll think on it.

Comment (bgogul) [reply src/analysis/souffle/taint.dl:147]: Perhaps, the simplest is to also do verify that `isAccessPath` is true as a part of the test case. We could force this by defining a macro?

File: src/analysis/souffle/taint.dl [hunk 3]
```diff
@@ -110,26 +110,68 @@ mayHaveTag(tgt, tag) :-
...
+isMemberOf(base, member) :-
+  isAccessPath(base),
+  isAccessPath(member),
+  strlen(base) + 1 < strlen(member),
+  cat(base, ".") = substr(member, 0, strlen(base) + 1).
+
 // An access path may have a tag if some subpath to a member field has that tag. This allows
 // checking for some inner node in the access path whether any leaf node of that node might have
 // a particular tag.
 mayHaveTag(base, tag) :-
```
aferr, line 171: @bgogul @markww 
I think I missed where this rule got added, but I think this rule is what the rest of this PR is about. We should not have this rule, because it is overly conservative. It essentially computes the join over the labels of the struct and applies this to the entire struct. A particle might only read a part of the struct rather than the entire struct. If it reads just a part of the struct, this will reject more programs than it should.

If we just track labels for completely unfolded paths rather than trying to compute labels for larger structures, the analysis should be more precise and secure. The analysis would also be simpler, it seems.

bgogul: > A particle might only read a part of the struct rather than the entire struct. If it reads just a part of the struct, this will reject more programs than it should.

This is valid point. However, won't we be able to control the precision by defining what `AccessPath` are valid for a particular handle. Suppose we have the following spec:
```
schema Foo 
  a: Text
  b: Text
  c: Text
}
particle FooWriter
  output: writes Foo

particle PartialFooReader
  input: reads Foo {a, b}

recipe R
  foo: create
  FooWriter:
    output: foo
  PartialFooReader
    input: foo
```

We will have the following relations:
```
  edge("output.a", "input.a")
  edge("output.b", "input.b")
 ```
  In particular, there will be no occurrence of `input.c`, and therefore, `isAccessPath("input.c")` will be `false`. Further, `mayHaveTag("input")` will only combine the labels for `input.a` and `input.b`.

WDYT?

Comment (aferr) [reply src/analysis/souffle/taint.dl:171]: Shouldn't the two particles expand to the following after we unfold them ?
```
particle FooWriter
     writes: {Foo.a, Foo.b, Foo.c }
  
particle PartialFooReader
     reads: {Foo.a, Foo.b}
```

And if we do this unfolding and the much simpler analysis (including the edges from `edges_from_graph.dl` along with the `isAccessPath(..)` predicates that come indirectly from `bindFlatParticle` and `isFlatParticle`) doesn't this also work? (The result should be that `Foo.a`, `Foo.b`, and `Foo.c` all have the same tags because FooWriter writes them all)

I thought the idea was to have two layers of IR: 1) is similar to arcs protobufs and includes nested structs, 2) has completely unfolded path names.

File: src/analysis/souffle/taint.dl (hunk 2 above)
Comment (markww) [reply src/analysis/souffle/taint.dl]: I've added `CHECK_TAG_PRESENT` and `CHECK_TAG_NOT_PRESENT` to do this.

File: src/analysis/souffle/taint.dl (hunk 3 above)
Comment (markww) [reply src/analysis/souffle/taint.dl:140]: For the Datalog, the plan is to jump right from the original Arcs manifest protos to Datalog facts written in terms of completely unfolded AccessPath names. We will perform the unfolding in the parser that translates from Arcs manifest protos to Datalog facts.

Early on, I wrote a PR that was more similar to what you described, basically directly mapping the proto structures into Datalog facts with a similar structure. @bgogul convinced me that this was not necessary: such complex structures are awkward to express and abstract from within Datalog facts and could be subject to a performance penalty due to Datalog's eager population of intermediate sets. We agreed that the the unfolding you describe should be done within the protobuf translator and that the Souffle datalog analysis should only push abstract state around on a dataflow graph described in Datalog `edge` and `claim` facts.

This PR assumes that edges will be drawn only between leaf fields and that the inner fields of an `AccessPath` will only be referenced in checks. This means that the taint can't flow up and then down the struct as you describe. While the taint flows up for serving checks upon inner nodes, it does not flow down because it does not need to propagate taint from edges connecting to inner nodes to their members, as edges to inner nodes do not exist.

Let me know if you find this argument convincing, or whether I've missed something.

Comment (bgogul) [reply src/analysis/souffle/taint.dl:140]: A few clarifications:

> Shouldn't the two particles expand to the following after we unfold them ?

Yes, they should expand, but it would happen at the level of the handle connections `input`, `foo`, and `output`.  That is, the edges would be as follows:

```
  edge("output.a", "foo.a")
  edge("output.b", "foo.b")
  edge("output.c", "foo.c")
  edge("foo.b", "foo.b")
  edge("foo.a", "input.a")
  edge("foo.b", "input.b")
```
I had missed the intermediate `foo` in the previous comment. Is this a source of confusion? 

> markww: This PR assumes that edges will be drawn only between leaf fields and that the inner fields of an AccessPath will only be referenced in checks. 

> aferr: If we just track labels for completely unfolded paths rather than trying to compute labels for larger structures, the analysis should be more precise and secure. The analysis would also be simpler, it seems.

It looks like we are aligned? The additional rule in this PR is about inferring a tag for `foo`, `input`, and `output` if needed for a `check` in the IR.  See https://github.com/PolymerLabs/arcs/blob/e7a79c2c885bd090f45c8fb9f8898ba8367f4907/javatests/arcs/core/analysis/testdata/fail_check_on_subpaths.arcs#L9 for an example. These tags would not be propagated automatically. 

> I thought the idea was to have two layers of IR: 1) is similar to arcs protobufs and includes nested structs, 2) has completely unfolded path names.

Yes, we would start with (1) as the input IR. Whether we have (2) as explicit IR is an implementation detail? Ultimately, we want to generate the `edge` relationships in datalog, right?

> For the Datalog, the plan is to jump right from the original Arcs manifest protos to Datalog facts written in terms of completely unfolded AccessPath names. We will perform the unfolding in the parser that translates from Arcs manifest protos to Datalog facts.

In the long run, we want to have an intermediate step, i.e., proto -> C++ data structures -> datalog facts/relations. For the time being, if it turns out to be straightforward to just iterate over the protos and generate the datalog facts directly, we just do that.

Comment (aferr) [reply src/analysis/souffle/taint.dl:140]: > Yes, we would start with (1) as the input IR. Whether we have (2) as explicit IR is an implementation detail? Ultimately, we want to generate the edge relationships in datalog, right?

> In the long run, we want to have an intermediate step, i.e., proto -> C++ data structures -> datalog facts/relations. For the time being, if it turns out to be straightforward to just iterate over the protos and generate the datalog facts directly, we just do that.

OK, yes this is just an implementation detail. It doesn't matter if we have these layers.

> See https://github.com/PolymerLabs/arcs/blob/e7a79c2c885bd090f45c8fb9f8898ba8367f4907/javatests/arcs/core/analysis/testdata/fail_check_on_subpaths.arcs#L9 for an example. These tags would not be propagated automatically.

OK. I see that this is needed for passing checks like that and that this code doesn't propagate inward to the leaves. The meaning of tags for outer structures that can't directly be read doesn't make too much sense to me, but I don't think it is harmful, so we can keep it.

Comment (aferr) [reply src/analysis/souffle/taint.dl:140]: Hope I was not too in the way!

**Review by aferr — APPROVED** [review]

File: src/analysis/souffle/taint.dl [hunk 4]
```diff
@@ -123,13 +123,24 @@ isTag(tag) :- mayHaveTag(_, tag).
 isTag(tag) :- ownsTag(_, tag).
 isTag(tag) :- saysDowngrades(_, _, tag).
 
+// A pair of (base, member) is in this set if the base access path is a non-trivial subpath of the
```
bgogul, line 126: Wait for #59  to be merged and move this code into `access_path.dl`?

src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
```diff
@@ -0,0 +1,55 @@
+#include "taint.dl"
+#include "fact_test_helper.dl"
+
+claimHasTag("R.foo.a", "userSelection").
+edge("R.foo.a", "R.bar.a").
+edge("R.foo.b", "R.bar.b").
+edge("R.foobaz.a", "R.bar.c").
```
bgogul, line 7: Are these needed for this test?

bgogul (approved):

src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
markww: They basically are just there to introduce access paths in a fashion that is more like what we might see in the wild. I can turn them into `isAccessPath` if that would make things more clear.

src/analysis/souffle/taint.dl
markww: Sure, can do.

src/analysis/souffle/taint.dl
bgogul: Of course not! Please feel free to bring things up for discussion.

src/analysis/souffle/taint.dl
bgogul: #59 is merged now.

src/analysis/souffle/taint.dl
markww: Yes, indeed. Program analysis is hard and subtle, better to have the discussion and decide it wasn't an issue than to miss something.

src/analysis/souffle/taint.dl
markww: Acknowledged.

src/analysis/souffle/tests/arcs_fact_tests/is_member_of_unit_test.dl
markww: Added a comment explaining the purpose of the edges and claims.

bgogul: Converting them into `isAccessPath` + adding a  comment is cleaner? Don't feel strongly about this though.

markww: Acknowledged, switched the test to do that instead.

PR #53, opened 2021-07-27
Make github workflow run all tests under 'src/...'.

bgogul:


markww:
This looks good. I didn't know about that particular syntax to run all tests, but it seems much more robust against human error. LGTM.

markww (approved):

PR #59, opened 2021-07-28
Reorganize datalog files for easier management and testing.

### LINKED ISSUE #47: Figure out a good way to organize the datalog files and relations
Even with such a small codebase, it is becoming harder to keep track of where the various relations are. We should come up with a good convention to organize the datalog files.

bgogul:
Closes #47


aferr (approved):
thanks, looks much better organized.

markww (approved):
Very nice!

PR #60, opened 2021-07-28
Fix import paths in the proto files and update CI.

bgogul:


markww (approved):
LGTM

markww (commented):
While you're altering this, do you want to see about fixing this warning I'm getting?

```
DEBUG: /usr/local/google/home/winterrowd/.cache/bazel/_bazel_winterrowd/9e65980e676170b53ccfa1ba73fad4ef/external/rules_proto_grpc/internal/compile.bzl:377:14: Inputs provided to 'deps' attr of target //third_party/arcs/proto:manifest_cc_proto_pb. Consider replacing with 'protos' attr to avoid transitive compilation. See https://rules-proto-grpc.aliddell.com/en/stable/transitivity.html
```

I think it involves moving the proto dependencies in `proto_library`'s `deps` list to a new `protos` list. But I'm not sure if that's right, it doesn't match the error message exactly. Feel free to ignore/punt if this seems like not necessary or not the point.

bgogul:
> While you're altering this, do you want to see about fixing this warning I'm getting?
> 
> ```
> DEBUG: /usr/local/google/home/winterrowd/.cache/bazel/_bazel_winterrowd/9e65980e676170b53ccfa1ba73fad4ef/external/rules_proto_grpc/internal/compile.bzl:377:14: Inputs provided to 'deps' attr of target //third_party/arcs/proto:manifest_cc_proto_pb. Consider replacing with 'protos' attr to avoid transitive compilation. See https://rules-proto-grpc.aliddell.com/en/stable/transitivity.html
> ```
> 
> I think it involves moving the proto dependencies in `cc_proto_library`'s (actually `rules_proto_grpc`'s `cpp_proto_library`'s, due to the alias in `native.oss.bzl`) `deps` list to a new `protos` list. But I'm not sure if that's right, it doesn't match the error message exactly. Feel free to ignore/punt if this seems like not necessary or not on-topic for this PR.

I filed #63 for now. I will take care of it in a separate PR.

PR #65, opened 2021-08-03
Fix issue where souffle parser/lexer files would not be found.

markww:
It turns out that this issue seems to have come from a file called BUILD
in third_party/souffle being used as an http_archive build_file. Because
it was called BUILD, this file was considered a BUILD file for a package
in the current repo, which caused issues on some machines. Possibly the
behavior differs on different versions of bazel. This commit fixes the
issue by naming the Souffle BUILD file BUILD.souffle and adding an empty
BUILD file to act as a package marker.


bgogul (approved):
That would have been painful to debug!

PR #66, opened 2021-08-03
Fix #63 by replacing deps with protos in cc_proto_library.

### LINKED ISSUE #63: Replace "deps" attrs for protobuf dependencies with "protos" attribute
In #60, @markww observed that they get the following warning: 
```
.../external/rules_proto_grpc/internal/compile.bzl:377:14: Inputs provided to 'deps' attr of target //third_party/arcs/proto:manifest_cc_proto_pb. Consider replacing with 'protos' attr to avoid transitive compilation. See https://rules-proto-grpc.aliddell.com/en/stable/transitivity.html
```

From @markww:
> I think it involves moving the proto dependencies in cc_proto_library's (actually rules_proto_grpc's cpp_proto_library's, due to the alias in native.oss.bzl) deps list to a new protos list. But I'm not sure if that's right, it doesn't match the error message exactly. Feel free to ignore/punt if this seems like not necessary or not on-topic for this PR.

Investigate and fix as needed.

markww:
Fix #63 by replacing deps with protos in cc_proto_library.


markww:
Note: merge after PR #65 .

third_party/arcs/proto/BUILD
```diff
@@ -36,5 +36,9 @@ proto_library(
 
 cc_proto_library(
     name = "manifest_cc_proto",
-    deps = [":manifest_proto"],
+    protos = [
```
bgogul, line 39: Do we need to enumerate all the protos? As discussed offline, I think what we are already doing the right thing. I believe the only issue is that we should not invoke `bazel build` on a `cc_proto_library` target directly (which we do in `main.yml`).

The suggested fix is to add a `cc_library` rule that depends on the `cc_proto_library` rule and use it:

```
cc_library(
  name = "manifest"
  deps = [":manifest_cc_proto"]
)
```
Can you check if that is the case?

bgogul (changes requested):

bgogul (approved):

third_party/arcs/proto/BUILD
markww: Checked. As we discussed offline, it turns out that this is a solution to a slightly different problem.

PR #68, opened 2021-08-12
Add the AccessPathSelectors, used to represent paths through type members while converting to Datalog.     

markww:
The AccessPath class represents a descent through the substructures of
    an Arcs Manifest, and is used for representing such paths when drawing
    edges and deciding which edges to draw (by, for instance, checking which
    partial AccessPaths match each other in ParticleSpec connections).
    
    The AccessPath is designed to be built bottom up. As such, the order of
    components are reversed internally, allowing for pushing on parent
    components as an access path is returned from leaf fields towards the
    containing Recipe.


markww:
This should only be merged after #67

src/transform/ir_proto_to_datalog/utils/access_path.h
```diff
@@ -0,0 +1,70 @@
...
+  // it is not empty and does not contain a dot separator.
+  static absl::Status CheckComponentAllowed(absl::string_view component);
+
+  // We store the path components in reverse order. We do this because we
+  // will be creating access paths by recursing down to the bottom of a
+  // manifest's tree and then returning up, collecting components as we go.
+  // Storing it as a vector in reverse order allows us to continually push
+  // the components onto the end of the vector in a single AccessPath that is
+  // moved from callee towards caller.
+  std::vector<absl::string_view> reverse_components_;
```
bgogul, line 65: It is OK to start with a vector of strings. However, in Arcs, we had much a somewhat richer structure. See https://github.com/PolymerLabs/arcs/blob/e7a79c2c885bd090f45c8fb9f8898ba8367f4907/java/arcs/core/data/AccessPath.kt#L21

I wonder if it is worth adding some structure here. How about defining something like a `Selector` and using `std:vector<Selector>`.

src/transform/ir_proto_to_datalog/utils/access_path.cc
```diff
@@ -0,0 +1,96 @@
...
+#include "absl/strings/str_cat.h"
+
+#include <string>
+
+namespace raksha::transform::utils {
+
+// Check the components of the two access paths for equality. First, ensure
+// that they have the same number of components, then check component by
+// component.
+bool AccessPath::operator==(const AccessPath &other) const {
```
bgogul, line 17: Won't the compiler generated implementation be sufficient?

src/transform/ir_proto_to_datalog/utils/access_path_test.cc
```diff
@@ -0,0 +1,159 @@
...
+#include "absl/status/statusor.h"
+#include "absl/strings/str_split.h"
+
+#include <string>
+
+namespace raksha::transform::utils {
+
+// Access paths are easier to read and write in their string form, so this is
+// very useful for creating readable test inputs.
+static absl::StatusOr<utils::AccessPath> make_access_path_from_string(
```
bgogul, line 13: Would it be worth adding this to core `AccessPath` data structure?

markww:
I split adding Abseil into its own commit and PR; as such, this PR now also depends upon #69 .

src/transform/ir_proto_to_datalog/utils/access_path.h
```diff
@@ -0,0 +1,70 @@
...
+  // Create a new AccessPath that is a parent of the parameter access_path by
+  // adding parent_component directly above the contents of access_path. If
+  // the string representation of access path is "a.b.c" and parent_component
+  // is "q", then the resulting access path would have the representation:
+  //
+  // "q.a.b.c"
+  //
+  // This will return an error status if
+  // check_component_allowed(parent_component) does not return Status::Ok.
+  static absl::StatusOr<AccessPath> CreateParent(
```
bgogul, line 45: Why `StatusOr`? If we have some structure to `AccessPath` as mentioned below, we could just use constructors directly?

bgogul (commented):
Left some high-level comments.

src/transform/ir_proto_to_datalog/utils/access_path.cc
markww: I think default `operator==` is a C++20 feature, and thus not available to us: https://en.cppreference.com/w/cpp/language/default_comparisons .

src/transform/ir_proto_to_datalog/utils/access_path_test.cc
markww: Perhaps. It does seem like it could be useful there, but I hadn't run into a use for it yet, so I was hesitant to add something to the implementation to serve the tests. Happy to move it if you think it would be better there though.

src/transform/ir_proto_to_datalog/utils/access_path.h
markww: Good point. That would also allow us to validate the names as part of the deserialization process, which is really when you would want to catch issues like this.

src/transform/ir_proto_to_datalog/utils/access_path.cc
markww: Simplified this by using `std::vector`'s `operator==`.

src/ir/access_path.cc
```diff
@@ -0,0 +1,86 @@
...
+
+#include <string>
+
+namespace raksha::ir {
+
+// Check the components of the two access paths for equality. This is the
+// same as their component vectors being equal according to std::vector's
+// operator==.
+bool AccessPath::operator==(const AccessPath &other) const {
+  return reverse_components_ == other.reverse_components_;
```
bgogul, line 18: You shouldn't even need this if all the fields of the class has a valid `==`

src/transform/ir_proto_to_datalog/utils/access_path.h
markww: Enriched the structure by defining a Selector, as you requested.

src/transform/ir_proto_to_datalog/utils/access_path.h
markww: Done.

src/ir/access_path.cc
markww: Are you sure? I think auto-generated C++ `operator==` is a C++20 thing. If I comment out my explicit implementation of `operator==`, my tests fail with compilation errors.

src/transform/ir_proto_to_datalog/utils/access_path_test.cc
markww: With the implementation of FieldSelector, it's clear that the logic here is a bit provincial to that particular selector. I'm going to hold off on moving this to the implementation until it's a bit more general.

src/ir/selector.cc
```diff
@@ -0,0 +1,29 @@
...
+  template<class T>
+  std::string operator()(const T &specific_selector) const {
+    return specific_selector.ToString();
+  }
+};
+
+// Apply the callable ToString struct to the specific_selector_ variant to
+// get the result of the specific_selector_'s ToString.
+std::string Selector::ToString() const {
+  return absl::visit(GenericSelectorToStringFunction(), specific_selector_);
```
aferr, line 20: Is it "better" to use absl::variant and absl::visit like you have here than to use inheritance polymorphism (i.e., define some class that all the specific types of selectors we have extend, and then just call toString (which would need to be virtual) on the specific_selector_)? Why? This is not a rhetorical question -- I am new to Abseil and just curious.

It's not easy for me to see that there is a performance gain by doing it this Abseil way: I would guess that the vtable lookup that inheritance would cause to happen would work similarly to visit anyway. Or is the point more to avoid problems that C++ has with inheritance in general (diamand problem.. others ? ) ?

src/ir/selector_access_path_test.cc
```diff
@@ -0,0 +1,140 @@
...
+#include <string>
+#include <vector>
+
+namespace raksha::ir {
+
+// Selector Access paths are easier to read and write in their string form, so
+// this is very useful for creating readable test inputs. This could possibly
+// be useful as a method in SelectorAccessPath itself, but, it's a bit
+// simplified right now, as it only splits on '.', the field separator
+// character. Also, I haven't found a use for it in non-test code yet.
```
aferr, line 19: This does seem useful. Maybe by the time we have the other kinds of fields it would make more sense to use a parser-generator (like yacc, antlr, etc.) to do this easily.

src/ir/selector_access_path_test.cc
```diff
@@ -0,0 +1,140 @@
...
+
+  auto selector_rev_iter = selector_strs.rbegin();
+
+  // Create a leaf selector from the first selector string.
+  SelectorAccessPath selector_access_path =
+      SelectorAccessPath(Selector(FieldSelector(*selector_rev_iter)));
+  ++selector_rev_iter;
+
+  // Add all others as parents.
+  for (; selector_rev_iter < selector_strs.rend(); ++selector_rev_iter) {
```
aferr, line 39: Is this kind of for loop better than the c++11 ones like `for(auto selector : selector_strs)` or is there a reason why this style of loop is needed? (Non-rhetorical though the C++11 style is more intuitive to me, where possible).

aferr (approved):

src/ir/selector.cc
markww: The reason to use `variant` in this case is not for performance reasons; I agree that performance gains would be minimal, if there were any at all.

The big reason why I'm using a variant here is to make it easy to answer where this `Selector` lives and for how long. Imagine that we're using inheritence-based polymorphism and are dealing with a specific selector implemented as an arbitrary base type `BaseSelector` deriving from `Selector`. In order to use inheritance and virtual function calls, I would have to have anything that wished to access this `BaseSelector` through a variable that may hold any `Selector` by pointing at the `BaseSelector` through a `Selector *` or `Selector &` pointer. This means that I'm introducing a burden upon myself and all future programmers to think about the lifetime of that pointer: how long does the pointed-to `Selector` live? Who owns it? Who may be pointing to it at any given time? Sometimes the answer to those questions are super easy, sometimes they are complicated, and sometimes they seem easy when they are complicated. But in many circumstances, a pointer is the right thing to use anyway, so we just use inheritance.

This is not one of those cases. At the current time, a `Selector` is created basically only to participate in a `SelectorAccessPath`. It is convenient to place it as a member of a `SelectorAccessPath` field of type `std::vector<Selector>`, whereupon `std::vector` manages all of the lifetime concerns for us. If we used inheritance, on the other hand, we'd have to make that field of type `std::vector<Selector *>` and handle the lifetime concerns ourselves. By using a `Selector` containing an `absl::variant`, we can just move `Selector`s around as values and completely ignore the lifetime concern.

src/ir/selector_access_path_test.cc
markww: There are two reasons why the regular range-based `for` syntax would not work here (or, at least, would not work without some additional mechanism which might look worse):

1. I'm treating the first element specially, using it to construct a leaf `SelectorAccessPath`. Of course, I could add an additional `bool first = true;` declaration outside the loop, use it to guard the stuff I'm doing before, and flip it to false after the first iteration.

2. I'm using the iterators returned by `rbegin()` and `rend()` to iterate through the vector in reverse order. The C++ range based for, of course, iterates over collections in standard order. You could imagine, of course, making a simple class that wraps a collection and has its `begin` and `end` be the inner collection's `rbegin` and `rend` respectively, but that might be more confusing than just using `rbegin` and `rend` directly.

The high level answer to your question: it's probably possible to use a range-based for in this example, but it would introduce its own less-standard readability challenges, so I'm not sure it's worth it.

src/ir/selector_access_path_test.cc
markww: Indeed. Once you start thinking about "how would you parse an array index or a tuple access or..." I become less confident that this current implementation would scale. As such, I'm leaving it as a toy implementation in a test for now and leaving building a real implementation for later.

src/ir/access_path_selectors.cc
```diff
@@ -0,0 +1,23 @@
...
+#include <string>
+
+#include "absl/strings/str_join.h"
+#include "src/ir/access_path_selectors.h"
+
+namespace raksha::ir {
+
+// We create the AccessPathSelectors string by just joining the contents of the
+// component vector in reverse, using '.' as a separator.
+std::string AccessPathSelectors::ToString() const {
```
bgogul, line 20: This could move into the header if you can use an inline lambda expression unless you plan to add more functionality to this file.

src/ir/access_path_selectors.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  // Use the default move constructor for AccessPathSelectors.
+  AccessPathSelectors(AccessPathSelectors &&other) = default;
+  AccessPathSelectors &operator=(AccessPathSelectors &&other) = default;
+
+  AccessPathSelectors(Selector parent_selector, AccessPathSelectors child_path)
+    : AccessPathSelectors(std::move(child_path))
+  {
+    reverse_selectors_.push_back(std::move(parent_selector));
```
bgogul, line 49: Use `emplace_back` instead of `push_back(std::move(...))`

src/ir/access_path_selectors.h
```diff
@@ -0,0 +1,91 @@
...
+  // together the string representations of the various selectors.
+  std::string ToString() const;
+
+  template<typename H>
+  friend H AbslHashValue(H h, const AccessPathSelectors &instance) {
+    return H::combine(std::move(h), instance.reverse_selectors_);
+  }
+
+  // Allow AccessPathSelectors to be copied explicitly.
+  AccessPathSelectors Copy() const {
```
bgogul, line 68: I am conflicted about `Copy` method. I think it would better to have the copy constructors so that the compiler has more information for optimizations?

src/ir/access_path_selectors.h
```diff
@@ -0,0 +1,91 @@
...
+    std::vector<Selector> new_reverse_selector_vec;
+    for (const Selector &selector : reverse_selectors_) {
```
bgogul, line 70: use `std::copy` or `std::reverse_copy` as needed?

src/ir/access_path_selectors_test.cc
```diff
@@ -0,0 +1,140 @@
...
+
+namespace raksha::ir {
+
+// Selector Access paths are easier to read and write in their string form, so
+// this is very useful for creating readable test inputs. This could possibly
+// be useful as a method in AccessPathSelectors itself, but, it's a bit
+// simplified right now, as it only splits on '.', the field separator
+// character. Also, I haven't found a use for it in non-test code yet.
+static absl::StatusOr<AccessPathSelectors> MakeSelectorAccessPathFromString(
+    const std::string str) {
```
bgogul, line 21: Isn't this a place where `const std::string&` is warranted to avoid a copy constructor invocation? Or, does simply having a `const` have the same effect?

src/ir/field_selector.h
```diff
@@ -0,0 +1,50 @@
...
+
+  // Allow moving the FieldSelector.
+  FieldSelector(FieldSelector &&other) = default;
+  FieldSelector &operator=(FieldSelector &&other) = default;
+
+  // Prints the string representing dereferencing the field in the AccessPath.
+  // This will just be the "." punctuation plus the name of the field.
+  std::string ToString() const {
+    return absl::StrCat(".", field_name_);
+  }
```
bgogul, line 24: Single line for better readability?

```suggestion
  std::string ToString() const { return absl::StrCat(".", field_name_); }
```

src/ir/field_selector.h
```diff
@@ -0,0 +1,50 @@
...
+    return field_name_ == other.field_name_;
+  }
+
+  template<typename H>
+  friend H AbslHashValue(H h, const FieldSelector &field_selector) {
+    return H::combine(std::move(h), field_selector.field_name_);
+  }
+
+  // Explicit copy method.
+  FieldSelector Copy() const {
```
bgogul, line 37: Related to the discussion about `Copy` and copy constructor in another thread. Based on the other discussion, we should update this as needed. 

Also, by making the copy constructor explicit as below, what is the advantage of a `Copy` function?

src/ir/field_selector_test.cc
```diff
@@ -0,0 +1,62 @@
+#include "src/ir/field_selector.h"
+
+#include <string>
+
+#include "absl/hash/hash_testing.h"
+#include "src/common/testing/gtest.h"
+
+namespace raksha::ir {
+
+static std::string sample_field_names[] = {
```
bgogul, line 10: Is this the idiomatic way to define parameters for `TestWithParam`?

src/ir/field_selector_test.cc
```diff
@@ -0,0 +1,62 @@
...
+    field_selectors_to_check.push_back(FieldSelector(field_name));
+  }
+  EXPECT_TRUE(
+      absl::VerifyTypeImplementsAbslHashCorrectly(field_selectors_to_check));
+}
+
+class FieldSelectorPairTest :
+   public ::testing::TestWithParam<::std::tuple<std::string, std::string>> {};
+
+TEST_P(FieldSelectorPairTest, FieldSelectorEqualsTest) {
```
bgogul, line 46: What is this testing exactly? Seems unnecessary?

src/util/copyable.h
```diff
@@ -0,0 +1,47 @@
+#ifndef SRC_UTIL_COPYABLE_H_
+#define SRC_UTIL_COPYABLE_H_
+
+#include <utility>
+
+namespace raksha::util {
+
+template<class T>
+class Copyable {
```
bgogul, line 9: What is the advantage of having this class? 

If you want to mark something as not copyable, you can simply delete the copy constructor with `= delete`? See https://en.cppreference.com/w/cpp/language/function#Deleted_functions

src/ir/access_path_selectors.h
```diff
@@ -0,0 +1,91 @@
...
+// except for move constructors. We define no "add", "insert", etc. methods;
+// "mutation" is done only by constructing a new object from a previous
+// object. That prevents many forms of state-manipulation bugs.
+//
+// 4. The name AccessPathSelectors is a bit more self-documenting.
+class AccessPathSelectors {
+ public:
+  // Create a leaf AccessPathSelectors from a single leaf selector.
+  explicit AccessPathSelectors(Selector leaf) {
+    reverse_selectors_.push_back(std::move(leaf));
```
bgogul, line 39: `emplace_back(...)` instead of `push_back(std::_move(...))`

src/ir/access_path_selectors.cc
```diff
@@ -0,0 +1,28 @@
...
+};
+
+// We create the AccessPathSelectors string by just joining the contents of the
+// component vector in reverse, using '.' as a separator.
+std::string AccessPathSelectors::ToString() const {
+  return absl::StrJoin(
+      reverse_selectors_.rbegin(),
+      reverse_selectors_.rend(),
+      "",
+      SelectorToStringFormatter());
```
bgogul, line 25: Can't you use an inline lambda?

src/ir/access_path_selectors_test.cc
```diff
@@ -0,0 +1,140 @@
...
+
+  const absl::StatusOr<AccessPathSelectors> access_path1 =
+      MakeSelectorAccessPathFromString(access_path_str1);
+  ASSERT_TRUE(access_path1.ok());
+
+  const absl::StatusOr<AccessPathSelectors> access_path2 =
+      MakeSelectorAccessPathFromString(access_path_str2);
+  ASSERT_TRUE(access_path2.ok());
+
+  ASSERT_EQ(*access_path1 == *access_path2, access_paths_strs_equal);
```
bgogul, line 66: Just to make sure I understand, are you checking here that the comparison on constructed access paths is the same as the strings?

bgogul (changes requested):
I am not sure if defining a `Copy` method is a good idea. I would like to understand the reasoning behind it.  Otherwise, the PR looks fine.

src/util/copyable.h
markww: The idea of this class is to make things that I have made not implicitly copyable through making the copy constructor explicit and/or private silently copyable while wrapped in this class. Making the copy functions private while having public move functions in classes has allowed the compiler to catch unintentional copies for me.

I'm currently using this because I want to make some canonical values to use as parameters to my unit tests and place them in a `const` array. Some of those values are classes for which I have made the copy constructor private. This makes those values ineligible to be moved. However, because they are tests, we don't care very much that they are making copies and so this provides a way to get those tests to work.

markww: @bgogul convinced me that this was undermining the compiler's ability to handle copying for us. Restored the default copy constructors.

src/ir/access_path_selectors.h
markww: Discussed offline; `push_back` is better in most circumstances due to the possibility of C++ implicit constructor calls.

src/ir/access_path_selectors.cc
markww: Done.

src/ir/access_path_selectors.h
markww: As discussed offline, `push_back` guards against unexpected behavior due to implicit constructors.

src/ir/access_path_selectors.h
markww: @bgogul convinced me offline that it's better to just let the C++ compiler do its job and we'll catch expensive, unnecessary copies in profiling.

src/ir/access_path_selectors.h
markww: Removed this whole method.

src/ir/field_selector.h
markww: Removed `Copy` as suggested.

src/ir/field_selector_test.cc
markww: From what I can see, it doesn't look that there is one prescribed, most-correct way to use `TestWtihParam`. This looks like a pattern that is used at least some of the time.

src/ir/field_selector_test.cc
markww: This tests that `FieldSelector`s are equal exactly when the field names from which they are built are equal.

src/ir/access_path_selectors_test.cc
markww: Yes. I'm checking that the AccessPaths that I generate from these strings compare equal only when the original strings compare equal. I've added a comment describing this above the test body.

src/ir/access_path_selectors_test.cc
markww: I've gotten rid of the const, which should make this more friendly to copy-elision.

src/ir/field_selector.h
markww: Done.

bgogul (approved):
Mostly style related comments.

src/ir/selector.cc
```diff
@@ -0,0 +1,29 @@
...
+}
```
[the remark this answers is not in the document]
markww: You are correct. I saw that template parameters in lambdas were a C++20 feature and didn't realize that using `auto` as a parameter would work in this circumstance.

src/ir/selector_test.cc
```diff
@@ -0,0 +1,94 @@
+#include "src/ir/selector.h"
+
+#include "absl/hash/hash_testing.h"
+#include "src/common/testing/gtest.h"
+
+namespace raksha::ir {
+
+// An enumeration of the different specific types of selectors that may
+// appear in a Selector.
+enum SelectorKind {
```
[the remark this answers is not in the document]
markww: Probably. The reason I did this was because I wanted to express my intention of the specific type independent of the bookkeeping `absl::variant` does in the implementation. That way, if I accidentally constructed the wrong specific type in an implementation (for instance) I wouldn't get an incorrect passing result.

src/ir/access_path_selectors_test.cc
```diff
@@ -0,0 +1,134 @@
...
+
+namespace raksha::ir {
+
+// Selector Access paths are easier to read and write in their string form, so
+// this is very useful for creating readable test inputs. This could possibly
+// be useful as a method in AccessPathSelectors itself, but, it's a bit
+// simplified right now, as it only splits on '.', the field separator
+// character. Also, I haven't found a use for it in non-test code yet.
+static absl::StatusOr<AccessPathSelectors> MakeSelectorAccessPathFromString(
+    std::string str) {
```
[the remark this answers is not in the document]
markww: In the body of this string, we split the input string into segments and copy those segments into `FieldSelector` objects. As such, it is currently more preferred to use a value parameter, as this allows for the possibility of copy elision at the callsite.

src/ir/access_path_selectors.h
```diff
@@ -0,0 +1,80 @@
...
+  // Turns this AccessPathSelectors into a string representation chaining
+  // together the string representations of the various selectors. Just
+  // concatenates together the string representation of all Selectors from
+  // parent selector to child selector (which is the reverse order of how
+  // they are stored internally).
+  std::string ToString() const {
+      return absl::StrJoin(
+      reverse_selectors_.rbegin(),
+      reverse_selectors_.rend(),
+      "",
```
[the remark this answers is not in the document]
markww: No. `FieldSelector` has the leading dot as part of its `ToString`. The idea is that the punctuation separating the various `Selector`s are a function of the specific type of `Selector` being used. For instance, we'd expect a array/container index selector to start with `[` instead of `.`. As such, the `.` is already in the string returned by `ToString`, and thus should use an empty string as the join delimiter.

src/ir/selector.cc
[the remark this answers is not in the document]
markww: Done.

src/ir/selector.cc
```diff
@@ -0,0 +1,29 @@
...
+
+// absl::variant has a pre-packaged operator equals that handles the dispatch
+// to the specific selector as we would expect.
+bool Selector::operator==(const Selector &other) const {
```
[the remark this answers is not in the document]
markww: Done.

src/ir/BUILD
```diff
@@ -0,0 +1,61 @@
...
+        exclude = ["*_test.cc"],
+    ),
+    hdrs = glob(["*.h"]),
+    linkopts = select({
+        "@bazel_tools//src/conditions:linux_x86_64": [
+            "-ldl",
+        ],
+        "//conditions:default": [
+        ],
+    }),
```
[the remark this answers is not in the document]
markww: I don't know. I deleted it and the build continues to work.

src/ir/BUILD
```diff
@@ -0,0 +1,61 @@
...
+    deps = [
+        "@absl//absl/container:flat_hash_set",
```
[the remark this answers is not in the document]
markww: Good catch, removed this (and a few other unused dependencies).

src/ir/access_path_selectors_test.cc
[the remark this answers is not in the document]
bgogul: Is specifying value parameter applicable to all methods or just constructors? I thought it was the later?

PR #70, opened 2021-08-17
Add SelectorAccessPathSet for building SelectorAccessPaths bottom up.

markww:


src/ir/selector_access_path.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "absl/strings/str_join.h"
+
+#include <string>
+
+namespace raksha::ir {
+
+// Check the components of the two access paths for equality. This is the
+// same as their component vectors being equal according to std::vector's
+// operator==.
+bool SelectorAccessPath::operator==(const SelectorAccessPath &other) const {
```
bgogul, line 13: inline this implementation in header file itself?

src/ir/selector_access_path.h
```diff
@@ -0,0 +1,85 @@
+#ifndef SRC_IR_SELECTOR_ACCESS_PATH_H_
+#define SRC_IR_SELECTOR_ACCESS_PATH_H_
+
+#include "selector.h"
```
bgogul, line 4: `src/ir/selector_access_path.h`

src/ir/selector_access_path.cc
```diff
@@ -0,0 +1,30 @@
...
+  return reverse_selectors_ == other.reverse_selectors_;
+}
+
+// We create the SelectorAccessPath string by just joining the contents of the
+// component vector in reverse, using '.' as a separator.
+std::string SelectorAccessPath::ToString() const {
+  std::vector<std::string> reverse_selector_strs;
```
bgogul, line 20: I think you can pass in a `Formatter` object and avoid constructing this temporary vector?

src/ir/field_selector.cc
```diff
@@ -0,0 +1,13 @@
+#include "src/ir/field_selector.h"
+
+#include "absl/strings/str_cat.h"
+
+#include <string>
+
+namespace raksha::ir {
+
+std::string FieldSelector::ToString() const {
```
bgogul, line 9: Such simple functions could move into the header itself?

src/ir/selector.h
```diff
@@ -0,0 +1,45 @@
...
+
+  template<typename H>
+  friend H AbslHashValue(H h, const Selector &selector) {
+    return H::combine(std::move(h), selector.specific_selector_);
+  }
+
+ private:
+  // The variant storing the specific type of selector. We will dispatch
+  // through this to perform all methods on the generic Selector.
+  absl::variant<FieldSelector> specific_selector_;
```
bgogul, line 40: Should we use inheritance instead of a variant? The types used in a variant has restrictions, right?

src/ir/selector_access_path.h
```diff
@@ -0,0 +1,85 @@
...
+// object owning variable-length memory in a bottom-up fashion, copies could
+// cause us to have quadratic performance.
+//
+// 3. Providing our own interface allows us to make this class immutable
+// except for move constructors. We define no "add", "insert", etc. methods;
+// "mutation" is done only by constructing a new object from a previous
+// object. That prevents many forms of state-manipulation bugs.
+//
+// 4. The name SelectorAccessPath is a bit more self-documenting.
+class SelectorAccessPath {
```
bgogul, line 35: `AccessPathSelectors` is a better name?

src/ir/selector_access_path.h
```diff
@@ -0,0 +1,85 @@
...
+ public:
+  // Create a leaf SelectorAccessPath from a single leaf selector.
+  explicit SelectorAccessPath(const Selector leaf) : reverse_selectors_{leaf} {}
```
bgogul, line 38: Should this be `const Selector&& leaf`?

src/ir/selector_access_path_set.cc
```diff
@@ -0,0 +1,70 @@
+#include "selector_access_path_set.h"
+#include "selector.h"
+
+namespace raksha::ir {
+
+void SelectorAccessPathSet::AddAll(SelectorAccessPathSet other) {
```
bgogul, line 6: Should the parameter be `&` or `&&`?

bgogul (commented):
Had mostly stylistic questions.

src/ir/selector.h
markww: I don't think there's any restrictions on what types may be used in an `absl::variant` other than that they must be able to be instantiated and they must be able to handle the calls dispatched to by `absl::visit`.

I'm using `absl::variant` so that I can use `Selector`s as values instead of as pointers; see the conversation with @aferr at https://github.com/google-research/raksha/pull/68#discussion_r690311522

src/ir/selector_access_path.h
markww: I'll make the copy constructors private on `Selector`s just to make sure I'm doing this right. In C++17, copy elision is much more powerful and you can use value types in many cases where you had to use `&&` before. But I might need to make the parameter not `const` and use `std::move` in that use of `leaf`.

src/ir/selector_access_path_set.cc
markww: Here's why I did it this way:

`SelectorAccessPathSet` can only be silently moved. It can be copied, but only with an explicit call to `Copy`.

If the parameter `other` is populated with a move, passing it by value is very cheap. If it is populated with a copy, then this frees us from having to worry about whether the pointee of the reference is handled in a way that makes mutating it safe. As such, this is no less performant than a reference parameter but also easier to reason about.

src/ir/selector_access_path.h
markww: Sure, I'll change it.

src/ir/selector.h
```diff
@@ -0,0 +1,45 @@
...
+namespace raksha::ir {
+
+// An individual selector in an AccessPath descending down the type tree.
+// This Selector is generic, and may contain any method of descending, such
+// as field access, array access, tuple element access, etc.
+//
+// For now, however, we only have implemented the FieldSelector portion.
+class Selector {
+ public:
+  explicit Selector(const FieldSelector field_selector)
```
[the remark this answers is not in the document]
markww: I think this is actually a case where I want to drop the `const` and use `std::move`. I'll make the copy constructors private to make sure I'm doing it right.

src/ir/field_selector_test.cc
```diff
@@ -0,0 +1,63 @@
+#include "field_selector.h"
```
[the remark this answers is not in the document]
markww: Addressed in #68

src/ir/field_selector.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <string>
+
+namespace raksha::ir {
+
+// Represents descending through the type tree by taking the field of an
+// EntityType in an AccessPath.
+class FieldSelector {
+ public:
+  FieldSelector(const std::string field_name) : field_name_(field_name) {}
```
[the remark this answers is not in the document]
markww: Addressed in #68

src/ir/field_selector.h
```diff
@@ -0,0 +1,35 @@
...
+  // This will just be the "." punctuation plus the name of the field.
+  std::string ToString() const;
+
+  // Two fields selectors are equal exactly when their names are equal.
+  bool operator==(const FieldSelector &other) const {
+    return field_name_ == other.field_name_;
+  }
+
+  template<typename H>
+  friend H AbslHashValue(H h, const FieldSelector &field_selector) {
```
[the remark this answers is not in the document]
markww: Yes, this allows it to be hashable by Absl's hashing infrastructure.

src/ir/selector.h
```diff
@@ -0,0 +1,45 @@
+#ifndef SRC_IR_SELECTOR_H_
+#define SRC_IR_SELECTOR_H_
+
+#include "field_selector.h"
```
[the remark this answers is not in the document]
markww: Addressed in #68

src/ir/selector.h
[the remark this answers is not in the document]
markww: Addressed in #68

src/ir/selector_access_path.cc
markww: Addressed in #68

src/ir/selector_access_path.h
markww: Addressed in #68.

src/ir/field_selector.cc
markww: Addressed in #68.

src/ir/selector_access_path.h
markww: Addressed in #68.

src/ir/selector_access_path.h
markww: Addressed in #68.

src/ir/selector_access_path.cc
markww: Addressed in #68.

src/ir/selector_access_path_set.cc
markww: As in #68, I've stripped out these explicit Copy calls and let the compiler do the work.

src/ir/BUILD
```diff
@@ -21,8 +22,22 @@ cc_test(
         ":ir",
         "//src/common/testing:gtest",
         "@absl//absl/hash",
+        "@absl//absl/hash:hash_testing",
```
bgogul, line 25: Not related to this CL. I wonder if we should gather all the `@absl` targets in a BUILD target.

src/ir/access_path_selectors_set.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+ private:
+  // The inner vector implementing the set. It is not guaranteed that the
+  // contents of this vector are unique (although, due to the invariants of
+  // the type tree, we suspect they may well be). It is not guaranteed (and
+  // not expected) that the contents of this vector are sorted. We can get
+  // away with this only because we do not allow access to the contents of
+  // this set until it is requested that we turn it into a flat_hash_set.
+  std::vector<AccessPathSelectors> inner_vec_;
```
bgogul, line 86: Make this a list if you don't need random access?

src/ir/access_path_selectors_set.h
```diff
@@ -0,0 +1,91 @@
...
+      }
+    }
+
+    return result;
+  }
+
+  // Returns a set that has the same elements as child_set but with
+  // parent_selector added as a parent to each of them.
+  static AccessPathSelectorsSet AddParentToAll(
+      Selector parent_selector, AccessPathSelectorsSet child_set) {
```
bgogul, line 58: I still am not getting used to the missing `const &` in the arguments. I believe we are making a copy of both the arguments in the function body?

src/ir/access_path_selectors_set.h
```diff
@@ -0,0 +1,91 @@
...
+        result.inner_vec_.push_back(std::move(path));
+      }
+    }
+
+    return result;
+  }
+
+  // Returns a set that has the same elements as child_set but with
+  // parent_selector added as a parent to each of them.
+  static AccessPathSelectorsSet AddParentToAll(
```
bgogul, line 57: Why not make this a constructor? Having the `AddParentToAll` states the intent clearly. Perhaps, it is better left as a static member?

bgogul (approved):

src/ir/access_path_selectors_set.h
markww: This was exactly my thought process. You could consider this analogous to a factory function or a constructor. It's not necessarily clear from context, however, what constructing an `AccessPathSelectorsSet` from an existing `AccessPathSelectorsSet` and a single `Selector` means. The name helps clarify that.

Because `AddParentToAll` this is small and simple, I've placed the implementation in the header. This guarantees that it will be inlined into each compilation unit that uses it and probably into most functions that use it. That gives the optimizer a lot of material to work with, so I'm not worried about it being not a constructor

src/ir/access_path_selectors_set.h
markww: What kind of list did you have in mind? I chose `std::vector` because it's common, simple to reason about, and usually pretty performant due to its contiguous memory backing. `std::list` and `std::forward_list` are doubly and singly linked lists, respectively; they tend to be more unusual to encounter in C++ code and would probably be slower for reading due to the additional pointer dereferences.

src/ir/BUILD
markww: Perhaps. I doubt we're saving much in link time/binary size/whatever by keeping them separate, but are definitely encountering some review friction due to them. If we do this, we should probably make two targets, `absl` and `absl_testing`, so we don't mix our test and production dependencies. But having two libraries instead of many would be a big improvement.

markww: Filed #85 about this.

src/ir/access_path_selectors_set.h
markww: I forgot to use `std::move` on `parent_selector`, so I should do that. I think I'm doing the right thing on the loop over the `child_set` elements by moving each element temporary.

src/ir/access_path_selectors_set.h
bgogul: I was thinking `std::list` mainly prompted by the copy that I saw in `union`.  A splice would be more efficient there. Other than that it looks like we only ever iterate over all the elements forwards or backwards, which is also efficient?

I have to admit that I did not think if the reads would be slower. Do you think there will be additional pointer dereferences in vector vs list when iterating?

src/ir/access_path_selectors_set.h
markww: Actually, I don't think moving `parent_selector` is the right thing. I'm putting `parent_selector` as the parent of each element of `child_set`; therefore, if `child_set` has N elements, I *do* need to make at least `N-1` copies of `parent_selector`. Therefore, `std::move` is probably not the right thing here, as we need the name to be attached to its value through each iteration of the loop.

src/ir/access_path_selectors_set.h
markww: I think I'm going to just go with `std::vector` for now, but open to changing this in a follow-up PR if there's a good reason to switch to a different inner collection.

PR #71, opened 2021-08-17
Deserialize manifest protos

markww:


src/transform/ir_proto_to_datalog/arcs_manifest_tree/recipe.cc
```diff
@@ -0,0 +1,75 @@
...
+
+#include <memory>
+#include <string>
+#include <cstdint>
+
+namespace raksha::transform::arcs_manifest_tree {
+
+namespace logging = ::raksha::transform::logging;
+
+Recipe::Recipe(const absl::string_view name, const uint64_t num_handles)
```
bgogul, line 16: I don't think we should be replicating the arcs manifest as is in C++. We want to introduce a level of abstraction like graph (which is that the recipe is).

src/transform/ir_proto_to_datalog/logging/context_logger.h
```diff
@@ -0,0 +1,62 @@
+#ifndef SRC_TRANSFORM_IR_PROTO_TO_DATALOG_LOGGING_CONTEXT_LOGGER_H_
+#define SRC_TRANSFORM_IR_PROTO_TO_DATALOG_LOGGING_CONTEXT_LOGGER_H_
+
+#include "absl/strings/string_view.h"
```
bgogul, line 4: Can't we use https://github.com/google/glog?

bgogul (commented):
As I mentioned in the comments, we don't want a 1-1 mapping from everything in `manifest.proto` to C++ data structure. Let us chat a bit more offline.

src/transform/ir_proto_to_datalog/logging/context_logger.h
markww: Thank you for this pointer, I will use that library.

markww:
This PR was never destined for submission directly; it was always a bit of a scrapheap to dig through for usable bits of code for more polished and focused PRs. I think, at this point, it's entirely picked over and it's time to close it.

PR #82, opened 2021-08-20
Add glog, the Google logger, to Raksha.

markww:


bgogul (approved):
Thanks for pulling this in!

PR #83, opened 2021-08-21
Move AccessPathSelectors factory CreateFromString to class, allow empty.

markww:
Making an AccessPathSelectors object from its string representation is
something that seems potentially useful in the future, so this PR moves
it from being a test utility to being a class member. In addition,
AccessPathSelectors is changed slightly to allow both the empty string
and an empty AccessPathSelectors.


markww:
Closing this because, at the least, this needs to get broken up into two pull requests and I'd rather just start with new ones since there's no conversation here yet.

PR #86, opened 2021-08-23
Add first few Type classes and a GetAccessPaths in each.

markww:
Add the PrimitiveType and EntityType classes and an interface Type
from which both inherit. Also add a method GetAccessPaths which produces
the SelectorAccessPathsSet describing the SelectorAccessPaths possible
through that type.

Also, make gtest.h pull in the gmock header, as it is required for more
advanced matchers, some of which we use in the tests generated for the
access path collecting functions.


src/transform/ir_proto_to_datalog/arcs_manifest_tree/BUILD
```diff
@@ -0,0 +1,40 @@
+cc_library(
+    name = "arcs_manifest_tree_lib",
```
bgogul, line 2: Simply `arcs_manifest_tree`.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/entity_type.h
```diff
@@ -0,0 +1,27 @@
...
+#include "absl/container/flat_hash_set.h"
+#include "src/transform/ir_proto_to_datalog/arcs_manifest_tree/schema.h"
+#include "src/transform/ir_proto_to_datalog/arcs_manifest_tree/type.h"
+#include "src/ir/access_path_selectors.h"
+#include "src/ir/access_path_selectors_set.h"
+#include "third_party/arcs/proto/manifest.pb.h"
+
+namespace raksha::transform::arcs_manifest_tree {
+
+class EntityType : public Type {
```
bgogul, line 13: I expect some notion of  `Type` will be useful in ir. Can we move the type related classes to `src/ir` itself?

src/transform/ir_proto_to_datalog/arcs_manifest_tree/entity_type.h
```diff
@@ -0,0 +1,27 @@
...
+ public:
+  explicit EntityType(Schema schema) : schema_(std::move(schema)) {}
+
+  raksha::ir::AccessPathSelectorsSet GetAccessPaths() const {
```
bgogul, line 17: Name says `GetAccessPaths`, but returns `AccessPathSelectorsSet`? Is `AccessPathSelectorsSet` incorrectly named?

bgogul (commented):
These type-related classes could be part of src/ir itself. I don't think the eventual ir types will be that much different from these classes for type. Lets us move it to `raksha::ir` namespace itself.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/type_test.cc
```diff
@@ -0,0 +1,231 @@
...
+  return std::make_unique<EntityType>(
+      MakeAnonymousSchema(std::move(field_map)));
+}
+
+// Given a vector of strings representing access paths to leaves, generate a
+// type having all of those access paths and no others.
+// The resulting type will not have information about the name of the type;
+// this is not presented by the selector access path and thus cannot be
+// derived from it.
+template<typename Iter>
```
bgogul, line 42: It seems like a very complicated logic to use for tests. It also makes the test harder to read. Let us leave this for now. However, I think we need a better mechanism to write these tests. I can think of a few options:
 - Use the `.arcs` format to write schemas, parse them as protos, and use in tests. (This requires the code to deserialize the `TypeProto` to `Type`).
 - Use something like json to express the nesting of fields and use that to generate the desired type. 
 - Create something like `class FieldMap(std::map<std::string, std::unique_ptr<FieldMap>)` with appropriate factory methods to make it easy to create types. 

The first option might be reasonable, but the only issue is that you will need to create `.arcs` files instead of using inline strings.  Can you file a bug?

src/transform/ir_proto_to_datalog/arcs_manifest_tree/type.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include "absl/container/flat_hash_set.h"
+#include "src/ir/access_path_selectors_set.h"
+#include "third_party/arcs/proto/manifest.pb.h"
+
+namespace raksha::transform::arcs_manifest_tree {
+
+class Type {
+ public:
+  virtual ~Type() {}
```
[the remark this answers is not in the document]
markww: We can't use `absl::variant` here because `Type`s recursively nest within `Type`s. `absl::variant` acts like a `union` with an `enum` describing which variant we are currently using. Because `EntityType`s contain an embedded map of field names to field `Type`s, using `absl::variant` turns into asking the compiler to embed a `Type` within a `Type`. This is impossible, as C++ class objects are of fixed width with fields at fixed offsets from the top of the object. Therefore, we have to introduce a pointer to point at the nested type.

Because we have to introduce a pointer, we have to deal with lifetime concerns anyway. The benefits of `absl::variant` have disappeared, and we can more cheaply and correctly use regular ol' C++ inheritance.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/BUILD
markww: Moved, so this is now `types`.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/entity_type.h
markww: Done.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/type_test.cc
markww: We could also use inline `.textproto` strings. I've done some of that in the follow-up PR #87. Perhaps these should just be moved to the test infrastructure I've put up there and we can get rid of these complex helper methods.

markww: Issue opened as #88.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/schema.h
```diff
@@ -0,0 +1,29 @@
...
+#include "third_party/arcs/proto/manifest.pb.h"
+
+namespace raksha::transform::arcs_manifest_tree {
+
+class Schema {
+ public:
+  explicit Schema(
+    absl::optional<std::string> name,
+    absl::flat_hash_map<std::string, std::unique_ptr<Type>> fields)
+    : name_(name), fields_(std::move(fields)) {}
```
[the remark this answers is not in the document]
markww: Added, done.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/type_test.cc
```diff
@@ -0,0 +1,231 @@
...
+static std::vector<std::string> GetAccessPathStrVecFromAccessPathSelectorsSet(
+    raksha::ir::AccessPathSelectorsSet access_path_set) {
+  absl::flat_hash_set<raksha::ir::AccessPathSelectors> absl_access_path_set =
+      raksha::ir::AccessPathSelectorsSet::CreateAbslSet(
+          std::move(access_path_set));
+  std::vector<std::string> result_strs;
+  for (raksha::ir::AccessPathSelectors path : absl_access_path_set) {
+    result_strs.push_back(path.ToString());
+  }
+  return result_strs;
```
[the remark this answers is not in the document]
markww: I don't think that will work here; I'm not concatenating the strings together, I'm adding them all to a `std::vector` of `AccessPathSelectors` strings. I would love if there was a `map` function in C++ that would take in a collection and produce that same collection with the elements transformed, but I haven't seen one with a nice interface yet.

src/transform/ir_proto_to_datalog/arcs_manifest_tree/entity_type.h
markww: I think for now I'll go the other way, and rename the method to `GetAccessPathSelectorsSet`.

markww: Done.

bgogul (approved):

PR #89, opened 2021-08-25
Add HandleConnectionSpec class and methods sending it to/from protos.

markww:
We only really need this at this point to link the name to
a HandleConnection and to determine whether it reads and/or writes.


src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
```diff
@@ -0,0 +1,59 @@
...
+    return name_;
+  }
+
+  bool reads() const {
+    return reads_;
+  }
+
+  bool writes() const {
+    return writes_;
+  }
```
bgogul, line 32:
```suggestion
  const std::string &name() const {   return name_;  }

  bool reads() const {   return reads_;  }

  bool writes() const {  return writes_;  }
```

bgogul (approved):

PR #95, opened 2021-08-27
Add TagClaim class; add structured representation of AccessPathRoot.

markww:
These two classes both represent a claim upon a particular path. The
difference is that the UnrootedTagClaim claims a tag is on a particular
AccessPathSelectors object, while TagClaim is a claim upon a full
AccessPath.

Why have two separate classes? The reason is that a ParticleSpec
contains claims that are rooted at the top of the ParticleSpec rather
than what Raksha considers the "true" root: the top of the Recipe. The
UnrootedAccessPath is the data that we derive from the ParticleSpec,
which is then instantiated at the various particles by associating it
with a root string. The resulting TagClaim is a fully-formed Datalog
fact, ready to be printed out.


src/ir/tag_claim.h
```diff
@@ -0,0 +1,47 @@
...
+ public:
+  // Create a tag claim from an arcs Assume proto. Despite the fact
+  // that the access path indicated by a claim indicates a root, it's not
+  // rooted deeply enough for us; the root goes only to the particle spec name,
+  // whereas we'd like to go all the way to the recipe name. Therefore, the
+  // resulting TagClaim has a SpecAccessPathRoot.
+  static TagClaim CreateFromProto(
+      const arcs::ClaimProto_Assume &assume_proto);
+
+  static TagClaim Instantiate(
```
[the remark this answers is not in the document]
markww: Done

src/ir/access_path_root.h
```diff
@@ -0,0 +1,85 @@
...
+// distinguishing between AccessPaths that have been constructed relative to
+// some abstract structure (such as a ParticleSpec) and those that have been
+// constructed relative to a concrete structure (such as a Particle).
+
+namespace raksha::ir {
+
+// A SpecAccessPathRoot describes the uninstantiated case. An attempt to
+// print it to a string will fail and it indicates that it is not yet
+// instantiated.
+class SpecAccessPathRoot {
```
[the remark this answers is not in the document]
markww: Done. Now I just keep around all of the information that's in the proto for HandleConnetionSpec roots.

src/ir/access_path_root_test.cc
```diff
@@ -0,0 +1,31 @@
...
+TEST_P(ConcreteAccessPathRootTest, ConcreteAccessPathRootTest) {
+  const std::string &root_string = GetParam();
+  ConcreteAccessPathRoot concrete_access_path_root(root_string);
+  AccessPathRoot test_access_path_root(concrete_access_path_root);
+  EXPECT_TRUE(test_access_path_root.IsInstantiated());
+  EXPECT_EQ(test_access_path_root.ToString(), root_string);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    ConcreateAccessPathRootTest, ConcreteAccessPathRootTest,
```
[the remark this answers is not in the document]
markww: Done.

markww:
I know you approved this already @bgogul , but I think it has changed enough to merit a second look.

src/ir/access_path_root.h
```diff
@@ -0,0 +1,137 @@
...
+// we have only two types of roots: a HandleConnectionSpecAccessPathRoot and
+// a HandleConnectionAccessPathRoot, describing a HandleConnectionSpec in a
+// ParticleSpec and a Handle in a Particle, respectively.
+namespace raksha::ir {
+
+// A HandleConnectionSpecAccessPathRoot describes the root as a
+// HandleConnectionSpec in a ParticleSpec. We consider this uninstantiated as
+// ParticleSpecs are abstract Particles which may be instantiated in multiple
+// locations.
+class HandleConnectionSpecAccessPathRoot {
```
bgogul, line 19: OK for now, but we might want to think about non-Arcs centric name.

src/ir/access_path_root.h
```diff
@@ -0,0 +1,137 @@
...
+      << "Expected a HandleRoot message to have a non-empty particle_spec.";
+    std::string handle_connection_spec_name =
+        handle_root_proto.handle_connection();
+    CHECK(!handle_connection_spec_name.empty())
+      << "Expected a HandleRoot message to have a non-empty handle_connection.";
+    return HandleConnectionSpecAccessPathRoot(
+        particle_spec_name, handle_connection_spec_name);
+  }
+
+  explicit HandleConnectionSpecAccessPathRoot(
```
bgogul, line 34: `explicit` is redundant, but OK if you prefer to leave it?

src/ir/access_path_root.h
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  const std::string &handle_connection_spec_name() const {
+    return handle_connection_spec_name_;
+  }
+
+ private:
+  std::string particle_spec_name_;
+  std::string handle_connection_spec_name_;
+
```
bgogul, line 71: redundant white space?

bgogul (approved):
Update PR description as well?

src/ir/access_path_root.h
markww: Indeed. I want to generalize most of this stuff, but I want to study at least one other system before I do to get a better sense of what abstraction to use.

src/ir/access_path_root.h
markww: Yeah, I try to keep in the habit of using `explicit` unless I want an implicit constructor, even in the cases where it's not necessary. Mostly so that I'm safe if I drop a parameter and forget to update the qualifier.

src/ir/access_path_root.h
markww: Good catch, removed.

PR #96, opened 2021-08-30
Add TagCheck class and add class for sharing code btw Claim and Check.

markww:
This commit introduces the TagCheck class. This class will eventually
represent a check of some arbitrary binary predicate involving tags
upon some AccessPath. For now, though, it represents a check that
a particular tag is present on an AccessPath.

Despite being semantically very different, the internal representation
of a TagClaim and a TagCheck are actually very similar: both annotate
a tag onto an AccessPath. The difference is merely that the claim
introduces the tag while the check asserts the tag. This similarity in
representation, however, allows us to use the new
TagAnnotationOnAccessPath to share code between the two classes.


markww:
This PR should only be merged after #95 .

bgogul:
@markww can you change the diff base of this PR to the branch corresponding to #95? That will help see only the additional changes.

markww:
Sorry, just changed the base. Will try to do better on that in the future.

bgogul (approved):

PR #100, opened 2021-08-31
Add type information and GetAccessPaths to HandleConnectionSpec.

markww:
This PR adds type information to HandleConnectionSpec, preserving the
information that was already in the manifest.proto. Initially, I thought
I wouldn't need this because my quick-and-dirty partial implementation
of the proto transformer looked like it did not need it. However, that
was more top-down than the cleaner approach has ended up being, and the
type information was necessary for building the ParticleSpec.

This PR also adds a method GetAccessPaths to HandleConnectionSpec, which
returns a vector containing all of the full AccessPaths that can be
accessed through the HandleConnectionSpec and end in a leaf field of the
type on HandleConnectionSpec. This is useful for building the edges
within a ParticleSpec.

This PR also adds operator== to the AccessPathRoots and some additional
field accessors to AccessPath. operator== on roots will eventually
become important for instantiating particle specs and was useful for
testing HandleConnectionSpec's new methods, so

### DIFF
--- a/src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
+++ b/src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
@@ -4,14 +4,14 @@
 #include <string>
 
 #include "src/common/logging/logging.h"
+#include "src/ir/access_path.h"
+#include "src/ir/types/type.h"
 #include "third_party/arcs/proto/manifest.pb.h"
 
 namespace raksha::xform_to_datalog::arcs_manifest_tree {
 
 // A class used to gather information about HandleConnectionSpecs from the
-// equivalent ArcsManifestProtos. Currently, the only thing we care about
-// from these is the name of the handle connection and whether it reads
-// and/or writes the handle to which it is connected.
+// equivalent ArcsManifestProtos.
 class HandleConnectionSpec {
  public:
   // Factory creating a HandleConnectionSpec from an arcs
@@ -19,11 +19,13 @@ class HandleConnectionSpec {
   static HandleConnectionSpec CreateFromProto(
       const arcs::HandleConnectionSpecProto &handle_connection_spec_proto);
 
-  const std::string &name() const {   return name_;  }
+  const std::string &name() const { return name_; }
 
-  bool reads() const {   return reads_;  }
+  bool reads() const { return reads_; }
 
-  bool writes() const {  return writes_;  }
+  bool writes() const { return writes_; }
+
+  const raksha::ir::types::Type &type() const { return *type_; }
 
   // Make an arcs HandleConnectionSpecProto from this object.
   arcs::HandleConnectionSpecProto MakeProto() const {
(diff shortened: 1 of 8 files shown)


src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
```diff
@@ -34,18 +36,45 @@ class HandleConnectionSpec {
...
+    ir::AccessPathSelectorsSet selectors_set =
+        type_->GetAccessPathSelectorsSet();
+    ir::AccessPathRoot root(
+        ir::HandleConnectionSpecAccessPathRoot(
+            std::move(particle_spec_name), name_));
+    for (ir::AccessPathSelectors selectors :
+         ir::AccessPathSelectorsSet::CreateAbslSet(selectors_set)) {
+      result_paths.push_back(ir::AccessPath(root, std::move(selectors)));
+    }
+    return result_paths;
```
bgogul, line 58: I recall seeing similar code snippets elsewhere. It looks like there is an opportunity for refactoring this into a helper method. (Not necessary for this PR, though.)

src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
```diff
@@ -34,18 +36,45 @@ class HandleConnectionSpec {
...
     return result;
   }
 
+  // Get all of the AccessPaths rooted at the associated ParticleSpec
+  // (indicated through the provided name) through leaf fields of type_.
+  std::vector<raksha::ir::AccessPath> GetAccessPaths(
+      std::string particle_spec_name) const {
+    namespace ir = raksha::ir;
+    std::vector<ir::AccessPath> result_paths;
+    ir::AccessPathSelectorsSet selectors_set =
```
bgogul, line 49: Is a copy needed here?

src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec_test.cc
```diff
@@ -38,18 +38,105 @@ TEST_P(RoundTripHandleConnectionSpecProtoTest,
...
 static const ProtoStringAndExpectations string_and_expections_array[] = {
-    { .proto_str = "name: \"foo\" direction: READS", .expected_name = "foo",
-      .expected_reads = true, .expected_writes = false},
-    { .proto_str = "name: \"bar\" direction: WRITES", .expected_name = "bar",
-      .expected_reads = false, .expected_writes = true },
-    { .proto_str = "name: \"baz\" direction: READS_WRITES", .expected_name =
-    "baz", .expected_reads = true, .expected_writes = true}
+    { .proto_str = "name: \"foo\" direction: READS type: { primitive: TEXT }",
+      .expected_name = "foo", .expected_reads = true, .expected_writes = false},
+    { .proto_str = "name: \"bar\" direction: WRITES type: { primitive: TEXT }",
```
bgogul, line 49: May be have a different primitive type?

bgogul (approved):

src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
markww: Agreed, I'll be looking for cleanup opportunities after MVP.

src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec.h
markww: There isn't actually a copy that happens here. `GetAccessPathSelectorsSet()` returns an `AccessPathSelectorsSet` that lives in a local by value, so inside the function the copy is elided. At the callsite (here), we are just providing a local name to which we can attach the value. As such, if I understand C++17 correctly, no copies occur.

src/xform_to_datalog/arcs_manifest_tree/handle_connection_spec_test.cc
markww: I have only implemented the `TEXT` primitive type at this point; no other primitive type will round trip correctly at this time.

PR #103, opened 2021-09-01
Add ParticleSpec class; add == operators for component pieces.

markww:
Add the ParticleSpec class, which contains the claims, checks, and edges
within a ParticleSpec in an uninstantiated form.

Also add a number of operator== functions on AccessPaths, Edges,
TagChecks, and TagClaims to allow testing this class.

### DIFF
--- a/src/ir/tag_check.h
+++ b/src/ir/tag_check.h
@@ -37,6 +37,11 @@ class TagCheck {
         "\").\n");
   }
 
+  bool operator==(const TagCheck &other) const {
+    return tag_annotation_on_access_path_ ==
+      other.tag_annotation_on_access_path_;
+  }
+
  private:
   // Internally, we represent the data as a TagAnnotationOnAccessPath object.
   // This allows us to share code with TagClaim. While TagCheck and TagClaim
--- a/src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
+++ b/src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
@@ -1,0 +1,80 @@
+#include "src/xform_to_datalog/arcs_manifest_tree/particle_spec.h"
+
+namespace raksha::xform_to_datalog::arcs_manifest_tree {
+
+namespace ir = raksha::ir;
+
+ParticleSpec ParticleSpec::CreateFromProto(
+    const arcs::ParticleSpecProto &particle_spec_proto) {
+  std::string name = particle_spec_proto.name();
+  CHECK(!name.empty()) << "Expected particle spec to have a name.";
+
+  std::vector<HandleConnectionSpec> handle_connection_specs;
+  for (const arcs::HandleConnectionSpecProto &hcs_proto :
+      particle_spec_proto.connections()) {
+    handle_connection_specs.push_back(
+        HandleConnectionSpec::CreateFromProto(hcs_proto));
+  }
+
+  std::vector<ir::TagClaim> claims;
+  for (const arcs::ClaimProto &claim : particle_spec_proto.claims()) {
+    CHECK(!claim.has_derives_from())
+      << "DerivesFrom claims not yet implemented";
+    CHECK(claim.has_assume()) << "Expected claim to carry an Assume field.";
+    claims.push_back(ir::TagClaim::CreateFromProto(claim.assume()));
+  }
+
+  std::vector<ir::TagCheck> checks;
+  for (const arcs::CheckProto &check : particle_spec_proto.checks()) {
+    checks.push_back(ir::TagCheck::CreateFromProto(check));
+  }
+
+  std::vector<ir::Edge> edges =
+      GenerateEdges(name, std::move(handle_connection_specs));
+
+  return ParticleSpec(
+      std::move(name), std::move(checks), std::move(claims), std::move(edges));
+}
+
+std::vector<ir::Edge> ParticleSpec::GenerateEdges(
+    std::string particle_spec_name,
+    std::vector<HandleConnectionSpec> connection_specs) {
+  // First, find all of the access paths that are input to the particle spec
+  // and output from the particle spec.
+  std::vector<ir::AccessPath> input_access_paths;
+  std::vector<ir::AccessPath> output_access_paths;
+  for (HandleConnectionSpec &connection_spec : connection_specs) {
+    std::vector<ir::AccessPath> access_paths =
+        connection_spec.GetAccessPaths(particle_spec_name);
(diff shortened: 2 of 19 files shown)


src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -0,0 +1,66 @@
...
+  // The checks this ParticleSpec performs upon its AccessPaths. All such
+  // checks are made upon uninstantiated AccessPaths.
+  std::vector<raksha::ir::TagCheck> checks_;
+  // The claims this ParticleSpec performs upon its AccessPaths. All such
+  // claims are made upon uninstantiated AccessPaths.
+  std::vector<raksha::ir::TagClaim> claims_;
+  // The dataflow edges within this ParticleSpec connecting its
+  // HandleConnectionSpecs. These edges are all between uninstantiated
+  // AccessPaths.
+  std::vector<raksha::ir::Edge> edges_;
```
bgogul, line 61: Does this take into account `derives from` claims in the spec?

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -0,0 +1,66 @@
...
+      std::vector<raksha::ir::Edge> edges)
+      : name_(std::move(name)), checks_(std::move(checks)),
+        claims_(std::move(claims)), edges_(std::move(edges)) { }
+
+  // Generate the edges between HandleConnectionSpecs within this ParticleSpec.
+  static std::vector<raksha::ir::Edge> GenerateEdges(
+      std::string particle_spec_name,
+      std::vector<HandleConnectionSpec> connection_specs);
+
+  // The name of this ParticleSpec.
```
bgogul, line 50: I think you might also want to capture the inputs and outputs of the particle spec. OK to defer it to when the need arises.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  // Now that we have populated all of the access paths, draw the edges.
+  // TODO(#99): Right now, we draw edges from all input access paths to all
+  // output access paths. This is quadratic; we should draw all edges without
+  // explicit DerivesFrom claims to and from a nexus AccessPath to produce a
+  // linear number of edges instead.
+  std::vector<ir::Edge> edges;
+  for (ir::AccessPath input : input_access_paths) {
```
bgogul, line 72: `const &` here and in the nested loop?

src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
```diff
@@ -0,0 +1,80 @@
...
+
+  std::vector<ir::Edge> edges =
+      GenerateEdges(name, std::move(handle_connection_specs));
+
+  return ParticleSpec(
+      std::move(name), std::move(checks), std::move(claims), std::move(edges));
+}
+
+std::vector<ir::Edge> ParticleSpec::GenerateEdges(
+    std::string particle_spec_name,
```
bgogul, line 40: Just an observation: I still cannot get used to value parameters.

src/ir/tag_check.h
```diff
@@ -37,6 +37,11 @@ class TagCheck {
         "\").\n");
   }
 
+  bool operator==(const TagCheck &other) const {
```
bgogul, line 40: Why weren't these needed earlier?

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -0,0 +1,66 @@
...
+  std::string name_;
+  // The checks this ParticleSpec performs upon its AccessPaths. All such
+  // checks are made upon uninstantiated AccessPaths.
+  std::vector<raksha::ir::TagCheck> checks_;
```
bgogul, line 54: `AccessPath` is redundantly replicated in so many places. We should canonicalize it somehow. Can you file a bug?

bgogul (approved):

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: At the moment, we ignore `derives from` claims. However, it would be relatively easy to extend `GenerateEdges` to take `derives from` claims into account while generating these edges.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: Sounds good. One of the reasons I'm deferring it is because, as written, `HandleConnectionSpec`s contain a `std::unique_ptr<Type>`, which makes them a bit harder to work with. That said, you can imagine a refactoring in which `HandleConnectionSpec` just performs the generation of its `AccessPath` set eagerly, allowing for holding the `unique_ptr` for an even briefer period of time.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
markww: You're right, that would be better. This was the site of my `std::move` misadventure; apparently I missed switching the iterator variable to a reference when fixing that.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
markww: All of my old C++ instincts scream "You're making an unnecessary copy! Why would you do that?". I'm pretty unnerved as well.

src/ir/tag_check.h
markww: We're not currently putting in any effort to force these things to be unique. Stating an equivalent `TagCheck` multiple times in the Datalog is idempotent, similar with `Claim`, `Edge`, etc. As such, we are just putting our "sets" of facts into `std::vector`s, which don't require hash and `==`. In addition, I think the parser is probably preventing cases of non-uniqueness from coming up. This might be useful in production code later, though.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: Opened #106 .

src/xform_to_datalog/arcs_manifest_tree/particle_spec.cc
markww: Done.

PR #104, opened 2021-09-01
Add BulkInstantiate, which instantiates structures using map entries.

markww:


src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -13,6 +13,14 @@
 
 namespace raksha::xform_to_datalog::arcs_manifest_tree {
 
+// This struct is used to return the members of a ParticleSpec after being
+// instantiated.
+struct InstantiatedParticleSpecFacts {
```
bgogul, line 18: Why can't we use `ParticleSpec` itself?

bgogul (approved):

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: I thought about that. Here's why I was thinking it made more sense to return this structure: `ParticleSpec`, named as it is, is inherently a specification. Mentally, it is somewhat strange to contemplate something called a `ParticleSpec` containing instantiated facts.

Another, more radical, way to do this would be to change `ParticleSpec` to `Particle` and have `Edge`, `TagClaim`, and `TagCheck` be uninstantiated on their own. Then, we could have an `Instantiated<T>` "smart pointer" that associates each of these structures with a pointer to the `instantiation_map`. The method for printing datalog fact strings is defined upon `Instantiated` and each class itself contains only a private method for printing a datalog fact given an instantiation map. Each of these structures will then instantiate itself on the fly given this map for printing to datalog.

The other thing I like about this is that it exposes the distinction between the instantiated and uninstantiated state to the type system, allowing us to shift the potential issue of printing an uninstantiated fact from runtime to compile time. That said, this is a deeper refactoring, so I'm not sure it makes sense at the moment.

bgogul: SG

Issue #107, opened 2021-09-03
Break up traversal of Arcs manifest protos and test for traversal of Arcs manifest protos.


markww:
To help get the MVP across the finish line, I wrote the portion of Raksha that traverses Arcs instantiation structures (`Recipe`s, `Particle`s, `HandleConnection`s, etc) as one big ugly function. In order to test this one big ugly function, I wrote one big ugly test that contains a rather unweildy Arcs manifest textproto string. We should break up the traversal of the Arcs manifest code into separate classes and the giant manifest proto test into separate tests, or replace it with a custom IR and delete the code entirely.

markww:
We have deprecated and are in the process of deleting the Arcs frontend. I believe the classes described here are already deleted and the protos are on their way out. Closing.

PR #108, opened 2021-09-03
Add DatalogFacts class, which gathers facts from Manifest proto.

### LINKED ISSUE #31: Convert IR to datalog relations

### LINKED ISSUE #74: Generate datalog facts related to dataflow edges from manifest.proto
Let us not define in memory data structures for the protos in manifest proto. It should be straightforward to just iterate over the protos themselves for the MVP.

markww:
When the DatalogFacts class is constructed from a ManifestProto, it
contains all of the objects that describe the facts needed by the
dataflow graph that are implied by that ManifestProto. DatalogFacts also
contains a ToString method which prints all of the facts out to
a string.

Note that this is pretty quick and dirty to move towards an MVP; we
should clean this up as described in issue #107.

This produces serviceable edge facts, and thus fixes #74 . I believe this also fixes #31 .


src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+// #TODO(#107): In the interest of prototyping speed, I wrote out the
+//  entirety of the traversal of instantiation structures as one big
+//  traversal. This works, but it is hard to read and hard to test. We should
+//  break this up.
+DatalogFacts DatalogFacts::CreateFromManifestProto(
+    const arcs::ManifestProto &manifest_proto) {
+  std::vector<ir::TagClaim> result_claims;
+  std::vector<ir::TagCheck> result_checks;
+  std::vector<ir::Edge> result_edges;
+  absl::flat_hash_map<std::string, amt::ParticleSpec> particle_specs;
```
bgogul, line 39: Document the purpose of the loops?

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+  for (const arcs::ParticleSpecProto &particle_spec_proto :
+    manifest_proto.particle_specs()) {
+    amt::ParticleSpec particle_spec =
+        amt::ParticleSpec::CreateFromProto(particle_spec_proto);
+    std::string particle_spec_name = particle_spec.name();
+    auto ins_res = particle_specs.insert({
+      std::move(particle_spec_name), std::move (particle_spec) });
```
bgogul, line 46: redundant space after `std::move`.

Also, is `std::move` even required?

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+#include "absl/container/flat_hash_map.h"
+#include "src/xform_to_datalog/datalog_facts.h"
+
+#include "src/xform_to_datalog/arcs_manifest_tree/particle_spec.h"
+
+namespace raksha::xform_to_datalog {
+
+namespace ir = raksha::ir;
+namespace types = raksha::ir::types;
+namespace amt = raksha::xform_to_datalog::arcs_manifest_tree;
```
bgogul, line 26: `amt` -> `arcs`

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+        raksha::ir::HandleConnectionAccessPathRoot
+          instantiated_handle_root(
+              recipe_name, particle_name, std::move(handle_spec_name));
+        raksha::ir::HandleAccessPathRoot handle_root(
+            recipe_name, std::move(handle_name));
+        CHECK(connection_proto.has_type())
+          << "Handle connection with absent type not allowed.";
+        std::unique_ptr<types::Type> connection_type =
+            types::Type::CreateFromProto(connection_proto.type());
+        ir::AccessPathSelectorsSet access_path_selectors_set =
```
bgogul, line 88: `const &`?

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+            connection_type->GetAccessPathSelectorsSet();
+
+        // For each Selector path from a root to a leaf, draw dataflow edges
+        // to and from the AccessPaths rooted at the handle at at the
```
bgogul, line 92: `at the handle of the`

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+        // particle's handle connection. In the case that the handle
+        // connection is input only, the edge out will be unreachable from
+        // the rest of the graph. In the case that the connection is output
+        // only, the edge in will be unreachable from the rest of the graph.
```
bgogul, line 96: Not sure I understand these comments. Why would they be unreachable?

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+        // In particular, the dataflow graph remains correct, if not as
+        // minimal as it could be.
+        for (const ir::AccessPathSelectors &selectors :
+             ir::AccessPathSelectorsSet::CreateAbslSet(
+                 access_path_selectors_set)) {
+          ir::AccessPath handle_access_path(
+              ir::AccessPathRoot(handle_root), selectors);
+          ir::AccessPath handle_connection_access_path(
+              ir::AccessPathRoot(instantiated_handle_root), selectors);
+          result_edges.push_back(
```
bgogul, line 110: We should add these only if this is a read/readwrite connection?

src/xform_to_datalog/datalog_facts.cc
```diff
@@ -0,0 +1,149 @@
...
+              ir::Edge(handle_access_path, handle_connection_access_path));
+          result_edges.push_back(
```
bgogul, line 112: We should add these only if this is a write/readwrite connection?

src/xform_to_datalog/datalog_facts.h
```diff
@@ -0,0 +1,76 @@
...
+    return claims_;
+  }
+
+  const std::vector<raksha::ir::TagCheck> checks() const {
+    return checks_;
+  }
+
+  const std::vector<raksha::ir::Edge> edges() const {
+    return edges_;
+  }
```
bgogul, line 66: All of these can be on a single line for better readability.

src/xform_to_datalog/datalog_facts.h
```diff
@@ -0,0 +1,76 @@
...
+    return absl::StrCat(
+        "// Claims:\n",
+        absl::StrJoin(claims_, "", tostring_formatter),
+        "\n// Checks:\n",
+        absl::StrJoin(checks_, "", tostring_formatter),
+        "\n// Edges:\n",
+        absl::StrJoin(edges_, "", tostring_formatter));
+  }
+
+  const std::vector<raksha::ir::TagClaim> claims() const {
```
bgogul, line 56: `const&` here and below?

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -0,0 +1,492 @@
...
+    { DatalogFacts(
+        { ir::TagClaim(ir::TagAnnotationOnAccessPath(
+            kHandleConnectionOutAccessPath, "tag")) },
+        { ir::TagCheck(ir::TagAnnotationOnAccessPath(
+            kHandleConnectionInAccessPath, "tag2")) },
+        { ir::Edge(kHandleH1AccessPath, kHandleConnectionInAccessPath),
+          ir::Edge(kHandleConnectionInAccessPath,
+                   kHandleConnectionOutAccessPath),
+           ir::Edge(kHandleConnectionOutAccessPath, kHandleH2AccessPath)}),
+      "// Claims:\nclaimHasTag(\"recipe.particle.out\", \"tag\").\n"
```
bgogul, line 67: Use rawstrings for better readability of the tests: https://abseil.io/tips/64

Also, can you file a bug to move all the other tests to use raw strings wherever possible as well?

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -0,0 +1,492 @@
...
+
+// Create a manifest textproto to test constructing DatalogFacts from
+// a ManifestProto. The ParticleSpecs will be pretty simple, as we have
+// tested creating ParticleSpecs from ParticleSpecProtos in more depth
+// elsewhere. We're focusing more on connecting these to particles here.
+//
+// TODO(#107): This is currently quite complex. After the MVP, we should break
+// this up into smaller, more focused tests.
+static const std::string kManifestTextproto =
+    "particle_specs: [\n"
```
bgogul, line 85: Use raw string?

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -0,0 +1,492 @@
...
+};
+
+static std::string MakeEdgeStr(std::string ap1, std::string ap2) {
+  return absl::StrCat("edge(\"", ap1, "\", \"", ap2, "\").\n");
+}
+
+static const std::string kExpectedEdgeStrings[] = {
+    // Named recipe edges:
+    // Edges connecting h1 to NamedR.PS1#0 for fields {field1, field2}.
+    MakeEdgeStr(kNamedRecipeH1Field1, kNamedRecipePs1Inst0InHandleField1),
```
bgogul, line 378: Having the strings explicitly written as raw strings would be much more readable, I think.

bgogul (changes requested):

src/xform_to_datalog/datalog_facts.cc
markww: The issue with doing that is that it causes a name collision: the Arcs manifest proto library already introduces an `arcs` namespace, in which proto classes such as `ParticleSpecProto` live.

src/xform_to_datalog/datalog_facts.cc
markww: `std::move` is possibly necessary here to force the use of a move instead of a copy. It is only guaranteed that a copy will be elided if the name is no longer accessible after the copy (not no longer accessed, but no longer accessible). This is the last usage of `particle_spec` and `particle_spec_name`, but both remain accessible through the end of the loop, which has the additional statement(s) starting with the `CHECK` on the next line.

src/xform_to_datalog/datalog_facts.cc
markww: Not necessary; `GetAccessPathSelectorsSet` always returns a copy-elided `AccessPathSelectorsSet`. That means that we can just use the value instead of a const reference. That is helpful because we don't need to worry about the lifetime of the set.

src/xform_to_datalog/datalog_facts_test.cc
markww: Bug filed as: https://github.com/google-research/raksha/issues/109

src/xform_to_datalog/datalog_facts.cc
markww: Because of how we construct the edges within the `ParticleSpec`. Within the `ParticleSpec`, where the `HandleConnectionSpec` is directly accessible, we shall draw edges that start at a particular `HandleConnectionSpec` only if that `HandleConnectionSpec` is `read`, and we will draw edges that end at a `HandleConnectionSpec` only if that `HandleConnectionSpec` is `write`. That means that we can draw bidirectional edges between a `Handle` and a `HandleConnection` and rely upon the edge construction within a `ParticleSpec` to effectively filter the unnecessary input or output edge.

src/xform_to_datalog/datalog_facts.cc
markww: Done.

src/xform_to_datalog/datalog_facts.cc
markww: Done.

src/xform_to_datalog/datalog_facts.cc
markww: Fixed redundant space.

src/xform_to_datalog/datalog_facts.cc
markww: I've updated the comment here; let me know if you buy my argument.

src/xform_to_datalog/datalog_facts.h
markww: Good catch. Done.

src/xform_to_datalog/datalog_facts.h
markww: Done.

src/xform_to_datalog/datalog_facts_test.cc
markww: Done.

src/xform_to_datalog/datalog_facts_test.cc
markww: Done.

src/xform_to_datalog/datalog_facts_test.cc
markww: Done.

src/xform_to_datalog/datalog_facts.cc
markww: As we discussed over video, while my reasoning is correct, it's not at all trivial to prove it and it produces more runtime overhead to produce these extra edges. I'm going to fix this to only produce the edges required.

markww: Done.

src/xform_to_datalog/datalog_facts.cc
markww: Done.

src/xform_to_datalog/datalog_facts.cc
markww: Done.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -62,14 +71,22 @@ class ParticleSpec {
       std::string name,
       std::vector<raksha::ir::TagCheck> checks,
       std::vector<raksha::ir::TagClaim> claims,
-      std::vector<raksha::ir::Edge> edges)
+      std::vector<HandleConnectionSpec> handle_connection_specs)
       : name_(std::move(name)), checks_(std::move(checks)),
-        claims_(std::move(claims)), edges_(std::move(edges)) { }
+        claims_(std::move(claims)) {
+    for (HandleConnectionSpec &handle_connection_spec :
```
bgogul, line 77: Is this tested?

src/xform_to_datalog/datalog_facts_test.cc
bgogul: No changes here? Did you forget to push the changes to the branch?

In any case, I meant that instead of `MakeEdgeStr(...)` let us have the strings like `"edge(...)"` here.

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -0,0 +1,495 @@
...
+  }
+
+  std::vector<std::string> edge_datalog_strings;
+  for (const ir::Edge &edge : datalog_facts.edges()) {
+    edge_datalog_strings.push_back(edge.ToString());
+  }
+  EXPECT_THAT(
+      edge_datalog_strings,
+      testing::UnorderedElementsAreArray(kExpectedEdgeStrings));
+}
```
bgogul, line 493: nit: Better as separate tests with a test fixture.

bgogul (approved):
Some stylistic suggestions and nits.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: Yes. All of the tests in `particle_spec_test.cc` use a textproto string to provide the ParticleSpec test input. `ParticleSpec::CreateFromProto` uses this constructor to build a `ParticleSpec` after constructing the input collections from the proto, and thus exercises this constructor.

src/xform_to_datalog/datalog_facts_test.cc
markww: Sorry, I misunderstood what you meant. I thought you wanted to have the format string within the helper function be written as a raw string.

I went back and forth on this a lot. I wanted to prevent a situation in which a typo in a commonly-used component of these strings caused frustration in searching for the result. However, I think I may have made the wrong call: reading the expectations readily is probably more important. I'll switch as you suggest.

markww: Alright, I pushed a commit that should actually do what you asked for here. Sorry about the confusion!

src/xform_to_datalog/datalog_facts_test.cc
markww: Done.

markww:
I've resolved all comments here and will merge after tests pass if there are no other comments. Note that some of the test output is not as pretty as I'd like, filed #110 about fixing that.

Issue #109, opened 2021-09-08
Use raw strings to make test inputs more readable.


markww:
Many of the Raksha test inputs contain strings of some formatted language (usually textproto). Some of these strings contain a fair amount of escaped characters, which makes them hard to read. We should use [raw strings](https://abseil.io/tips/64) to make these test inputs more readable.

PR #112, opened 2021-09-09
Switch to using raw strings for textproto in particle_spec_test.cc

markww:
Use raw strings for textprotos in particle_spec_test.cc. This makes
progress on issue #109, but does not fix it.


markww:
On the timing of this PR: I know switching to raw strings isn't high priority, but I felt the urge to copy-paste some existing textproto in writing tests for `ParticleSpec`'s use of the `DerivesFromClaim` class. I decided it was best to switch this test file over to raw strings now before I multiplied the amount of work I had to do.

bgogul (approved):

Issue #77, opened 2021-08-18
Add support for derivesFrom claims in datalog generation from manifest.proto


bgogul:
The support would mean changing the generation of intra-particle dataflow edge generation.

Issue #58, opened 2021-07-28
Create C++ data structures to represent the IR


bgogul:
I don't think this should not be a direct 1-to-1 translation from manifest.proto to C++. The C++ data structures should be at a higher abstraction level.

bgogul:
https://github.com/google-research/raksha/pull/51 from @aferr is related.

PR #113, opened 2021-09-09
Add new DerivesFromClaim, alter edge drawing in ParticleSpec to use it.

markww:
The DerivesFromClaim claims that, within some ParticleSpec, an output
AccessPath derives from an input AccessPath. A DerivesFromClaim both
acts as an explicitly-specified dataflow edge and a request to exclude
the target AccessPath from participating in the default process that
draws all input handle connections to all output handle connections.

This commit also adds a hash function to AccessPath (which was not
needed before) and tests it.

In addition, because we now have two different kinds of claim, this
renames claim fields to tag_claim in all cases except those in
DatalogFacts. We refrained from doing this there to avoid disrupting
@bgogul's work on an overarching binary to produce datalog facts.

### DIFF
--- a/src/ir/BUILD
+++ b/src/ir/BUILD
@@ -76,6 +76,7 @@ cc_test(
     deps = [
         ":ir",
         "//src/common/testing:gtest",
+        "@absl//absl/hash:hash_testing",
     ],
 )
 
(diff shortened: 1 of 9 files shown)


markww:
This PR fixes #77.

markww:
Also, by implementing the last remaining IR structure we need for the MVP, I believe it also fixes #58.

bgogul (approved):
Nice!

PR #114, opened 2021-09-09
Add a `checkHasTag` relation to the taint.dl.

bgogul:
This is generated by `DatalogFacts`, but is not declared in taint.dl. No tests added as the existing tests will make sure `taint.dl` is compilable with this change.

### DIFF
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -38,6 +38,9 @@
 // There is a claim that `accessPath` definitely has `tag` taint.
 .decl claimHasTag(accessPath: AccessPath, tag: Tag)
 
+// Check that `accessPath` definitely has `tag` taint.
+.decl checkHasTag(accessPath: AccessPath, tag: Tag)
+
 //-----------------------------------------------------------------------------
 // Rules
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/analysis/souffle/taint.dl
```diff
@@ -78,4 +81,6 @@ mayHaveTag(base, tag) :-
   mayHaveTag(member, tag),
   isMemberOf(base, member).
 
+.output checkHasTag
```
markww, line 84:
```suggestion
.output checkHasTag(IO=stdout, delimiter=",")
```

The reason why I suggest the above is that the default behavior by Souffle is to print the output to a file named similarly to the output relation in the current directory. I find that behavior rather surprising, especially when I've run Souffle in interpreter mode to get a sense of what it's doing when I have a test failure. I think that outputting to `stdout` is generally much less surprising; I also think that the `,` delimiter is much more readable than a `<tab>`.

markww (approved):

src/analysis/souffle/taint.dl
bgogul: You can dump to stdout by passing `-D-` on the souffle command line. Specifying `","` as a delimiter is a good idea, but I am going to abandon this PR completely as we should probably do something along the lines of https://github.com/google-research/raksha/blob/main/src/analysis/souffle/fact_test_helper.dl

bgogul:
Closing this PR as we should do something along the lines of https://github.com/google-research/raksha/blob/main/src/analysis/souffle/fact_test_helper.dl 

Moreover, `checkHasTag` is analogous to `mayHaveTag`.

PR #115, opened 2021-09-14
Generate a conditional `checkHasTag` rule for a checks in a spec.

bgogul:
Currently, a check in a spec just asserts `checkHasTag`. Instead, it should generate `checkHasTag(...) : mayHaveTag(...)`. Otherwise, the checks get trivially satisfied.

### DIFF
--- a/src/xform_to_datalog/datalog_facts_test.cc
+++ b/src/xform_to_datalog/datalog_facts_test.cc
@@ -79,7 +79,7 @@ static std::tuple<DatalogFacts, std::string>
 claimHasTag("recipe.particle.out", "tag").
 
 // Checks:
-checkHasTag("recipe.particle.in", "tag2").
+checkHasTag("recipe.particle.in", "tag2") :- mayHaveTag("recipe.particle.in", "tag2").
 
 // Edges:
 edge("recipe.h1", "recipe.particle.in").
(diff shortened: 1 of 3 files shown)


src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -295,12 +295,18 @@ TEST_F(ParseBigManifestTest, ManifestProtoClaimsTest) {
...
 
 static std::string kExpectedCheckStrings[] = {
-    R"(checkHasTag("NamedR.PS1#0.in_handle.field1", "tag2").)",
-    R"(checkHasTag("NamedR.PS1#1.in_handle.field1", "tag2").)",
-    R"(checkHasTag("NamedR.PS2#2.in_handle.field1", "tag4").)",
-    R"(checkHasTag("GENERATED_RECIPE_NAME0.PS1#0.in_handle.field1", "tag2").)",
-    R"(checkHasTag("GENERATED_RECIPE_NAME0.PS2#1.in_handle.field1", "tag4").)",
-    R"(checkHasTag("GENERATED_RECIPE_NAME0.PS2#2.in_handle.field1", "tag4").)",
+    R"(checkHasTag("NamedR.PS1#0.in_handle.field1", "tag2") :-)"
+    R"( mayHaveTag("NamedR.PS1#0.in_handle.field1", "tag2").)",
```
markww, line 299: Rather than two raw strings which are concatenated together, it would be more readable to have one raw string that exceeds the 80 character limit. Going over the 80 character limit is allowed for raw strings.

markww (approved):

src/xform_to_datalog/datalog_facts_test.cc
bgogul: Done.

PR #117, opened 2021-09-14
Create a tool to generate datalog facts from manifest proto.

bgogul:
Creates a simple commmand line tool to generate the datalog program that will be used for analysis.

### DIFF
--- a/src/xform_to_datalog/datalog_facts.h
+++ b/src/xform_to_datalog/datalog_facts.h
@@ -28,22 +28,29 @@
 
 namespace raksha::xform_to_datalog {
 
+// A utility class that is used to generate a datalog program by combining
+// datalog facts from different sources and adding the necessary headers.
 class DatalogFacts {
  public:
   DatalogFacts(ManifestDatalogFacts manifest_datalog_facts)
       : manifest_datalog_facts_(std::move(manifest_datalog_facts)) {}
 
-  // Print out all contained facts as a single datalog string. Note: this
-  // does not contain the header files that would be necessary to run this
-  // against the datalog scripts; it contains only facts and comments.
+  // Returns the datalog program with necessary headers.
   std::string ToString() const {
-    return absl::StrCat(
-        "// GENERATED FILE, DO NOT EDIT!\n\n", "#include \"taint.dl\"\n",
-        manifest_datalog_facts_.ToString(), "\n.output checkHasTag\n");
+    return absl::StrCat(kDatalogFilePrefix, manifest_datalog_facts_.ToString());
   }
 
  private:
   ManifestDatalogFacts manifest_datalog_facts_;
+
+  // The prefix that should be added to the datalog program.
+  static constexpr char kDatalogFilePrefix[] =
+      R"(// GENERATED FILE, DO NOT EDIT!
+
+#include "taint.dl"
+.output checkHasTag
+)";
+
 };
 
 }  // namespace raksha::xform_to_datalog
--- a/src/xform_to_datalog/datalog_facts.h
+++ b/src/xform_to_datalog/datalog_facts.h
@@ -1,0 +1,51 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+
+#ifndef SRC_XFORM_TO_DATALOG_DATALOG_FACTS_H_
+#define SRC_XFORM_TO_DATALOG_DATALOG_FACTS_H_
+
+#include <vector>
+
+#include "absl/strings/str_format.h"
+#include "src/ir/edge.h"
+#include "src/ir/tag_check.h"
+#include "src/ir/tag_claim.h"
+#include "src/xform_to_datalog/manifest_datalog_facts.h"
+#include "third_party/arcs/proto/manifest.pb.h"
+
+namespace raksha::xform_to_datalog {
+
+class DatalogFacts {
+ public:
+  DatalogFacts(ManifestDatalogFacts manifest_datalog_facts)
+      : manifest_datalog_facts_(std::move(manifest_datalog_facts)) {}
+
+  // Print out all contained facts as a single datalog string. Note: this
+  // does not contain the header files that would be necessary to run this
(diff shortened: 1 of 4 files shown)


src/xform_to_datalog/datalog_facts.h
```diff
@@ -0,0 +1,51 @@
...
+  DatalogFacts(ManifestDatalogFacts manifest_datalog_facts)
+      : manifest_datalog_facts_(std::move(manifest_datalog_facts)) {}
+
+  // Print out all contained facts as a single datalog string. Note: this
+  // does not contain the header files that would be necessary to run this
+  // against the datalog scripts; it contains only facts and comments.
+  std::string ToString() const {
+    return absl::StrCat(
+        "// GENERATED FILE, DO NOT EDIT!\n\n", "#include \"taint.dl\"\n",
+        manifest_datalog_facts_.ToString(), "\n.output checkHasTag\n");
```
markww, line 42: Should this just be a raw format string that can be placed in a `const` or `constexpr` static variable and use `StrFormat` instead of `StrCat`? Also, the [directives in Souffle can come in any order](https://souffle-lang.github.io/program#language), so this could all just be a prefix, which might be less complex and would allow us to cleanly separate the file into a static header and a generated suffix.

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -0,0 +1,57 @@
...
+
+class DatalogFactsTest : public testing::TestWithParam<ManifestDatalogFacts> {};
+
+TEST_P(DatalogFactsTest, IncludesManifestFactsWithCorrectPrefixAndSuffix) {
+  const ManifestDatalogFacts &manifest_datalog_facts = GetParam();
+  std::string expected_result_string =
+      absl::StrFormat(kDatalogFileTemplate, manifest_datalog_facts.ToString());
+  DatalogFacts datalog_facts(manifest_datalog_facts);
+  EXPECT_EQ(datalog_facts.ToString(), expected_result_string);
+}
```
markww, line 42: Nit: This is basically reimplementing the implementation. I know this is super simple code, which makes it sometimes hard and unsatisfying to do something else, but one way to approach this that feels a bit more like an independent proof would be to take the produced string and extract the generated prefix and suffix from it. That said, there's also not much to prove about this method, so don't worry about it if you don't think it's worth it.

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -0,0 +1,57 @@
...
+    testing::ValuesIn(
+        {ManifestDatalogFacts({}, {}, {}),
+         ManifestDatalogFacts(
+             {ir::TagClaim(ir::TagAnnotationOnAccessPath(
+                 ir::AccessPath(
+                     ir::AccessPathRoot(ir::HandleConnectionAccessPathRoot(
+                         "recipe", "particle", "out")),
+                     ir::AccessPathSelectors()),
+                 "tag"))},
+             {}, {})}));
```
markww, line 55: Nit: If you're going to inline the collection, using `Values` rather than `ValuesIn` might be a bit more readable.

src/xform_to_datalog/generate_datalog_program.cc
```diff
@@ -0,0 +1,88 @@
...
+constexpr char kUsageMessage[] =
+    "This tool takes a manifest proto and generates a datalog program.";
+
+int main(int argc, char* argv[]) {
+  google::InitGoogleLogging("generate_datalog_program");
+  absl::SetProgramUsageMessage(kUsageMessage);
+  absl::ParseCommandLine(argc, argv);
+
+  // Verify command line arguments
+  //
```
markww, line 44: Is this blank comment line intentional?

src/xform_to_datalog/generate_datalog_program.cc
```diff
@@ -0,0 +1,88 @@
...
+  std::filesystem::path manifest_filepath(absl::GetFlag(FLAGS_manifest_proto));
+  if (!std::filesystem::exists(manifest_filepath)) {
+    LOG(ERROR) << "Manifest proto file " << manifest_filepath
+               << " does not exist!";
+    return -1;
+  }
+
+  bool force_overwrite = absl::GetFlag(FLAGS_overwrite);
+  std::filesystem::path datalog_filepath(absl::GetFlag(FLAGS_datalog_file));
+  if (std::filesystem::exists(datalog_filepath) && !force_overwrite) {
```
markww, line 54: Nit: checking `force_overwrite` is an in-memory flag while `exists` is a system call, probably more performant to check `force_overwrite` on the left hand side of the `&&`. It will probably never matter this far from an inner loop, though, so feel free to disregard this.

src/xform_to_datalog/generate_datalog_program.cc
```diff
@@ -0,0 +1,88 @@
...
+  if (!manifest_proto.ParseFromIstream(&manifest_proto_stream)) {
+    LOG(ERROR) << "Error parsing the manifest proto " << manifest_filepath;
+  }
+
+  auto datalog_facts = raksha::xform_to_datalog::DatalogFacts(
+      raksha::xform_to_datalog::ManifestDatalogFacts::CreateFromManifestProto(
+          manifest_proto));
+
+  std::ofstream datalog_file(
+      datalog_filepath, std::ios::out | std::ios::trunc | std::ios::binary);
```
markww, line 78: Why `ios::binary` here? Just curious.

src/xform_to_datalog/generate_datalog_program.cc
```diff
@@ -0,0 +1,88 @@
...
+               << " does not exist!";
+    return -1;
+  }
+
+  bool force_overwrite = absl::GetFlag(FLAGS_overwrite);
+  std::filesystem::path datalog_filepath(absl::GetFlag(FLAGS_datalog_file));
+  if (std::filesystem::exists(datalog_filepath) && !force_overwrite) {
+    LOG(ERROR) << "Output datalog file " << datalog_filepath
+               << " exists! Forgot `--overwrite`?";
+    return -1;
```
markww, line 57: I'm not sure about using a negative exit code here. Different systems have different widths and signedness for dealing with exit codes, which may cause weird issues with matching the error exit code if we wanted to do so:

https://en.wikipedia.org/wiki/Exit_status

src/xform_to_datalog/datalog_facts.h
```diff
@@ -0,0 +1,51 @@
...
+#include "absl/strings/str_format.h"
+#include "src/ir/edge.h"
+#include "src/ir/tag_check.h"
+#include "src/ir/tag_claim.h"
+#include "src/xform_to_datalog/manifest_datalog_facts.h"
+#include "third_party/arcs/proto/manifest.pb.h"
+
+namespace raksha::xform_to_datalog {
+
+class DatalogFacts {
```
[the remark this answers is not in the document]
bgogul: Added.

src/xform_to_datalog/generate_datalog_program.cc
bgogul: ios::binary makes sure newline characters are written as is. Without `\n` will be written as `\r\n` on Windows machines, for example.

markww (approved):
LGTM

PR #119, opened 2021-09-14
Allow all policy code to come from auth logic.

aferr:
Every relation now has a speaker so that all aspects of policy code
aside from the graph (which comes from manifests) can come from
authorization logic. Aspects of how this works are likely to be iterated
on.

### DIFF
--- a/src/analysis/souffle/authorization_logic.dl
+++ b/src/analysis/souffle/authorization_logic.dl
@@ -64,7 +64,7 @@ claimHasTag(principal, path, tag) :- saysHasTag(principal, path, tag).
 // ownsTag from something with a speaker. This rule is likely to change. It
 // does not rule out the case that two different principals claim ownership of
 // a tag.
-ownsTag(speaker, tag) :- saysOwnsTag(speaker, owner, tag).
+ownsTag(owner, tag) :- saysOwnsTag(_, owner, tag).
 
 //-----------------------------------------------------------------------------
 // Universe Populating Rules
--- a/src/analysis/souffle/authorization_logic.dl
+++ b/src/analysis/souffle/authorization_logic.dl
@@ -56,7 +56,7 @@ downgrades(path, tag) :- ownsTag(owner, tag), saysDowngrades(owner, path, tag).
 // a tag, then it has the tag according to the taint analysis. This rule is
 // likely to be adjusted later. For confidentiality tags this should be OK to
 // start with.
-claimHasTag(path, tag) :- ownsTag(owner, tag), saysHasTag(_, path, tag).
+claimHasTag(principal, path, tag) :- saysHasTag(principal, path, tag).
 
 // This rule says that any principal can say that another principal owns a tag.
 // This rule is just a starting point to allow all aspects of the policy to
@@ -76,9 +76,6 @@ ownsTag(speaker, tag) :- saysOwnsTag(speaker, owner, tag).
 // cause all principals to be associated with all tags.
 // isTag(tag) :- ownsTag(_, tag).
 
-isTag(tag) :- saysDowngrades(_, _, tag).
-
-isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
 isPrincipal(owner) :- ownsTag(owner, _).
 
 #endif // SRC_ANALYSIS_SOUFFLE_AUTHORIZATION_LOGIC_DL_
--- a/src/analysis/souffle/authorization_logic.dl
+++ b/src/analysis/souffle/authorization_logic.dl
@@ -34,11 +34,13 @@
 
 .decl isPrincipal(p: Principal)
 
-isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
-
 // This party is the owner of a taint Tag (they have the privilege to downgrade
 // it).
 .decl ownsTag(owner: Principal, tag: Tag)
+
+.decl saysCreateTag(speaker: Principal, tag: Tag)
+
+.decl saysHasTag(speaker: Principal, path: AccessPath, t: Tag)
 
 // Some accessPath is actually downgraded (by taking into consideration the
 // privileges of a party trying to downgrade it, and as a result of that party
@@ -52,12 +54,34 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
 //-----------------------------------------------------------------------------
 downgrades(path, tag) :- ownsTag(owner, tag), saysDowngrades(owner, path, tag).
 
+claimNotHasTag(path, tag) :- downgrades(path, tag).
+
+// This rule conservatively says that if _any_ principal claims that a path has
+// a tag, then it has the tag according to the taint analysis. This rule is
+// likely to be adjusted later. For confidentiality tags this should be OK to
+// start with.
+claimHasTag(path, tag) :- saysHasTag(_, path, tag).
+
+// This rule says that any principal that creates a tag owns it. This rule is 
+// just a starting point to allow all aspects of the policy to come from 
+// authorization logic -- it is just a simple way to generate ownsTag from 
+// something with a speaker. This rule is likely to change. It does not rule
+// out the case that two different principals both create the same tag.
+ownsTag(speaker, tag) :- saysCreateTag(speaker, tag).
+
+//-----------------------------------------------------------------------------
+// Universe Populating Rules
+//-----------------------------------------------------------------------------
+
 // TODO(#126): Fix the universe populating rules, potentially by deleting
 //  them entirely.
-//
 // For now, we need to not populate isTag to handle cases where the tests
 // cause all principals to be associated with all tags.
 // isTag(tag) :- ownsTag(_, tag).
+
 isTag(tag) :- saysDowngrades(_, _, tag).
+
+isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
+isPrincipal(owner) :- ownsTag(owner, _).
 
 #endif // SRC_ANALYSIS_SOUFFLE_AUTHORIZATION_LOGIC_DL_
--- a/src/analysis/souffle/authorization_logic.dl
+++ b/src/analysis/souffle/authorization_logic.dl
@@ -28,8 +28,6 @@
 
 // Some party asserts the claim that the label of an access path should be
 // downgraded by removing the referenced tag from the set.
-// This relation needs to be in snake case because outputs from the 
-// authorization logic will be in snake case.
 .decl saysDowngrades(speaker: Principal, p: AccessPath, t: Tag)
 
 .decl isPrincipal(p: Principal)
@@ -38,7 +36,7 @@
 // it).
 .decl ownsTag(owner: Principal, tag: Tag)
 
-.decl saysCreateTag(speaker: Principal, tag: Tag)
+.decl saysOwnsTag(speaker: Principal, owner: Principal, tag: Tag)
 
 .decl saysHasTag(speaker: Principal, path: AccessPath, t: Tag)
 
@@ -54,20 +52,19 @@
 //-----------------------------------------------------------------------------
 downgrades(path, tag) :- ownsTag(owner, tag), saysDowngrades(owner, path, tag).
 
-claimNotHasTag(path, tag) :- downgrades(path, tag).
-
 // This rule conservatively says that if _any_ principal claims that a path has
 // a tag, then it has the tag according to the taint analysis. This rule is
 // likely to be adjusted later. For confidentiality tags this should be OK to
 // start with.
-claimHasTag(path, tag) :- saysHasTag(_, path, tag).
+claimHasTag(path, tag) :- ownsTag(owner, tag), saysHasTag(_, path, tag).
 
-// This rule says that any principal that creates a tag owns it. This rule is 
-// just a starting point to allow all aspects of the policy to come from 
-// authorization logic -- it is just a simple way to generate ownsTag from 
-// something with a speaker. This rule is likely to change. It does not rule
-// out the case that two different principals both create the same tag.
-ownsTag(speaker, tag) :- saysCreateTag(speaker, tag).
+// This rule says that any principal can say that another principal owns a tag.
+// This rule is just a starting point to allow all aspects of the policy to
+// come from authorization logic -- it is just a simple way to generate
+// ownsTag from something with a speaker. This rule is likely to change. It
+// does not rule out the case that two different principals claim ownership of
+// a tag.
+ownsTag(speaker, tag) :- saysOwnsTag(speaker, owner, tag).
 
 //-----------------------------------------------------------------------------
 // Universe Populating Rules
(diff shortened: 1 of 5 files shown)


src/analysis/souffle/examples/vision_pipeline/VisionPipelineTagOwners.dl
```diff
@@ -13,19 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //-----------------------------------------------------------------------------
-#include "authorization_logic.dl"
+#include "../../authorization_logic.dl"
```
markww, line 16: Using `..` in our include paths is not ideal. We should probably set up the `#include` roots in our invocation to souffle so that no prefix is needed or so that we use a full path from the root of the repo. I don't think this needs to be fixed right now, but something we should keep in mind once we get these things running automatically.

markww (commented):
This looks fine, and might be good for the short term, but as mentioned in the call we had earlier, I'm not sure whether this is necessary.

Summarizing what @aferr and I talked about on our call for the benefit of @bgogul and anyone else with the deep technical interest and/or copious free time to read this PR review:

My main concern in the discussion we had earlier was that we, as much as possible, encapsulate consideration of AccessPaths and other data layout details within Manifest files and that our interface with authorization logic abstract that away as much as possible. @aferr was worried that we wanted to keep the number of Datalog claims that the authorization logic down to just claiming that things had tags and that `.downgrades` would not be allowed. I said that if `.downgrades` was important for authorization logic to express the properties it needs to express we should definitely handle it eventually (although maybe not for MVP?). I suggested to @aferr that we could attach principals to `ParticleSpec`s in manifests for now using the `Annotation` mechanism, or even just conflating the domains of `ParticleSpec` names and principal names.

@bgogul, what do you think about the above? Do you think eliminating the speaker-ness is a good simplification for the MVP, or is it too deep a cut? If the former, then this PR is probably useful. If the latter, then this PR should probably be abandoned.

src/analysis/souffle/authorization_logic.dl
```diff
@@ -52,7 +54,29 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
 //-----------------------------------------------------------------------------
 downgrades(path, tag) :- ownsTag(owner, tag), saysDowngrades(owner, path, tag).
 
+claimNotHasTag(path, tag) :- downgrades(path, tag).
```
bgogul, line 57: What does this additional indirection buy us? Why not use `downgrades` directly as we used to?

src/analysis/souffle/authorization_logic.dl
```diff
@@ -52,7 +54,29 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
...
+
+// This rule conservatively says that if _any_ principal claims that a path has
+// a tag, then it has the tag according to the taint analysis. This rule is
+// likely to be adjusted later. For confidentiality tags this should be OK to
+// start with.
+claimHasTag(path, tag) :- saysHasTag(_, path, tag).
```
bgogul, line 63: Don't we need check the `ownsTag`? Otherwise, anyone can add a claim?

bgogul (commented):
> My main concern in the discussion we had earlier was that we, as much as possible, encapsulate consideration of AccessPaths and other data layout details within Manifest files and that our interface with authorization logic abstract that away as much as possible. @aferr was worried that we wanted to keep the number of Datalog claims that the authorization logic down to just claiming that things had tags and that downgrades would not be allowed. I said that if downgrades was important for authorization logic to express the properties it needs to express we should definitely handle it eventually (although maybe not for MVP?). I suggested to @aferr that we could attach principals to ParticleSpecs in manifests for now using the Annotation mechanism, or even just conflating the domains of ParticleSpec names and principal names.
>
> @bgogul, what do you think about the above? Do you think eliminating the speaker-ness is a good simplification for the MVP, or is it too deep a cut? If the former, then this PR is probably useful. If the latter, then this PR should probably be abandoned.

Why do say that authorization logic is limited to making claims on tags? Perhaps, we have been assuming that the manifest processing will only generate datalog facts that does not have principals? If we consider particles to be principals (as suggested here), there is nothing preventing us from generate a fact like `saysHasTag(particle, x, foo)` for a `claim x is foo`. Then we could combine with other mechanisms like delegation to generate a `claimHasTag(x, foo)` fact.

That said, I think the key idea to retain from this PR is the `saysHasTag` with slight modifications: 
 - Rename `claimHasTag` to `hasTag`
 - Reintroduce `claimHasTag` with an additional argument `claimHasTag(principal, access-path, tag)`. Note that this same as the `saysHasTag` in this PR.
 - Have a rule `hasTag(..) :- ownsTag(...), claimHasTag(...)`.
 
 Then, we can use them as follows: 
 - Make manifest processing generate `claimHasTag` facts with the particle as the principal.
 - We will need delegation claims so that ownership of tag is propagated to the corresponding particle. 
 
I am not sure if we need the other relations introduced in this PR. I have left some questions/comments at the appropriate places.

aferr:
I think this PR started because I misunderstood an earlier conversation we had. 

Here is what I thought the conversation we had was about (which I think I misunderstood):
 - Mark/Gogul wanted better encapsulation of auth-logic and manifests, and ideally the only relation auth-logic would need to to populate is claimHasTag. (Also, at the time of our conversation, I thought that this would work).
 - The [integrated example in this PR](https://github.com/google-research/raksha/pull/92) included manifests, authorization logic **and other datalog code** all as a part of the policy. The other datalog code was needed because 1) every relation produced by auth logic is of the form `says_<RelationName>(<PrincipalName>, <other_args>)`, 2) I am not sure how to make some aspects of policy indexed by speakers (tag ownership, application of tags to handles) or if this is actually best yet, so these parts needed to not have a speaker (i.e. they need to be globally true) .

Here is what I now think the conversation was about:
 - The [integrated example in this PR](https://github.com/google-research/raksha/pull/92) applies tags to accessPaths by writing out the full path names. It also does not distinguish between handles in declared particles and instantiated ones -- but what we want is that tags are specified on declarations but checked on instantiations.
 - A good way to solve these problems (at least for our MVP) is to have claims about paths come from manifests because manifests have all the information about paths.

This PR is more like a "hack" to allow tag ownership and application of tags to handles to have speakers (so that these facts can be produced by auth logic) rather than a completely worked out design for doing this.

src/analysis/souffle/examples/vision_pipeline/VisionPipelineTagOwners.dl
```diff
@@ -13,19 +13,12 @@
...
-// just because these facts are "universally true" rather than having a 
-// "speaker". We might also try making these "SomePrincipal says ownsTag(...)"
-// instead later, but this seems like a simpler way to start.
-ownsTag("EndUser", "raw_video_tag").
-ownsTag("EndUser", "detected_images_tag").
-ownsTag("EndUser", "user_selection_tag").
-ownsTag("EndUser", "product_id_tag").
-ownsTag("ServiceProvider", "image_detection_model_tag").
+// It comes from authorization logic where it looks like "P says createTag(t)"
+saysCreateTag("EndUser", "raw_video_tag").
```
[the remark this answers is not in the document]
aferr: (Context: this PR is just a "hack" to allow tag ownership, application of tags to be produced by authorization logic and to have a speaker. Because it's just a hack, it's not really a proper design).

All the facts from authorization logic start with `says_`, so to have tag ownership come from authorization logic, this needs to involve a relation that starts with says. Before this PR "ownsTag" was not a fact with a speaker, it was just globally believed by everyone.

Another way to do this could be

```
.decl says_OwnsTag(p: Principal, t: Tag)
ownsTag(principal, tag) :- saysOwnsTag(principal, principal, t).
```

This does not stop an adversarial principal from taking ownership of a tag owned by another one, though. A way we can at least make this similar to the case where tag ownership is a "unversal truth" is to have this rule and some other rule that says the whole policy fails if there are multiple principals that own the same tag. Roughly:

```
FAIL() :- ownsTag(p1, tag), ownsTag(p2, tag), p1 != p2
```

(There are probably even more aspects of giving speakers to tag ownership / applying tags to handles that would need to be fleshed out more if we want to keep this.)

src/analysis/souffle/authorization_logic.dl
aferr: I think it doesn't actually buy us much and we can get rid of it. As of this PR, I was thinking that you and Mark wanted more encapsulation between auth logic and DFA, and perhaps having a relation that looks like `claim path is NOT tag` would help with this encapsulation somehow?

Probably we should get rid of this extra indirection and keep it as downgrades.

src/analysis/souffle/authorization_logic.dl
aferr: Checking ownership here does sound reasonable at a glance. Although, since in this MVP we just have confidentiality tags, allowing anyone to add a tag is harmless, so we don't strictly need that yet. I can add the ownsTag check anyway since it does sound reasonable.

aferr:
> * Reintroduce `claimHasTag` with an additional argument `claimHasTag(principal, access-path, tag)`. Note that this same as the `saysHasTag` in this PR.

The name `saysHasTag(...)` rather than `claimHasTag()` does actually matter just for the reason that relations produced as an output of authorization logic all start with `says_...` in the name. (We could change this later, but for now that's how it is).

src/analysis/souffle/examples/vision_pipeline/VisionPipelineTagOwners.dl
aferr: OK I'll keep this in mind. I don't have a build script for this. I was just running souffle on it manually.

src/analysis/souffle/examples/vision_pipeline/VisionPipelineTagOwners.dl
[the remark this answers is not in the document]
aferr: I made this `saysOwnsTag(speaker, owner, tag)`, just so that it is possible to pull this in from the auth logic code for now. Hopefully the name `saysOwnsTag` makes its relationship with `ownsTag` more clear.

Fine to leave it for now?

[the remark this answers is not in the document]
aferr: Marking this resolved but do feel free to unresolve (I think that's our procedure, right?) :)

src/analysis/souffle/taint.dl
```diff
@@ -52,7 +52,8 @@ hasTag(tgt, tag) :- claimHasTag(principal, tgt, tag), ownsTag(principal, tag).
 mayHaveTag(tgt, tag) :- hasTag(tgt, tag).
 
 mayHaveTag(tgt, tag) :-
-  edge(src, tgt), mayHaveTag(src, tag), !downgrades(tgt, tag).
+    edge(src, tgt), mayHaveTag(src, tag),
+    !downgrades(tgt, tag).
```
markww, line 56: Nit: Why put `!downgrades(tgt, tag)` on its own line? There's plenty of room on the previous line with a line length of 80 characters. Gogul and I agreed to make the line length 100 characters in Datalog to further reduce the number of newlines. I think this was early on though, so it might not have been stated in a group chat.

src/analysis/souffle/authorization_logic.dl
```diff
@@ -52,12 +52,33 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
 //-----------------------------------------------------------------------------
 downgrades(path, tag) :- ownsTag(owner, tag), saysDowngrades(owner, path, tag).
 
+// This rule conservatively says that if _any_ principal claims that a path has
+// a tag, then it has the tag according to the taint analysis. This rule is
+// likely to be adjusted later. For confidentiality tags this should be OK to
+// start with.
+claimHasTag(path, tag) :- ownsTag(owner, tag), saysHasTag(_, path, tag).
```
markww, line 59: This is in conflict with #121 , which introduces a `claimHasTag` with a `Principal`.

src/analysis/souffle/authorization_logic.dl
```diff
@@ -52,12 +52,33 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
...
+
+// This rule says that any principal can say that another principal owns a tag.
+// This rule is just a starting point to allow all aspects of the policy to
+// come from authorization logic -- it is just a simple way to generate
+// ownsTag from something with a speaker. This rule is likely to change. It
+// does not rule out the case that two different principals claim ownership of
+// a tag.
+ownsTag(speaker, tag) :- saysOwnsTag(speaker, owner, tag).
```
markww, line 67: Is this right? I would think that "any speaker may say that any principal may own a tag" would be written as

```
ownsTag(owner, tag) :- isPrincipal(speaker), saysOwnsTag(speaker, owner, tag).

File: src/analysis/souffle/authorization_logic.dl [hunk 8]
```diff
@@ -52,12 +52,33 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
...
 //  them entirely.
-//
 // For now, we need to not populate isTag to handle cases where the tests
 // cause all principals to be associated with all tags.
 // isTag(tag) :- ownsTag(_, tag).
+
 isTag(tag) :- saysDowngrades(_, _, tag).
 
+isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
+isPrincipal(owner) :- ownsTag(owner, _).
```
markww, line 82: We probably shouldn't introduce more universe populating rules that are based upon inferred relations, see #126.

src/analysis/souffle/authorization_logic.dl
aferr: Ah, I must have missed this piece when resolving the rebase. I made it just `claimHasTag(p, q, t) :- saysHasTag(p, q, t)` so that we can still pull facts from the auth logic implementation (since it emits things beginning with says and these are still somewhat separate pieces of software for this MVP).

src/analysis/souffle/authorization_logic.dl
markww: The key point of my comment isn't about the `isPrincipal` part. What I'm saying is that your comment says that:

```
This rule says that any principal can say that another principal owns a tag.
```

My interpretation of the rule you wrote in English is "A `speaker` owns a `tag` if `speaker` says any `owner` owns that `tag`". I may be misunderstanding, but the best way to match the comment to the rule would be Datalog that instead says "An `owner` owns a `tag if any `speaker` says that `owner` owns that `tag`". An equivalent way to write my suggestion is:

```
ownsTag(owner, tag) :- saysOwnsTag(_, owner, tag).
```

markww (approved):
This looks good. Please see and address the comment I reopened if it's a real issue. Speculatively approving to reduce timezone friction.

src/analysis/souffle/authorization_logic.dl
aferr: Oh yes. The rule you wrote there is what I actually meant to put. I overindexed on the `isPrincipal` part. Thanks and good catch.

bgogul (approved):

Issue #118, opened 2021-09-14
Disentangle the close intertwining of the IR classes and the protos.


bgogul:
The IR classes have static methods like `CreateFromProto`, which creates a close dependency between protos and the IR classes. Ideally, the proto-related functionality should be split off of the IR classes and moved to a separate file/class.

PR #123, opened 2021-09-16
Add kind() virtual method to Type.

bgogul:
This is needed for #118

### DIFF
--- a/src/ir/types/type_test.cc
+++ b/src/ir/types/type_test.cc
@@ -303,4 +303,10 @@ INSTANTIATE_TEST_SUITE_P(
     RoundTripTypeProtoThroughTypeTest,
     testing::ValuesIn(type_proto_and_access_path_strings));
 
+// TODO(#122): This test should be moved to appropriate file while refactoring.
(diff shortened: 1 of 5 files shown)


src/ir/types/type_test.cc
```diff
@@ -303,4 +303,10 @@ INSTANTIATE_TEST_SUITE_P(
     RoundTripTypeProtoThroughTypeTest,
     testing::ValuesIn(type_proto_and_access_path_strings));
 
+// TODO(#122): This test should be moved to appropriate file while refactoring.
+TEST(EntityTypeTest, KindReturnsCorrectKind) {
+  EntityType entity_type(Schema(std::nullopt, {}));
```
markww, line 308: This is using `std::nullopt` when the `Schema` signature uses `absl::optional`. I'm guessing that this works because `absl::optional` is actually just `typedef`ed to `std::optional` internally for C++17. However, we should probably use `absl::nullopt` for consistency reasons.

markww (approved):

PR #124, opened 2021-09-16
Split off protobuf handling of types to a separate package.

bgogul:
Apologize for the big PR. It wasn't possible to split it up further due to interdependencies between types. This PR does the following:
  - Move the code from `CreateFromProto` and `MakeProto` methods to the `ir::types::proto` namespace.
  - Migrate some of the tests and also adds new tests.

Makes progress towards  #118.


src/ir/types/BUILD
```diff
@@ -37,6 +37,7 @@ cc_test(
         "//src/common/logging",
         "//src/common/testing:gtest",
         "//src/ir",
+        "//src/ir/proto:types",
```
markww, line 40: Should we also remove the `//third_party/arcs/proto:manifest_cc_proto` dependency from the `:types` library?

(Sorry for the odd placement, GitHub won't let me put comments outside of the context it draws around modified lines).

markww (approved):

PR #125, opened 2021-09-16
An Expanded Vision Pipeline Policy Example

aferr:


src/analysis/souffle/examples/vision_pipeline/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,96 @@
...
+// For each particle P, for each tag t in union over tags in read edges of P,
+//      for each statement of the form:
+//          `P says will <RELATION> using <TAG>`
+//      prove
+//          `ownsTag(Q, <TAG>)`
+//          `Q says may <RELATION> using <TAG>`
+//
+// Notably "may" is a keyword rather than just another relation itself because it 
+// seems useful to let usages be arbitrary relations (with potentially varying 
+// signatures).
```
markww, line 96: On this `may` and `will` part, I think that `may` is definitely in the scope of the authorization logic, but I'm wondering whether `will` is.

The distinction between `may` and `will` actually sums up the distinction in my mind between the concepts that should go in the Manifest (or what the manifest will become) vs. what goes in the authorization logic. The Manifest is an abstract description of the behaviors that either will definitely happen or will possibly happen. The authorization logic, on the other hand, is the description of the behaviors that are permitted to occur. The point of Raksha is to abstractly execute the abstract description of the program's behavior to see if what can happen can be made to conflict with what may happen.

The abstract description of behavior that you are placing on `will` via arbitrary relations may speak to a desire to have `ParticleSpec`s be more than a black box; perhaps they could be augmented with some abstract description of their internal behavior, which may then be governed via the `may` statements you use here.

What do you think about this distinction?

src/analysis/souffle/taint.dl
```diff
@@ -38,13 +38,17 @@
...
+.decl claimNotHasTag(accessPath: AccessPath, tag: Tag)
+
 //-----------------------------------------------------------------------------
 // Rules
 //-----------------------------------------------------------------------------
 mayHaveTag(tgt, tag) :- claimHasTag(tgt, tag).
 mayHaveTag(tgt, tag) :-
     edge(src, tgt), mayHaveTag(src, tag),
-    !downgrades(tgt, tag).
+    !claimNotHasTag(tgt, tag).
```
markww, line 51: Again, this is likely in conflict with changes introduced by #121 .

src/analysis/souffle/authorization_logic.dl
```diff
@@ -52,7 +54,29 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
 //-----------------------------------------------------------------------------
 downgrades(path, tag) :- ownsTag(owner, tag), saysDowngrades(owner, path, tag).
 
+claimNotHasTag(path, tag) :- downgrades(path, tag).
+
+// This rule conservatively says that if _any_ principal claims that a path has
+// a tag, then it has the tag according to the taint analysis. This rule is
+// likely to be adjusted later. For confidentiality tags this should be OK to
+// start with.
+claimHasTag(path, tag) :- saysHasTag(_, path, tag).
```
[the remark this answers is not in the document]
aferr: fixed

src/analysis/souffle/authorization_logic.dl
```diff
@@ -52,7 +54,29 @@ isPrincipal(speaker) :- saysDowngrades(speaker, _, _).
...
+
+// This rule says that any principal that creates a tag owns it. This rule is 
+// just a starting point to allow all aspects of the policy to come from 
+// authorization logic -- it is just a simple way to generate ownsTag from 
+// something with a speaker. This rule is likely to change. It does not rule
+// out the case that two different principals both create the same tag.
+ownsTag(speaker, tag) :- saysCreateTag(speaker, tag).
```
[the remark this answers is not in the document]
aferr: yes,  that makes sense to me.

src/analysis/souffle/examples/vision_pipeline/VisionPipelineAuthLogic
aferr: Will we need this distinction after the MVP? I thought we were moving towards having just one language. Once we have our own (one) parser, I think it should be fairly easy to have language constructs from authorization logic in the relevant parts of what is currently expressed with manifests as well. If we do it that way, we can easily use `will ...` in a way that refers to paths in particles accurately (which is what I think you want to gain by putting this functionality in manifests) but would still allow this language feature to interact with principals and other concepts from authorization language in the intended way. This would also allow it to be applied to relations of arbitrary arity (which I think we would need if the `may` construct can be applied to arbitrary relations). 

I don't know a ton about the parser combinators that Gogul mentioned last time, but using one of these seems like it would help us both: 1) use authorization logic constructs in arbitrary places of the raksha syntax, and 2) if we want to, keep an implementation of authorization logic as a separate modular block. (Where doing 2) could be useful because it might be possible to re-use  authorization logic for other use-cases that are not about use of privacy-sensitive data -- then we'll all have two broadly applicable pieces of software that we are all involved in producing. Though this is probably lower priority than getting our one combined thing to work well first).

src/analysis/souffle/taint.dl
aferr: fixed.

src/analysis/souffle/examples/vision_pipeline/VisionPipelineAuthLogic
aferr: I think for the MVP, there is probably a way to move the `will` parts into the manifests where we can just have some additional kind of annotation that means `ParticleAsPrincipal says ParticleAsPrincipal will Relation`. Exactly the cheapest/easiest way to do this is something Gogul/you might know better. 

Just for the purpose of writing demos while are still working on the MVP, it's useful for me to also be able to write this in auth logic code because these relations interact with other parts of auth logic. If it's useful to be able to write `will()` relations in either place (manifests or auth logic code) we can do the same thing we are doing with `claimHasTag(_, _, _)` and `saysHasTag(_, _, _)` which lets us do it in both places.

markww: Even after we have one language for both the authorization and dataflow aspects, it may be useful to keep an eye on how much we entangle them. If we have things separated cleanly enough, it could enable nice properties such as an administrator running multiple experiments with hypothetical policies against a single dataflow graph. That said, the benefits are speculative and I'm not sure such a separation would necessarily work, so let's not worry about it too much right now.

markww (dismissed):

aferr:
The authorization logic part is now fully working and does not require any new features on that side. Just the "check" of usages described for particles would need to be implemented.

src/analysis/souffle/examples/vision_pipeline/VisionPipelineAuthLogic
aferr: If we use the `ParticleAsPrincipal says will <Relation>(<args>...)` syntax within a manifest file, it should be possible to test this one manifest that describes how the application will use data against several distinct policies about allowed data use (the different policies about allowed use could be in separate files from the manifest).

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,153 @@
...
+    // These lines say the end user trusts (i.e., delegates to) the TPPR to
+    // faithfully say when a data transformation from raw camera video to the 
+    // name of a product has taken place, but only when the user has selected 
+    // an object.
+    //
+    // It might look like this ideally:
+    //
+    // "ThirdPartyPrivacyReviewer" canSay becomes(pathX, "raw_video_tag",
+    //              "product_id_tag") :-
+    //     objectSelected("EndUser", pathX),
```
bgogul, line 31: What does `objectSelected` capture? Does this say that `EndUser` selected some object that is referred to by `pathX`?

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,153 @@
...
+    "ThirdPartyPrivacyReviewer" canSay declassifies(pathX, "raw_video_tag") :-
+        hasTag(pathX, "product_id_tag"),   // raw_video_tag is only removed if 
+        objectSelected("EndUser", pathX),  // product_id_tag is added
+        isAccessPath(pathX).
+
+    "ThirdPartyPrivacyReviewer" canSay
+        objectSelected("EndUser", "identified_product").
+
+    // Delegates TPPR the ability to apply tags anywhere
+    "ThirdPartyPrivacyReviewer" canSay hasTag(x, y) :-
```
bgogul, line 44: Shouldn't this `tag` also be owned by `EndUser`?

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,153 @@
...
+        isAccessPath(pathX).
+}
+
+// Maybe we want this part of the policy to sit closer to the particles
+// because sometimes relations will refer to path names ?
+// (We can ostensibly do this as long as we can write
+// `claim <RelationName(...)>` in manifests)
+"VideoDataSource" says "VideoDataSource"
+    will("do_pure_computation", "raw_video_tag").
+"ImageDetector" says "ImageDetecor"
```
bgogul, line 93: Typo: `ImageDetecor` -> `ImageDetector`?

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,153 @@
...
+        isAccessPath(pathX).
+    "ImageSelector" canSay objectSelected("EndUser", pathX) :-
+        isAccessPath(pathX).
+}
+
+// Maybe we want this part of the policy to sit closer to the particles
+// because sometimes relations will refer to path names ?
+// (We can ostensibly do this as long as we can write
+// `claim <RelationName(...)>` in manifests)
+"VideoDataSource" says "VideoDataSource"
```
bgogul, line 91: Why is "VideoDataSource" repeated?

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Some queries to test this policy:
+video_is_local = query "EndUser" says   
+    "VideoDataSource" canActAs "LocalDevice"?
+
+local_can_do_computation = query "EndUser" says
+    "LocalDevice" may("do_pure_computation", "raw_video_tag")?
+
+////// These are the queries from the "will" relations that need to be checked
```
bgogul, line 109: What is the high-level policy these queries are capturing? Or, is it the case that these are just some test queries?

bgogul (approved):

src/analysis/souffle/examples/vision_pipeline/VisionPipelineAuthLogic
aferr: We just discussed in our 1:1 and I think we realized we were essentially in agreement that this information should go with specification of the application's behavior. I think we are kind of undecided about what exactly the syntax of allowed usage will be.

(Closing for now, but feel free to re-open)

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: Yes, that's what it's supposed to mean. If `ThirdPartyPrivacyReviewer says objectSelected("EndUser", "some_path")`, it's supposed to mean that reviewer read the code and confirmed that the value on `some_path` is really something `EndUser` wants to perform some action on.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: If that's the behavior we want, it doesn't need to be written in this policy code because this is already checked in Raksha's analysis.  
` X says hasTag(path, tag)` becomes `saysHasTag(x, path, tag)` (modulo camel->snake :) and this fact doesn't do anything unless X owns the tag (or there is some delegation like the one you are commenting on). In the Raksha implementation this is true because in 
authorization_logic.dl there is the line
`claimHasTag(principal, path, tag) :- saysHasTag(principal, path, tag).`
and in `taint.dl` there is the line `hasTag(tgt, tag) :- claimHasTag(principal, tgt, tag), ownsTag(principal, tag).`
which checks for ownership.

(I think checking this is not really needed though because the taint analysis estimates whether a handle contains sensitive data, and adding more sensitive data can only make it more likely that the code is rejected by the policy. So it is always safe to say there is more sensitive data on a handle).

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: `"VideoDataSource" will("do_pure_computation", "raw_video_data_tag")` is very similar to `will("VideoDataSource", "do_pure_computation", "raw_video_tag") except that when it's written this way it can interact with the `canActAs` construct to pass the `will(...)` predicate to other principals (like the "LocalDevice" group). Predicates with the syntax `<Principal>    <Relation>(<args>)` (and with this interaction with `canActAs`) are kind of like "properties" of the principal and `canActAs` is kind of like saying one principal can pick up the properties of another.

So that's the inner instance of "VideoDataSource".

The outer "VideoDataSource"  before `says` is because VDS is the principal claiming that VDS has this property. VDS is describing its own behavior in this example.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: These are not meant to be part of the policy. These are just tests that try to ensure that this part of the code would work if we plug it into the other parts. The `will` statements are meant to be converted into "checks" (not necessarily the arcs keyword) that we can prove the facts that these queries test.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: Oops fixed.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: The check I'm imagining is the one that the comment towards the bottom tries to describe in natural language. As you pointed out it also fits nicely with how another policy language we have already works, but ties it to principals, delegation, and expands what can be expressed to predicates.

PR #129, opened 2021-09-16
How we might accomplish multimic example with manifest/policy.

markww:
Using the Arcs manifest format and policy format, this attempts to
encode how we might accomplish the example of a user with a microphone
that can

* export its text to speech model to a microphone closer to the user

* detect when a user comes in with an incompatible policy and reason
  about whether to keep recording or stop the text to speech.


src/analysis/souffle/examples/multimic/multimic.arcs
```diff
@@ -0,0 +1,59 @@
...
+particle MicIn
+  audio: writes AudioData { }
+// SPECULATIVE: claim a QoS integrity tag.
+//claim audio is userSpeech and is integrity#speechToTextQoS
+  claim audio is userSpeech and is speechToTextQoS
+
+// Represents the CPU within the mic.
+particle MicCompute
+  audioIn: reads AudioData { }
+  asrNeuralNet: reads writes ASRNeuralNet { }
```
bgogul, line 14: Why does this have to write also? Isn't `ASRNeuralNet` the ML model?

src/analysis/souffle/examples/multimic/multimic.arcs
```diff
@@ -0,0 +1,59 @@
...
+  SpeechToTextStorageSink
+    speechToTextIn: reads m1AsrStorage
+
+recipe HandOffASRToOtherMic
+  m2AsrNetStorage: create
+  m1AsrNetStorage: map 'm1AsrNetStorage'
+  m1AsrStorage: use 'm1AsrStorage'
+  ForwardAsrNet
+    inNeuralNet: reads m1AsrNetStorage
+    outNeuralNet: writes m2AsrNetStorage
```
bgogul, line 50: Are you saying that for hand off, `mic1` will also forward an ML model to `mic2`?

src/analysis/souffle/examples/multimic/multimic_micowner_policy.arcs
```diff
@@ -0,0 +1,30 @@
...
+//@PrincipalPriority(“UserB”, 50)
+//@PrincipalPriority(“UserC”, 50)
+@egressType("Disk")
+policy UserASmartMicOwnerPolicy {
+  // textTranscript is a hypothetical label describing text derived from user
+  // speech instead of the raw audio itself.
+  @allowedUsage(label: 'textTranscript', usageType: '*')
+  @allowedRetention(medium: 'Ram', encryption: false)
+  @allowedRetention(medium: 'Disk', encryption: false)
+  from ASRData access { }
```
bgogul, line 28: Shouldn't this be from `AudioData` instead of `ASRData`? Essentially, don't you want to put restrictions on how raw audio stream is treated?

src/analysis/souffle/examples/multimic/multimic_userc_policy.arcs
```diff
@@ -0,0 +1,13 @@
...
+// User C's personal policy
+//SPECULATIVE:
+//@PrincipalPriority("UserC", 0)
+@intendedPurpose('Personal Data Policy of User C')
+@egressType("NONE")
+policy User_C_Policy {
+  @allowedUsage(label: 'textTranscript', usageType: "noEgress")
+  @allowedRetention(medium: 'Ram', encryption: true)
+  @maxAge('1hour')
+  from ASRData access { }
```
bgogul, line 12: `AudioData`?

src/analysis/souffle/examples/multimic/multimic.arcs
```diff
@@ -0,0 +1,59 @@
+// Note: This file marks speculative syntax with a SPECULATIVE comment before
+// the speculative syntax.
+
+// Represents an actual microphone within a smart mic listening to audio.
+particle MicIn
+  audio: writes AudioData { }
+// SPECULATIVE: claim a QoS integrity tag.
+//claim audio is userSpeech and is integrity#speechToTextQoS
```
bgogul, line 8: What is the motivation for the integrity tags?

src/analysis/souffle/examples/multimic/multimic.arcs
markww: It is. I thought it might be continually training on the user's speech patterns to improve the model. But if that's not the case, then writing would not be needed.

src/analysis/souffle/examples/multimic/multimic.arcs
markww: Yes, I was thinking that the model would be continually personalized to the user(s) of `mic1`, and that exporting the model would allow `mic2` to continue with the same level of experience. But this probably isn't a necessary feature.

src/analysis/souffle/examples/multimic/multimic_userc_policy.arcs
markww: Agreed, see other comment.

src/analysis/souffle/examples/multimic/multimic.arcs
markww: I'm trying to express in the Arcs Manifest language a user's desire that the data path turning raw audio into a text transcript of speech stored upon the user's smart mic is not interrupted. A guest policy could introduce restrictions at multiple points along this path that could interrupt this path, so we verify that this path exists by proving that a tag may flow from the data source to the storage device.

Now, why an integrity tag vs a secrecy tag? Let's expand the example and say that the microphone's processing is, in addition to ASR data, some microphone health monitoring data, like which frequencies it has heard lately to ensure that the microphone is working properly. Imagine that this is stored in the same store as the actual ASR data. In this case, a user policy might interrupt the critical text to speech pipeline with their policy but leave the health metrics, through which the secrecy QoS tag could pass and cause the check to falsely pass, causing a false negative on the QoS check.

I know that we can work around that with separating the stores or adding `derives from` claims. That is probably what we will do until we have integrity tags. But to use the secrecy tags in this way, we'd have to actively work to pen the secrecy QoS tag onto the path that we want to prevent false negatives. So, long term, the propagation behavior of an integrity tag seems like a better fit for ensuring the type of property that we are trying to guarantee.

src/analysis/souffle/examples/multimic/multimic.arcs
markww: Fixed.

src/analysis/souffle/examples/multimic/multimic.arcs
markww: Fixed.

src/analysis/souffle/examples/multimic/multimic_micowner_policy.arcs
markww: Fixed.

src/analysis/souffle/examples/multimic/multimic_userc_policy.arcs
markww: Fixed.

src/analysis/souffle/examples/multimic/multimic.arcs
markww: Added an annotation which indicates that the original user may attempt to block dataflow to accomodate UserC's policy. In this case, however, the QoS tag would also be interrupted, ending in a new check, forcing the discarding of UserC's policy.

src/analysis/souffle/examples/multimic/multimic_micowner_policy.arcs
```diff
@@ -0,0 +1,69 @@
...
+  // not very sensitive.
+  @allowedUsage(label: 'textTranscript', usageType: '*')
+  @allowedRetention(medium: 'Ram', encryption: false)
+  @allowedRetention(medium: 'Disk', encryption: false)
+
+// SPECULATIVE:
+// The @canDowngrade annotation indicates principals that can downgrade labels.
+// In this case, UserA trusts MicCompute to correctly downgrade its raw
+// userSpeech to less sensitive data.
+// @canDowngrade(label: userSpeech, principals: ['MicCompute'])
```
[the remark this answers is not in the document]
aferr: Ahhh got it. Sorry for the extremely delayed review :)

src/analysis/souffle/examples/multimic/multimic_micowner_policy.arcs
```diff
@@ -0,0 +1,69 @@
...
+  from ASRData access { }
+
+// SPECULATIVE:
+// We have discussed qualifying tags with the owners of those tags, so that
+// multiple principals can use the same tag without causing collisions in the
+// dataflow graph. In this case, however, the mic is creating data from signal
+// that many users may have a stake in. Open the tag up for this behavior by
+// making it a sharedTag.
+// @sharedTag('userSpeech')
```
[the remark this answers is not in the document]
aferr: (Comment above is probably deprecated).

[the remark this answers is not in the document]
markww: Not at all, this is helpful info for my mental model. In particular, I wasn't clear on how multiple ownership interacted with downgrades.

PR #130, opened 2021-09-17
Replace absl variant,optional with std versions.

### LINKED ISSUE #128: Replace `absl` compatibility types with `std` equivalents.
Currently, we are using `absl::optional` and `absl::variant` within the Raksha code. We should switch to the C++ standard library versions instead.

markww:
Fixes #128 .


bgogul (approved):
Nice! Thanks for taking care of this.

PR #131, opened 2021-09-17
Add an `overloaded` utility method for using inplace labmdas for visiting variants.

bgogul:
This utility is useful if we want to add variant visitors, where each variant needs a different behavior. See overloaded_test.cc for a usage.

### DIFF
--- a/src/utils/BUILD
+++ b/src/utils/BUILD
@@ -1,0 +1,32 @@
+#-------------------------------------------------------------------------------
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


markww (approved):
Cool trick. There could be readability hazards if we overused it, but I can definitely see how it could make for more readable code for `std::variant`s with a small number of alternatives and/or a small number of methods. LGTM.

bgogul:
> Cool trick. There could be readability hazards if we overused it, but I can definitely see how it could make for more readable code for `std::variant`s with a small number of alternatives and/or a small number of methods. LGTM.

Not my idea. This is from the C++ documentation for visit. Agreed that we should use it sparingly.

Issue #98, opened 2021-08-31
Give `AccessPathSelectors` an iterator that allows forward iterating in the semantic order of the contained Selectors.


markww:
`AccessPathSelectors` stores the `Selector`s it contains in a `std::vector` in reverse order. This is to allow for performant addition of `Selector`s while returning partial `AccessPathSelectors` from a type tree in a bottom-up fashion. This is somewhat annoying internally, however, as it means that regular C++ range-based `for` loops cannot be used. Instead, we have to use an explicit `for` loop on the `rbegin()` and `rend()` iterators.

To improve readability, we should add a `begin()` and `end()` method on the `AccessPathSelectors` object that return iterators that loop over the `Selector`s in semantic order (opposite of the reversed mechanical order). This would allow us to use regular range-based `for` loops and produce more readable code.

markww:
The structured form of `AccessPath` is no longer used outside of the Raksha Arcs manifest frontend, which is deprecated. Closing.

PR #133, opened 2021-09-17
Add forward iterator for AccessPathSelectors.

bgogul:
A quick and dirty solution to address #98. Ideally we want to have use something like std::ranges (which is only in C++20). However, this should serve as a good solution for now.


markww (approved):
LGTM

Issue #105, opened 2021-09-01
Rename functions that print fact classes as strings containing Datalog from `ToString` to `ToDatalog`.


markww:
While constructing the data structures containing Datalog facts, I gave each a method called `ToString`, which prints that fact as a string containing the fact expressed in Souffle Datalog syntax. @bgogul pointed out that a name like `ToDatalog` would be better. Because most of these functions have already been created, I'm going to keep naming the remaining ones `ToString` for now, for consistency's sake, then rename all of them `ToDatalog` in a single PR that I will target at this issue.

markww:
This has been done.

PR #134, opened 2021-09-20
Rename ToString to ToDatalog for IR objects that become Datlog facts.

markww:
This leaves structures below the level of a Datalog fact (such as
AccessPathSelectors) as ToString for now.


markww:
This is relevant to (and, depending upon what reviewers think, may completely fix) #105 .

bgogul (approved):

PR #141, opened 2021-09-22
    Dispute Example

aferr:
This expands the vision pipeline example with two stakeholders
    that both own data that might be egressed. They have different
    preferences with how this "dispute" should be handled, but they
    both delegate to a policy that resolves this dispute. There are 
    different ways of writing how disputes should be handled.

### DIFF
--- a/rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
+++ b/rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
@@ -50,8 +50,16 @@
         isPrincipal(principalX).
 
     "LocalDevice" may("do_pure_computation", "raw_video_tag").
-    "LocalDevice" may("egress_to_shopping_service", "product_id_tag").
 
+    // "EndUser" opts-in to a policy about data egress.
+    // It delegates to this policy the right to decide what modules can egress
+    // any of "EndUser's" data.
+    "DataEgressPolicy1" canSay principalX may("egress_to_shopping_service", tagX) :-
+        isPrincipal(principalX), isTag(tagX).
+    // The EndUser also expresses a preference to this policy, which is 
+    // "allowEgress":
+    "EndUser" hasPrivacyPreference("allowEgress").
+        
     // Could also give different permissions to RemoteService for once it
     // has "product_id_tag".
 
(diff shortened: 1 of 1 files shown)


aferr:
I think it's useful to point out that this dispute outcome where one party does not get what they want is only possible because all parties _did_ agree to sign up for (delegate to) the policy that resolves the dispute (`DataEgressPolicy1`).

bgogul (commented):
@aferr I find the example hard to follow as the PR also seems to have some code that is already merged in.  Can you rebase to latest HEAD?

It might also be useful for you to walk me through the example.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+// Similar to policy 1, but we need agreement from BOTH parties:
+"DataEgressPolicy2" says {
+    // Gathers preferences from both stakeholders:
+    "EndUser" canSay "EndUser" hasPreference(prefX) :-
+        isPreference(prefX).
+    "ServiceProvider" canSay "ServiceProvider" hasPreference(prefX) :-
+        isPreference(prefX).
+
+    "ProductIdOutput" may("egress_to_shopping_service", "product_id_tag") :-
+        "EndUser" hasPreference("allowEgress").
```
bgogul, line 135: Should this be `"EndUser" says hasPreferenece("allowedEgress")`?

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+    "ProductIdOutput" may("egress_to_shopping_service",
+        "image_detection_model_tag") :-
+        // This policy DOES care about ServiceProvider's preference:
+        "ServiceProvider" hasPreference("allowEgress:).
```
bgogul, line 139: is `"allowedEgress:` a typo? Should it be `"allowedEgress"`?

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+    // (Both of these facts are used to make the same one decision about 
+    // egressing a handle that depends on both of these tags)
+    "ProductIdOutput" may("egress_to_shopping_service", "product_id_tag") :-
+        "EndUser" hasPreference("allowEgress").
+    "ProductIdOutput" may("egress_to_shopping_service", 
+        "image_detection_model_tag"). // Doesn't care about ServiceProvider pref
+}
+
+// Similar to policy 1, but we need agreement from BOTH parties:
+"DataEgressPolicy2" says {
```
bgogul, line 127: How the disparity in preferences reconciled?

bgogul (commented):
@aferr, I left some comments in the PR. Perhaps, I am missing something, but it is not clear to me how the disparity in the policies are reconciled.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+    "LocalDevice" may("do_pure_computation", "raw_video_tag").
+
+    // "EndUser" opts-in to a policy about data egress.
+    // It delegates to this policy the right to decide what modules can egress
+    // any of "EndUser's" data.
+    "DataEgressPolicy1" canSay principalX may("egress_to_shopping_service", tagX) :-
+        isPrincipal(principalX), isTag(tagX).
+    // The EndUser also expresses a preference to this policy, which is 
+    // "allowEgress":
+    "EndUser" hasPrivacyPreference("allowEgress").
```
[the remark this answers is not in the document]
aferr: "allowEgress" is not a tag. The predicate `hasPrivacyPreference` is an attribute of "EndUser". Our manual describes this feature ( [here](https://github.com/google-research/raksha/tree/main/rust/tools/authorization-logic#attributes-and-canactas) (It's similar to just writing `hasPrivacyPreference("EndUser", "allowEgress") but if you write it as an attribute like this, it can be passed with `canActAs`). It's meant to capture that the _preference_ of `EndUser` is to allow egress of data that depends on objects selected by the user.

So all together `EndUser says EndUser hasPrivacyPreference("allowEgress")` that the `EndUser` is claiming that `EndUser` has the privacy preference "allowEgress".

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+}
+
+"ServiceProvider" says {
+    ownsTag("image_detection_model_tag").
+
+    // ServiceProvider also opts-in to "DataEgressPolicy1"
+    "DataEgressPolicy1" canSay principalX may("egress_to_shopping_service", tagX) :-
+        isPrincipal(principalX), isTag(tagX).
+    // ServiceProvider's preference is NOT to egress:
+    "ServiceProvider" hasPreference("disallowEgress").
```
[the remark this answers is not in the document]
aferr: Yes it is indirectly related. The first line says that the service provider delegates the actual decision to allow egress to the `"DataEgressPolicy1"` principal and then the second line expresses the service provider's preference which is that the data not be egressed.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+    // Delegates TPPR the ability to apply tags anywhere
+    "ThirdPartyPrivacyReviewer" canSay hasTag(x, y) :- 
+        isAccessPath(x), isTag(y).
+
+    // Delegate the right to claim that particles are "Local" to TPPR.
+    // (We can think of the principal "LocalDevice" as a group)
+    "ThirdPartyPrivacyReviewer" canSay principalX canActAs "LocalDevice" :-
+        isPrincipal(principalX).
+
+    "LocalDevice" may("do_pure_computation", "raw_video_tag").
```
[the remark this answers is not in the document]
aferr: This is not handled as a distinct language feature. This is just a normal predicate at the moment. The language feature is avoided because we can probably get away with not having it while we are in a MVP phase.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -0,0 +1,274 @@
...
+    isPrincipal("VideoDataSource").
+    isPrincipal("ImageDetector").
+    isPrincipal("ImageSelector").
+    isPrincipal("ProductIdOutput").
+}
+
+
+"DataEgressPolicy1" says {
+    // Gathers preferences from both stakeholders:
+    "EndUser" canSay "EndUser" hasPreference(prefX) :-
```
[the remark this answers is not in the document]
aferr: The parenthesis are like this:

`"DataEgressPolicy1" says ("EndUser" canSay ("EndUser" hasPreference(prefX) :- ...)`

where the inner part `"EndUser" hasPreference(prefX)` means that "EndUse" has the preference prefx

and `"DataEgressPolicy1" says  "EndUser" canSay <FOO>` means DEP1 believes EndUser when EndUser says FOO so altogether this means that DEP1 believes EndUser when it states its own privacy preference (i.e., DEP1 trusts EU about its privacy preference or imports the privacy preference from the end user).

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: In `DEP2` unlike `DEP1`, both parties need to have the "allowEgress" preference for the data that involves the tags owned by each participant to be removed (to remove EndUser's tag the user needs to say allow and to remove the service provider's tag, the service provider needs to say allow).

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: You can't have "says" on the RHS of a rule in SecPal. (This restriction makes the language decidable).

So when rules are conditional, you always need to prove the predicates to the original speaker. In this case, since the same policy has `"EndUser" canSay "EndUser" hasPreference(<args>)` this means when `"EndUser" says "EndUser" hasPreference(...)` we also get `DEP1 says "EndUser" hasPreference(...)`

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: Yes, it's a typo. Will fix shortly.

aferr:
@bgogul btw this policy is a bit like a contrived example, and might not really map well to the "Vision Pipeline" example's needs. So if it's better, I can move this dispute aspect out of the policy and into a separate small example that gives one way of resolving disputes.

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
aferr: The preferences are used as inputs to make the decision, but arbitrary rules can go in this policy to make the decision. Other decision-making procedures that can be encoded include:
 - voting (this requires aggregation which is supported in souffle, but not yet in authorization logic)
 - deciding randomly, or randomly in some constrained way
 - using an ordering over principals

Issue #78, opened 2021-08-18
Create a script to combine datalog rules authorization logic and the manifest


bgogul:
We will need something like this to get an E2E example working.

This script should be wrapped in an skylark rule.

PR #142, opened 2021-09-22
Add rust_rules to the project's WORKSPACE file

bgogul:
Makes progress on issue #78.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -162,3 +162,21 @@ http_archive(
     strip_prefix = "glog-0.5.0",
     urls = ["https://github.com/google/glog/archive/v0.5.0.zip"],
 )
+
+#-------------------
+# Rust
+#-------------------
+http_archive(
+    name = "rules_rust",
+    sha256 = "531bdd470728b61ce41cf7604dc4f9a115983e455d46ac1d0c1632f613ab9fc3",
+    strip_prefix = "rules_rust-d8238877c0e552639d3e057aadd6bfcf37592408",
+    urls = [
+        # `main` branch as of 2021-08-23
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -162,3 +162,21 @@ http_archive(
...
+    urls = [
+        # `main` branch as of 2021-08-23
+        "https://github.com/bazelbuild/rules_rust/archive/d8238877c0e552639d3e057aadd6bfcf37592408.tar.gz",
+    ],
+)
+
+load("@rules_rust//rust:repositories.bzl", "rust_repositories")
+
+# We need this specific repository for authorization logic.
+rust_repositories(version = "nightly", iso_date = "2021-07-01", edition="2018")
```
markww, line 182: I think it's a little unfortunate to rely upon unstable features in our project. We should probably file a bug about this. Especially since the particular feature we're relying upon, try blocks, is not necessarily going to survive in its current syntax (see discussion towards end of tracking bug): https://github.com/rust-lang/rust/issues/31436 .

markww (approved):

WORKSPACE
bgogul: We will probably replace some of this code with C++ versions by the time that happens! I don't think the need for this build is related to `try` blocks, but to some other feature. I can't recall which one though. Given this context, do you think it makes sense to file a bug. Will be happy  to file it, it there is value in tracking this.

bgogul: Filed #143 and also added a TODO in any case.

markww: Yeah, you're right, it probably won't be necessary, but thanks for filing it!

PR #144, opened 2021-09-23
Build  rust/tools/authorization-logic using bazel.

bgogul:
This PR contains the following:
  - Adds some meta data to `Cargo.toml` that is required for `cargo raze`. 
  - Creates a BUILD file for `rust/tools/authorization-logic`
  - Other than that, this PR mostly contains the BUILD files that were _auto_-generated by running `cargo raze` on the `Cargo.toml`.

Makes progress towards fixing #78.

### DIFF
--- a/rust/tools/authorization-logic/BUILD
+++ b/rust/tools/authorization-logic/BUILD
@@ -1,0 +1,57 @@
+#-------------------------------------------------------------------------------
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#-------------------------------------------------------------------------------
+package(default_visibility = ["//visibility:public"])
+
+load("@rules_rust//rust:rust.bzl", "rust_binary")
+
+licenses(["notice"])
+
+rust_binary(
+    name = "authorization_logic",
+    srcs = [
+        "src/ast.rs",
+        "src/compilation_top_level.rs",
+        "src/main.rs",
+        "src/parsing/antlr_gen/authlogiclexer.rs",
+        "src/parsing/antlr_gen/authlogiclistener.rs",
+        "src/parsing/antlr_gen/authlogicparser.rs",
+        "src/parsing/antlr_gen/mod.rs",
+        "src/parsing/astconstructionvisitor.rs",
+        "src/parsing/mod.rs",
+        "src/signing/export_assertions.rs",
+        "src/signing/import_assertions.rs",
+        "src/signing/mod.rs",
+        "src/signing/tink_interface.rs",
+        "src/souffle/datalog_ir.rs",
+        "src/souffle/lowering_ast_datalog.rs",
+        "src/souffle/mod.rs",
+        "src/souffle/souffle_emitter.rs",
+        "src/souffle/souffle_interface.rs",
+        "src/test/mod.rs",
+        "src/test/test_claim_importing.rs",
+        "src/test/test_export_signatures.rs",
+        "src/test/test_queries.rs",
+        "src/test/test_signing.rs",
+    ],
+    deps = [
+        "//rust/tools/authorization-logic/cargo:antlr_rust",
+        "//rust/tools/authorization-logic/cargo:bincode",
+        "//rust/tools/authorization-logic/cargo:serde",
+        "//rust/tools/authorization-logic/cargo:tink_core",
+        "//rust/tools/authorization-logic/cargo:tink_signature",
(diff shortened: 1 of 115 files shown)


bgogul:
@markww @aferr Hold off the review on this one. I need to incorporate the invocation of build.rs into the BUILD ruies as well.

bgogul:
@aferr, @markww This is ready for review now.

rust/tools/authorization-logic/BUILD
```diff
@@ -0,0 +1,98 @@
...
+    srcs = ["src/parsing/antlr_gen/mod.rs"]
+)
+
+# Generate the parsing-related files using antlr4rust
+genrule(
+    name = "antlr_gen_srcs",
+    srcs = ["src/parsing/AuthLogic.g4"],
+    tools = [
+        ":antlr4rust_jar",
+	":antlr_gen_mod_rs",
```
markww, line 39: Nit: One of these two lines is indented incorrectly.

rust/tools/authorization-logic/BUILD
```diff
@@ -0,0 +1,98 @@
...
+        "@bazel_tools//tools/jdk:current_java_runtime",
+        "@bazel_tools//tools/jdk:java",
+    ],
+    outs = [
+       "antlr_gen/authlogiclexer.rs",
+       "antlr_gen/authlogiclistener.rs",
+       "antlr_gen/authlogicparser.rs",
+       "antlr_gen/lib.rs",
+    ],
+    cmd = "pwd; $(execpath @bazel_tools//tools/jdk:java) " +
```
markww, line 49: Why the `pwd`? Is this for letting someone know in the build log what is happening? If so, shouldn't this be part of a larger `echo` command providing some context?

rust/tools/authorization-logic/build.rs
```diff
@@ -22,13 +22,13 @@ use std::{env, error::Error, process::Command};
 fn run_antlr_on(grammar_file: &str, antlr_path: &str) -> Result<(), Box<dyn Error>> {
     let parsing_subdir = env::current_dir().unwrap().join("src/parsing");
     let full_file = grammar_file.to_owned() + ".g4";
-
+    let java_bin = env::var("JAVA").unwrap_or("java".to_string());
```
markww, line 25: Why the `to_string` here? Isn't `"java"` already a string?

rust/tools/authorization-logic/cargo/BUILD.bazel
```diff
@@ -0,0 +1,66 @@
+"""
+@generated
+cargo-raze generated Bazel file.
+
+DO NOT EDIT! Replaced on runs of cargo-raze
+"""
```
markww, line 6: Not a fan in general of checking in generated files that have to be manually regenerated; it is asking for things to go out of sync. I gather it's hard to do otherwise with Bazel's model, though, and it currently looks like the Rust side of this project isn't long for this world. I don't know that there's anything to be done here, just wish there was a better way.

markww (approved):
This looks good, modulo some minor comments. I usually prefer regenerating generated files on the fly rather than checking them in, but I gather that Bazel doesn't like that much, and we're probably getting rid of the Rust soon anyway, so not going to worry too much about that.

rust/tools/authorization-logic/BUILD
```diff
@@ -0,0 +1,98 @@
...
+#-------------------------------------------------------------------------------
+package(default_visibility = ["//visibility:public"])
+
+load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_library")
+
+licenses(["notice"])
+
+filegroup(
+    name = "antlr4rust_jar",
+    srcs = ["third_party/antlr4-rust/antlr4rust.jar"]
```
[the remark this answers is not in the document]
bgogul: Yep, I noticed that!

rust/tools/authorization-logic/BUILD
bgogul: Done.

rust/tools/authorization-logic/BUILD
bgogul: This was an oversight. I removed it.

rust/tools/authorization-logic/cargo/BUILD.bazel
bgogul: Agree with the sentiment about not checking in generated files. I did consider a few options before creating the PR:
 - Configure `cargo raze` to be runnable with a bazel  ([using-cargo-raze-through-bazel](https://github.com/google/cargo-raze#using-cargo-raze-through-bazel))
 - Define a bazel or macro that invokes `cargo raze`. 
 - Make the above macro/bazel rule to be a dependency of `authorization_logic` target. 
It required some additional work. Therefore, I decided to defer it to later if this becomes an issue.

rust/tools/authorization-logic/build.rs
bgogul: Either the type checker or the borrow checker insisted that `to_string` is needed.

rust/tools/authorization-logic/BUILD
bgogul: This was left over from debugging. Removed it now.

rust/tools/authorization-logic/BUILD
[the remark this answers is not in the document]
bgogul: Yep, I noticed that too!

PR #145, opened 2021-09-23
Support safe removal of edges

aferr:
This rule is meant to support the functionality of removing edges from
manifests:
     - that clearly does not introduce an unsafe escape hatch that could
     break policies (because it does not introduce downgrades that would not
     otherwise be possible).
     - without data owners needing to know anything about the application
     structure
     - without app behavior specification writers needing to know anything
     about the taint analysis or the tags that might appear on a handle

claimNotEdge(p, src, tgt) is an _attempt_ to remove an edge from the
analysis (by principal p)
saysDowngrades(p, path, tgt) is an _attempt_ to remove just one tag
from a path (by principal p).

The rule for downgrades(p, t) in authorization_logic.dl describes when
a downgrade is "safe" and the downgrade really happens (by consulting the
owners).

Because the analysis propagates the tags on source edges to dest edges, it is safe
to remove an edge when ALL of the tags on the source edge can be downgraded.
Therefore, a request to remove an edge is a request to remove all tags from
the source path. This rule populates a request to downgrade each tag on the
source path.
     - it is safe because these requests are only accepted if they would
     normally be by the downgrades(_, _) rule
     - it does not require the manifest writer to know anything about the
     tags in the analysis
     - data owners do not need to know about paths, and they will grant
     permission for edge removal indirectly by granting permission to remove
     tags (abstract representations of their data).

### DIFF
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -39,6 +39,7 @@
 // In this formalization, accessPaths also act as IFCLabels (in the sense that
 // they have Tags).
 .decl mayHaveTag(accessPath: AccessPath, tag: Tag)
+.decl mayHaveTagNoDown(accessPath: AccessPath, tag: Tag)
 
 // There is a claim that `accessPath` definitely has `tag` taint. This claim
 // is made by `principal`.
@@ -86,7 +87,12 @@
 //      permission for edge removal indirectly by granting permission to remove 
 //      tags (abstract representations of their data).
 
-saysDowngrades(principal, path, tag) :- claimNotEdge(principal, path, _), isTag(tag).
+mayHaveTagNoDown(tgt, tag) :- hasTag(tgt, tag).
+mayHaveTagNoDown(tgt, tag) :- 
+    edge(src, tgt), mayHaveTagNoDown(src, tag).
+
+saysDowngrades(principal, tgt, tag) :-
+    mayHaveTagNoDown(src, tag), edge(src, tgt), claimNotEdge(principal, src, tgt).
 
 hasTag(tgt, tag) :- claimHasTag(principal, tgt, tag), ownsTag(principal, tag).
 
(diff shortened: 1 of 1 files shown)


src/analysis/souffle/taint.dl
```diff
@@ -44,15 +44,57 @@
...
+// source path.
+//      - it is safe because these requests are only accepted if they would 
+//      normally be by the downgrades(_, _) rule
+//      - it does not require the manifest writer to know anything about the 
+//      tags in the analysis
+//      - data owners do not need to know about paths, and they will grant 
+//      permission for edge removal indirectly by granting permission to remove 
+//      tags (abstract representations of their data).
+
+saysDowngrades(principal, path, tag) :- claimNotEdge(principal, path, _), isTag(tag).
```
bgogul, line 89: I haven't looked at it closely, but had some quick thoughts:
 - The rule does not take into account the current tag of the path. Is that sound? 
 - Also, should the `path` match on the target of an edge? i.e., should it be claimNotEdge(principal, _, path)? Otherwise, I don't see how propagation of a tag will stop along the edge. (See `mayHaveTag` rule below.)

src/analysis/souffle/taint.dl
```diff
@@ -44,15 +44,57 @@
...
+// Manifests can produce base facts of this form, where usually the principal 
+// is a particle (and this fact is entered by a code reviewer with no knowledge 
+// of the taint analysis).
+.decl claimNotEdge(principal: Principal, src: AccessPath, dst: AccessPath)
+
 //-----------------------------------------------------------------------------
 // Rules
 //-----------------------------------------------------------------------------
+
+// This rule is meant to support the functionality of removing edges from
```
bgogul, line 56: We should start adding tests whenever we change the algorithm.

aferr:
As @markww pointed out, this has a flaw: if a particle has two outputs `o1` and `o2`, having `claimNotEdge(principal, src, o1)` will also generate **requests** to remove these tags from `o2`. This can't violate the data owner's policies though, because if these requests `saysDowngrades` are fulfilled (cause instances of `downgrades` to be proved), the data owner must have consented through `canSay downgrades`. It's still awkward to generate downgrade requests that the `claimNotEdge` did not ask for (and maybe still causes errors in practice), though so that should be fixed.

Issue #146, opened 2021-09-23
Introduce a way to perform variants of tests in Raksha.


markww:
There are a few different places where we would like to run different variants of tests in Raksha:

* The arcs manifest tests we inherited from the Arcs repo, which @bgogul pointed out have a lot of duplicated code.
* The arcs_fact_tests, in which we may want to run the same dataflow structure with different authorization logic fact variants.

This issue tracks coming up with a non-hacky way to handle these cases. If we end up needing different tools for these different cases, this issue may be split up.

bgogul:
Isn't this related or duplicate of  #64?

markww:
Thank you for pointing that one out; I thought we had an issue about this already but I didn't see it when I looked. This is related to #64 for sure. I was prompted to create this issue because I wanted to introduce variants in fact tests, whereas that talks of manifest tests. I'm going to alter this bug to make it scoped only for fact tests and let #64 track manifest tests.

PR #147, opened 2021-09-23
Use edge midpoints to make @aferr's derives from via downgrades more precise.

markww:
This PR is a riff off of @aferr's work in #145 . It builds on that work by introducing a concept of an `edge` midpoint, which can be used as a target for `downgrades`, allowing for clearing all taint that flows across some edge if the principal requesting the `downgrades` is trusted. This PR affects datalog only; if y'all like it, a follow-up PR can hook it up to `manifest2proto`.


build_defs/souffle.bzl
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #-----------------------------------------------------------------------------
 
-def souffle_cc_library(name, src, included_dl_scripts=[], visibility = None):
+def souffle_cc_library(name, src, all_principals_own_all_tags = False, included_dl_scripts=[], visibility = None):
```
bgogul, line 17: Can we restrict `all_principals_own_all_tags` to `test_only` targets somehow? Or, is it already the case?

src/analysis/souffle/dataflow_graph.dl
```diff
@@ -21,18 +21,49 @@
...
+edgeToMidpointAccessPath(src, tgt, cat("##midpoint_", to_string($))) :-
+  edge(src, tgt), claimNotEdge(_, src, tgt).
+
+// Generate inner edges to and from that midpoint for each edge with a midpoint.
+innerEdge(src, midpoint), innerEdge(midpoint, tgt) :-
+  edgeToMidpointAccessPath(src, tgt, midpoint).
+
+// For edges for which a midpoint was not generated, generate an innerEdge
+// corresponding to the original edge.
+innerEdge(src, tgt) :- edge(src, tgt), !edgeToMidpointAccessPath(src, tgt, _).
```
bgogul, line 59: I am not sure what defining a midpoint gets us. Why isn't this simply the following?
```
innerEdge(src, tgt) :- edge(src, tgt), !claimNotEdge(src, tgt).
```

src/analysis/souffle/dataflow_graph.dl
```diff
@@ -21,18 +21,49 @@
...
+// 1. The edge is not the subject of a claimNotEdge, in which case an
+// innerEdge directly corresponding to the edge is created.
+//
+// 2. The edge is the subject of a claimNotEdge, in which case three facts
+// are created: a generated midway point AccessPath, an innerEdge going from the
+// src of the original edge to the midway point, and the midway point AccessPath
+// to the tgt. That allows us to effectively eliminate edges that are the
+// subject of a claimNotEdge from the dataflow graph by eliminating all tags
+// from the midway point. Paths are defined in terms of innerEdges.
+.decl innerEdge(src: AccessPath, tgt: AccessPath)
```
bgogul, line 40: `edgeInternal` (or `internalEdge`) might be better name?

Or, even something like `resolvedEdge` might work.

src/analysis/souffle/taint.dl
```diff
@@ -44,15 +44,58 @@
...
+// request to downgrade each tag on the edge midpoint.
+//      - it is safe because these requests are only accepted if they would 
+//      normally be by the downgrades(_, _) rule
+//      - it does not require the manifest writer to know anything about the 
+//      tags in the analysis
+//      - data owners do not need to know about paths, and they will grant 
+//      permission for edge removal indirectly by granting permission to remove 
+//      tags (abstract representations of their data).
+
+saysDowngrades(principal, midpoint, tag) :-
```
bgogul, line 90: @aferr @markww  Why do we need to associate edges with downgrades at all?

Is this to connect tag ownership to edge-related claims? If so, why do we need to do that? 

Shouldn't there be a completely separate ownership involving edges/access paths?

src/analysis/souffle/tests/arcs_fact_tests/BUILD
```diff
@@ -39,3 +42,36 @@ ALL_DL_TEST_FILES = glob(["*.dl"])
         dl_script.replace(".dl", "_souffle_cc_library"),
     ],
 ) for dl_script in ALL_DL_TEST_FILES]
+
+[souffle_cc_library(
```
bgogul, line 46: Instead of creating a separate target, can we add something like `#define ALL_PRINCIPALS_OWN_ALL_TAGS` in the relevant test files?

bgogul (commented):
Left some comments and questions. PTAL.

src/analysis/souffle/taint.dl
aferr: Claims that edges are not present do roughly the same thing as downgrades -- they give a way to "relax" the taint analysis. The downgrade rule for our analysis is designed to allow information release (or similarly, the data to be transformed from sensitive to less-sensitive data) when the owner of the data permits this. It is somewhat similar to the downgrade rules in many systems for decentralized information flow control like [DLM (the orignal)](https://www.cs.cornell.edu/andru/papers/iflow-tosem.pdf) or [Flume(which is a faster read because it is a systems paper)](https://pdos.csail.mit.edu/papers/flume-sosp07.pdf) or many others. 

Because taking edges out of the analysis could inadvertantly allow information leaks that the data owner did not consent to, it is important to make sure that these don't introduce an escape hatch. One relatively clear way to ensure that is to implement the removal of edges in terms of downgrades, which should be possible.

Ideally, it would be nice formalize and prove an information flow theorem that defends against active attacks (i.e., it shows that even if other participants in the system are malicious, there are no "egresses" of data that the owner has not consented to). [Robust declassification](https://www.cs.cornell.edu/andru/papers/csfw01.pdf) is one way of formalizing this for more conventional IFC systems than ours and [this type system](https://www.cis.upenn.edu/~stevez/papers/Zda03.pdf) shows how to generalize it to decentalized label models like DLM. Unfortunately, since we have an authorization mechanism that is fundamentally new (by combining authorization logic with IFC), we would need to also invent a new property ([Opacity](https://www.researchgate.net/publication/220905175_Opacity_Analysis_in_Trust_Management_Systems) also appears related because it involves active attacks in authorization logic).

Even if we don't manage to get to the point where we are proving formal security properties, it would still be nice to give an informal description of this nice property (there are no data transformations / data "egresses" without the consent of the owner) and have this be clearly true from the rules of our system. By introducing a way to remove edges that does not go through this downgrade rule, we probably lose this informal property.

aferr:
As far as I can, tell this works correctly. I think it does not create an escape hatch that undermines the property the downgrades rule is trying to achieve. It also looks like the midpoints faithfully implement the semantics of "claimNotEdge". 

I think it would be even clearer that this is "safe" if there was some way to implement this purely in the analysis/check and without changing the edges that are presented to the analysis. I am not sure at the moment if there is a way to do that, but we can continue to iterate. (This is more about aesthetics for presenting this in a paper for example, or for making future proofs easier). I think it is OK to get this in as-is, though.

aferr (approved):

build_defs/souffle.bzl
markww: That is worth looking into. It's currently the case that `all_principals_own_all_tags` is only used in test contexts, but it is not the case that there is any automatic method restricting that to be the case.

src/analysis/souffle/dataflow_graph.dl
markww: So, first, this isn't the "interesting" case for midpoints. This is generating inner edges for the case where there was not a claim that an edge wasn't there, which produces an inner edge exactly corresponding to the original edge requested by the user.

In the example code that you wrote, we emit an `innerEdge` corresponding to an original `edge` fact when there is not a claim that the edge does not exist. The problem with this approach comes down to the principal that claims that the edge is not there and the degree to which their word is universally taken. `claimNotEdge` takes an argument indicating the principal making that claim like so:

```
.decl claimNotEdge(principal: Principal, src: AccessPath, dst: AccessPath)
```

In the above, `innerEdge`s are shared resources that all tags flow across. If we end up not producing the `innerEdge`, this blocks all tags from flowing along the original `edge` as requested by the claim, thus producing behavior as if all principals trusted the claiming `principal`. If we end up producing the `innerEdge`, all tags can then flow along the edge, producing behavior as if all principals distrusted the claiming `principal`.

The property we are trying to get with this PR is that the tag propagation behavior behaves as if the edge indicated by `claimNotEdge` is removed from the dataflow graph only for those tags belonging to principals which trust the `principal` indicated in `claimNotEdge`. The current solution accomplishes that by generating a `downgrades` said by `principal` for every `tag` in the current DL script's universe of tags on all tags flowing through this edge. Each of these produced `downgrades` will be honored only if the `tag` being downgraded can be downgraded by the `principal` attempting the downgrade. The midpoint is just a way to clear tags on an `AccessPath` on the edge without speaking of an `AccessPath` that may be shared by other edges (ie, the `src` or the `dst`).

src/analysis/souffle/dataflow_graph.dl
markww: Sure, I think those are all better. I was bothered by the potential confusion between "an edge inner to a particle" and "an edge internal to the Datalog script's reasoning about the world" after I committed this too. I like `resolvedEdge`, it clearly indicates that this is an edge that we have settled on after performing some kind of reasoning about `edge`s.

src/analysis/souffle/tests/arcs_fact_tests/BUILD
markww: Right now, I'm trying to run this test with two variants: one in which the `derives from` claim is trusted and one in which it is untrusted. I can't do that with preprocessor macros within the test. Clearly this is not the best way to do test variants, though; I've filed #146 to come up with a better way of producing test variants in the future.

src/analysis/souffle/dataflow_graph.dl
bgogul: Ah, I understand the proposal now. It seems like `midpoint` relations are where you do the downgrade. Can't we get achieve the same without the `midpoint` relation?

```
saysDowngrades(principal, tgt, tag):
    mayHaveTag(src, tag), edge(src, tgt), claimNotEdge(principal, src, tgt).
```

(I am not sure if this breaks stratification of the datalog relations for the solver.)

aferr: What @bgogul suggested looks like the answer, and we should probably use it and get rid of the midpoints :). Because it does the downgrade on the output (`tgt`) handle, but uses the input (`src`) handle to figure out which tags to use, does not have the problem of requesting downgrades on other output edges like [this one does](https://github.com/google-research/raksha/pull/145). It also removes another bit of weirdness from #145 which is that #145 would request to remove tags that are not even on the edge.

I mistakenly thought that the downgrade needed to pull tags off of the input to a particle, but pulling them off of the outputs works (I did a check on paper).

aferr: Ach! No it does break stratification :)

aferr: This PR now has a solution that regains stratifiability with a loss of precision but not safety

[#145 ](https://github.com/google-research/raksha/pull/145/files)

See also this commit message for an explanation: https://github.com/google-research/raksha/pull/145/commits/1e8496249af9e64d42e11a9dfce322be719f0fb0

See what you think. (I have no preference for which PR we use. I am just using this as a way to send code. We probably do need tests which the other PR lacks).

markww: I'm not sure that @bgogul 's suggestion works exactly. Imagine that there are two edges from different sources, `edge1` and `edge2`, each of which ends at `tgt`. Imagine that a dataflow graph without authorization logic would propagate `tag` along both edges. Imagine that there is a claim that `edge1` does not exist that is believed by the owner of `tag`. Using the rule proposed by @bgogul above, `principal` will `saysDowngrades` on `tgt`, clearing `tag` from *any* edge source, thus clearing the `tag` that should have propagated down `edge2`.

This is why I've introduced midpoints, as odd as they are: they allow clearing tag taint on a particular edge that is guaranteed to not interfere with the propagation of taint along other edges.

bgogul: I see. The issue is that we need the ability to make a different downgrade claim for each edge. The midpoints are creating "lookup keys" for each edge, which lets us make additional transformations before combining them. The results of different edges will get combined afterwards to give the desired result. 

An alternative way of achieving the same would be to add `src` to the `saysDowngrades` relation, i.e., `saysDowngrade(principal, src, tgt, tag)`, which is the effectively the same as `saysDowngrade(principal, midpoint, tag)` in the PR. We will also need to change `downgrades` to have `src`, i.e., `downgrades(src, tgt, tag)`. Isn't that much easier to understand?  The downside is that downgrades need to know about edges, which may not be ideal.

In any case, I am fine with us going forward with this PR. The "midpoint" access path seems like a useful notion to keep around as it allows us to do additional transformations along an edge. 

However, we should think of a different name as this is generally tied to another relation (`claimNotEdge`) in this case.

bgogul (approved):
I believe that there is an alternative way to achieve the same thing (as I mentioned in my comments), but let us move forward as I see the value in going with this. 

@markww can you please address the BUILD related comments as well?

src/analysis/souffle/dataflow_graph.dl
markww: Done.

build_defs/souffle.bzl
markww: I think that, while something like `additional_souffle_flags` might be valuable for other flags, this behavior deserves its own name. This behavior is very test-only, and so I'm going to be tagging the generated rules as `testonly` if this flag is set. Because this will both add the flag and alter the rule, it should have one name that does both in concert to prevent the behavior from accidentally getting sheared someplace.

markww: If `all_principals_own_all_tags` is `True`, all rules generated by the macro will be `testonly`, which should guard against this leaking into non-test builds.

src/analysis/souffle/dataflow_graph.dl
markww: Opened https://github.com/google-research/raksha/issues/158 to track this discussion.

markww:
I believe I've addressed all of the comments for which there was something to do. Going to merge this after the tests pass.

PR #149, opened 2021-09-28
Add a class to convert authorization logic into a datalog program

bgogul:
This is just a first of several PRs that incorporates the rust toolchain into C++ libraries.


src/xform_to_datalog/authorization_logic_datalog_facts_test.cc
```diff
@@ -0,0 +1,55 @@
...
+  
+    std::vector<std::string> actual_datalog =
+      absl::StrSplit(auth_facts->ToDatalog(), "\n", absl::SkipEmpty());
+    std::vector<std::string> expected_datalog =
+      ReadFileLines(test_data_dir / (program + ".dl"));
+
+    // Need to compare individual lines as the output order is non-deterministic.
+    ASSERT_THAT(actual_datalog,
+                testing::UnorderedElementsAreArray(expected_datalog));
+  }
```
markww, line 47: Rather than putting this in a loop, might be better to use `TEST_P` and have `AuthorizationLogicDatalogFactsTest` inherit from `TestWithParam`. That said, I see that there are only two elements there and you're using this in other places, so feel free to ignore this if you think the tradeoff would be uglier.

src/xform_to_datalog/authorization_logic_test.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "src/common/testing/gtest.h"
+
+namespace fs = std::filesystem;
+
+namespace raksha::xform_to_datalog {
+
+class AuthorizationLogicTest : public ::testing::Test {
+public:
+  static fs::path GetTestDataDir() {
```
markww, line 31: Why are these methods `static` and `public`? I think the preferred pattern is to make test fixtures fill in `protected` fields via their default constructor or via `SetUp`, defining all members as protected. The `protected` fields can then be accessed directly from `TEST_F` or `TEST_P` tests.

src/xform_to_datalog/authorization_logic_test.h
```diff
@@ -0,0 +1,55 @@
...
+    const char* test_srcdir_env = std::getenv("TEST_SRCDIR");
+    const char* test_workspace_env = std::getenv("TEST_WORKSPACE");
+    EXPECT_NE(test_srcdir_env,  nullptr);
+    EXPECT_NE(test_workspace_env, nullptr);
```
markww, line 35: If you take my advice to convert these to `protected` fixture methods, these might work better as `CHECK`s. In a sense, that makes semantic sense: this isn't a predicate that you're testing about the program, but an assertion you're making about the test setup. If this continues to work as is, though, don't worry about this comment.

src/xform_to_datalog/authorization_logic_datalog_facts.h
```diff
@@ -0,0 +1,45 @@
...
+namespace raksha::xform_to_datalog {
+
+class AuthorizationLogicDatalogFacts {
+ public:
+  // Creates the datalog logic facts from the given authorization logic program.
+  // Returns std::nullopt if there is any error due to processing of the program.
+  //
+  static std::optional<AuthorizationLogicDatalogFacts> create(
+      const std::string& path,							      
+      const std::string& authorization_logic_filename);
```
markww, line 31: I think both of these strings are expected to be read in their entirety, by the `create` method in the common case. As such, it's probably better to just pass a `std::string` to remove lifetime concerns and make these eligable for copy elision.

src/xform_to_datalog/authorization_logic_test.h
```diff
@@ -0,0 +1,55 @@
...
+#include "src/xform_to_datalog/authorization_logic.h"
+
+#include <cstdlib>
+#include <filesystem>
+#include <fstream>
+#include <vector>
+
+#include "src/common/testing/gtest.h"
+
+namespace fs = std::filesystem;
```
markww, line 25: We aren't allowed to use namespace aliases in headers except in function bodies.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
```diff
@@ -0,0 +1,51 @@
...
+  if (res) return std::nullopt;
+
+  // Read the file into a string.
+  std::filesystem::path output = output_dir / (program + ".dl");
+  std::ifstream file_stream(output);
+  if (!file_stream) return std::nullopt;
+
+  std::string datalog_program;
+  file_stream.seekg(0, std::ios::end);
+  datalog_program.resize(file_stream.tellg());
```
markww, line 44: Might be a bit more concise (and potentially save a `realloc` or similar under the hood) to just declare `datalog_program` a bit later and use the constructor that allocates a string with a provided size like so:

```C++
file_stream.seekg(0, std::ios::end);
std::string datalog_program(file_stream.tellg(), '\0');
```

src/xform_to_datalog/authorization_logic_datalog_facts.cc
```diff
@@ -0,0 +1,51 @@
...
+  file_stream.seekg(0, std::ios::beg);
+  file_stream.read(&datalog_program[0], datalog_program.size());
```
markww, line 46: I think `&datalog_program[0]` is more idiomatically written as `datalog_program.data()`:

https://en.cppreference.com/w/cpp/string/basic_string/data

src/xform_to_datalog/authorization_logic_datalog_facts.cc
```diff
@@ -0,0 +1,51 @@
...
+  const std::string& program_dir, const std::string& program
+) {
+  auto output_dir = std::filesystem::temp_directory_path();
+  int res = generate_datalog_facts_from_authorization_logic(
+    program.c_str(), program_dir.c_str(), output_dir.c_str());
+  if (res) return std::nullopt;
+
+  // Read the file into a string.
+  std::filesystem::path output = output_dir / (program + ".dl");
+  std::ifstream file_stream(output);
```
markww, line 39: Why are we accessing a path called `output` in a read-only fashion? Seems a little misleadingly named. Similar question about `output_dir`.

src/xform_to_datalog/authorization_logic_test.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace raksha::xform_to_datalog {
+
+class AuthorizationLogicTest : public ::testing::Test {
+public:
+  static fs::path GetTestDataDir() {
+    const char* test_srcdir_env = std::getenv("TEST_SRCDIR");
+    const char* test_workspace_env = std::getenv("TEST_WORKSPACE");
```
markww, line 33: Use of C-style `char *` is discouraged; these should be `std::string` or `const absl::string_view` instead.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
markww: Another thought: the code as written currently is safe, but filling a buffer like this is potentially laying a trap for a future maintainer to introduce a buffer overflow. It might be safer to read the file line by line as you currently do in the test fixture's `ReadFileLines` and send each line to a `std::ostringstream`, which we can then call `.str()` on. Also, it might improve performance by using fewer system calls; I'm not 100% positive but I think both `seekg` and `tellg` wrap system calls under the hood.

markww: Another issue here: `read` isn't guaranteed to get you all of the characters you asked for. For instance, if you encounter an `EOF` before reading all of your characters. That is technically possible if you have some kind of TOCTOU race on the filesystem. You might want to check `gcount` or the error bits on the stream after this `read` to prevent confusion during debugging if such a thing were to occur (assuming you continue to use the read, of course):

https://en.cppreference.com/w/cpp/io/basic_istream/read

src/xform_to_datalog/authorization_logic_datalog_facts.cc
```diff
@@ -0,0 +1,51 @@
...
+namespace raksha::xform_to_datalog {
+
+std::optional<AuthorizationLogicDatalogFacts>
+AuthorizationLogicDatalogFacts::create(
+  const std::string& program_dir, const std::string& program
+) {
+  auto output_dir = std::filesystem::temp_directory_path();
+  int res = generate_datalog_facts_from_authorization_logic(
+    program.c_str(), program_dir.c_str(), output_dir.c_str());
+  if (res) return std::nullopt;
```
markww, line 35: It looks here like this function can return `std::nullopt` in two distinct circumstances:

1. The Rust program panics (this point)
2. The `ifstream` has trouble reading the Rust program's result (line 40).

These are very different cases, but currently they do not have logging indicating which has occurred. I think we should put some log messages in both of these cases to ease debugging in the future.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
markww: Also, we might want to check the value of `tellg` before sending it in here. According to the docs, `tellg` can return `-1` in at least some circumstances:

https://en.cppreference.com/w/cpp/io/basic_istream/tellg

But `resize` takes a `size_t`, which is unsigned, as its parameter. As such, there may exist a circumstance in which we may accidentally request a resize to `ULONG_MAX`.

src/xform_to_datalog/authorization_logic_test.cc
```diff
@@ -21,35 +21,12 @@
 #include <vector>
 
 #include "src/common/testing/gtest.h"
+#include "src/xform_to_datalog/authorization_logic_test.h"
 
 namespace fs = std::filesystem;
```
markww, line 26: When declaring namespace aliases, we're supposed to always place them in the innermost namespace (so inside `namespace raksha::xform_to_datalog {` here).

src/xform_to_datalog/authorization_logic_datalog_facts_test.cc
```diff
@@ -0,0 +1,55 @@
...
+
+namespace fs = std::filesystem;
+
+namespace raksha::xform_to_datalog {
+
+class AuthorizationLogicDatalogFactsTest : public AuthorizationLogicTest {};
+
+TEST_F(AuthorizationLogicDatalogFactsTest, InvokesRustToolAndGeneratesOutput) {
+  for (const std::string& program : {"simple_auth_logic", "empty_auth_logic"}) {
+    const fs::path& test_data_dir = GetTestDataDir();
```
markww, line 35: No need to use a reference here. We are returning a temporary value from `GetTestDataDir`, so it is friendlier to copy elision to make this just `fs::path test_data_dir`.

src/xform_to_datalog/authorization_logic_datalog_facts_test.cc
bgogul: I did consider it. Unfortunately, this relies on  some methods defined in `AuthorizationLogicTest` (which in turn inherits from `::testing::Testing`). I have filed #150 to clean it up. I will move to a parameterized test at that time.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
bgogul: Good point about `tellg` and `gcount`. I made the changes. 

Won't `std::string datalog_program(file_stream.tellg(), '\0');` also generate instructions for zeroing out the bytes? It is not a big deal, but just wondering if it would essentially be the same.

src/xform_to_datalog/authorization_logic_datalog_facts.h
bgogul: We don't make a copy of these strings internally. If that is the case, isn't `const &` preferred?

Even in the current code, we pass in some strings that need not be copied.

src/xform_to_datalog/authorization_logic_datalog_facts_test.cc
bgogul: Done. However, won't this get RVOed even with `const &` assignment?

src/xform_to_datalog/authorization_logic_test.h
bgogul: Removed it.

src/xform_to_datalog/authorization_logic_test.h
bgogul: Done.

src/xform_to_datalog/authorization_logic_test.h
bgogul: Done.

src/xform_to_datalog/BUILD
```diff
@@ -18,19 +18,45 @@ licenses(["notice"])
 cc_library(
     name = "authorization_logic",
     hdrs = ["authorization_logic.h"],
+    linkstatic = True,
```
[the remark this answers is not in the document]
bgogul: `linkstatic` is not needed here. I have removed it.

bgogul (commented):
@markww, addressed your comments. PTAL.

src/xform_to_datalog/authorization_logic_datalog_facts_test.cc
markww: Good enough for me.

src/xform_to_datalog/authorization_logic_datalog_facts.h
markww: You are right,this is not a good case for copy elision, sorry for the red herring. In this case, however, the standard recommendation is to use `absl::string_view` rather than `const std::string &` for function parameters.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
markww: It would issue instructions to zero the array, yes. However, `resize` claims to *also* add characters to the array:  https://en.cppreference.com/w/cpp/string/basic_string/resize . It says that if resize expands the size of the string, it adds `CharT()` to the end. `cppreference` doesn't provide much guidance as to what `CharT()` evaluates to, but I'm guessing it's '\0'.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
markww: Sounds good, that works for me.

markww (commented):
LGTM. A few comment responses for you to consider, but nothing major.

src/xform_to_datalog/authorization_logic_datalog_facts_test.cc
markww: It's less about the performance here and more about other functions that `test_data_dir` might be used as an argument to. In those cases, having it be a value rather than a reference/pointer gives the compiler more leeway to elide copies at the callsite.

markww (approved):

src/xform_to_datalog/authorization_logic_datalog_facts.cc
bgogul: Done.

PR #153, opened 2021-09-29
Sort the declarations in souffle_emitter

bgogul:
The order of decls are not consistent across runs, which makes the tests brittle. This PR makes the order of decls  consistent by sorting them.


markww (approved):
LGTM

PR #154, opened 2021-09-29
Make Cargo Test Great Again

aferr:
Just prior to this commit cargo test would fail. It looks like this is
the commit where the regression began:
[PR 148](https://github.com/google-research/raksha/pull/148).
Rustdoc executes code examples, causing syntax errors to be thrown because these comments are not
rust code. This fixes that by making rustdoc aware that this code is actually datalog.

### DIFF
--- a/rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
+++ b/rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
@@ -86,16 +86,18 @@
 //! Neither Souffle nor SecPal support queries, but this language does.
 //! Queries check of a single unconditional assertion is true. A query of
 //! the form:
-//!
-//!      // Q_NAME = query ASSERTION(<ARGS>) ?
+//! ```Datalog
+//!      Q_NAME = query ASSERTION(<ARGS>) ?
+//! ```
 //!
 //! is translated into an assertion with just one argument, and this
 //! assertion is made an output, so we can simply check the CSV that
 //! souffle emits as in:
-//!
-//!      // Q_NAME("dummy_var") :- grounded("dummy_var"), ASSERTION(<ARGS>).
-//!      // grounded("dummy_var").
-//!      // .output Q_NAME
+//! ```Datalog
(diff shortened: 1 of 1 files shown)


rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
```diff
@@ -87,15 +87,15 @@
...
 //! is translated into an assertion with just one argument, and this
 //! assertion is made an output, so we can simply check the CSV that
 //! souffle emits as in:
 //!
-//!      Q_NAME("dummy_var") :- grounded("dummy_var"), ASSERTION(<ARGS>).
-//!      grounded("dummy_var").
-//!      .output Q_NAME
+//!      // Q_NAME("dummy_var") :- grounded("dummy_var"), ASSERTION(<ARGS>).
+//!      // grounded("dummy_var").
+//!      // .output Q_NAME
```
markww, line 98: I'm not sure if you're aware of this feature, but Rust likes to make sure that the code that you put in as examples are actually valid Rust:

https://doc.rust-lang.org/rustdoc/documentation-tests.html

It appears to do this whenever there is a Markdown code block, and assumes that any code without a language attached is Rust. Therefore, another way to solve this might be to explicitly put in a language other than Rust, like so:

//! ```Datalog
//!      Q_NAME("dummy_var") :- grounded("dummy_var"), ASSERTION(<ARGS>).
//!      grounded("dummy_var").
//!      .output Q_NAME
//! ```

There is no guarantees that Rustdoc will be aware of Datalog syntax highlighting, but the above would be most correct if it did. If not, any language (is there a text one?) would suppress Rust trying to ensure that the code compiles and runs.

That said, your commenting solution also works, so I'm fine with it as is if you are ok with it currently.

markww (approved):
See comment about `rustdoc`'s example code execution feature. But the kluge you made is also fine for our purposes right now.

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
aferr: OK that works. So it's that way :)

PR #155, opened 2021-09-29
Refactor facts to instantiate with pointers rather than copies.

markww:
Previously, we have instantiated facts such as checks, claims, and edges
by creating a copy of the original fact with a new AccessPathRoot
indicating the instantiated root. This was always dubious from
a performance standpoint. However, as I started another PR in which
I was introducing we have started introducing complex checks with
predicate trees pointed to by unique_ptrs, the memory ownership
considerations became too complex to keep the status quo.

This commit instead introduces a wrapper for facts that may not be
instantiated called InstanceFact. InstanceFact contains a variant that
associates a pointer to the original fact with an Instantiator, which
contains the information needed to replace the uninstantiated root with
an instantiated root. There is another variant for facts that were
instantiated initially that just wraps that fact and implicitly
associates it with a NoopInstantiator (which just asserts that the roots
it encounters are already instantiated).

Rather than having an instantiated copy, Instantiators allow for
instantiating AccessPathRoots as they are encountered, preventing a copy
from being necessary.

The use of pointers rather than copies introduced lifetime
considerations in a number of places where they were not present before,
most notably in ManifestDatalogFacts, which now must store a list of
unique_ptrs to ParticleSpecs and Instantiators that it relies upon being
live at least as long as its component InstanceFacts.

### DIFF
--- a/src/ir/instance_fact.h
+++ b/src/ir/instance_fact.h
@@ -1,0 +1,119 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#ifndef SRC_IR_INSTANCE_FACT_H_
+#define SRC_IR_INSTANCE_FACT_H_
+
+#include "src/ir/instantiator.h"
+#include "src/ir/noop_instantiator.h"
+
+// An InstanceFact is an instantation of some other fact that may be
+// initially not fully instantiated.
+//
+// Consider, for instance, the case of checks and claims on a ParticleSpec.
+// Checks and claims describe the check or claim they wish to perform on
+// AccessPaths rooted at the ParticleSpec. At some point, however, they must be
+// printed out as performing those claims and checks upon the access paths of
+// an actual instantiated particle.
+//
+// To do this, we create an InstanceFact upon the check or claim (or any
+// other instantiatable fact). The InstanceFact points at the original,
+// uninstantiated fact, and an Instantiator, which is just an object that
+// will give us a map from uninstantiated AccessPath roots to instantiated
+// AccessPath roots. Using the mapping provided by the instantiator, we can
+// transform details of the uninstantiated fact to be instantiated on the fly.
+// This prevents having to copy the data initially on the uninstantiated
+// fact, sidestepping the pointer ownership quandries and performance
+// penalties we may otherwise have to deal with.
+
+namespace raksha::ir {
+
+// This class represents an instantiation of a fact that was originally
+// stated about some uninstantiated root (such as a HandleConnectionSpec)
+// using an Instantiator.
+template<class T>
+class InstantiatedInstanceFact {
+ public:
+  InstantiatedInstanceFact(
+      const T &original_fact, const Instantiator &instantiator)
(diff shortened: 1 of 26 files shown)


src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -17,9 +34,9 @@ namespace raksha::xform_to_datalog::arcs_manifest_tree {
 // This struct is used to return the members of a ParticleSpec after being
 // instantiated.
 struct InstantiatedParticleSpecFacts {
-  std::vector<raksha::ir::TagClaim> tag_claims;
-  std::vector<raksha::ir::TagCheck> checks;
-  std::vector<raksha::ir::Edge> edges;
+  std::vector<raksha::ir::InstanceFact<raksha::ir::TagClaim>> tag_claims;
```
bgogul, line 37: Do we need to have the `InstanceFact`? What is the advantage? Isn't the instantiated class representable using `TagClaim` itself?

src/ir/instance_fact.h
```diff
@@ -0,0 +1,119 @@
...
+// This prevents having to copy the data initially on the uninstantiated
+// fact, sidestepping the pointer ownership quandries and performance
+// penalties we may otherwise have to deal with.
+
+namespace raksha::ir {
+
+// This class represents an instantiation of a fact that was originally
+// stated about some uninstantiated root (such as a HandleConnectionSpec)
+// using an Instantiator.
+template<class T>
```
bgogul, line 47: I am not sure we gain too much value from this and other wrapper classes in this file. Isn't the instantiated class the same type as the uninstantiated class?

src/ir/instance_fact.h
```diff
@@ -0,0 +1,119 @@
...
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#ifndef SRC_IR_INSTANCE_FACT_H_
+#define SRC_IR_INSTANCE_FACT_H_
+
+#include "src/ir/instantiator.h"
+#include "src/ir/noop_instantiator.h"
+
+// An InstanceFact is an instantation of some other fact that may be
```
bgogul, line 23: If I understand the PR correctly, we are keeping a `pair<Fact, Instantiator>`, where the `Instantiator` has enough information to create an instance of the `Fact` later as needed. Is that right? 

Given that the classes we have are basically a part of some AST,  all of this seems more complicated than necessary. What do we gain by using the instantiators because we still need to instantiate them later as needed?

markww: Not quite. We don't create an instance later, the pair of the `Instantiator` and the original fact *is* the instance. We just use the `Instantiator` to substitute the `AccessPathRoot` on the fly while performing operations where we need the instantiation instead of the spec version.

Previously, we had a spec versions of claims, checks, and edges on `ParticleSpec`s that had one or more `AccessPath`s with `AccessPathRoot`s indicating the `ParticleSpec`. If you accidentally called `ToDatalog()` on that fact, would get a runtime error. For each Particle, we would create an instantiated copy that was almost identical to the original spec fact but with the spec `AccessPathRoot` replaced with an instance `AccessPathRoot`. If we had N particles, we would create N copies, and the only difference between those copies would be the `AccessPathRoot`s in the contained `AccessPath`s.

As of this PR, we create no copies for instances. We augment certain methods that expect to only be called on instantiated facts (such as `ToDatalog`) to take an `Instantiator`, and they run on the original uninstantiated spec fact. The difference is that when these methods encounter an `AccessPathRoot`, they perform whatever operation they intended to perform upon the replacement `AccessPathRoot` provided by the `Instantiator` instead of the original one.

Not making copies greatly simplifies the ownership story if these facts have a `unique_ptr` hanging off of them, which will be the case once `Predicate` trees are added to claims. In addition, it addresses some of the concerns you had about making copies of variable-sized `AccessPath`s everywhere.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: Yes, but doing so would require copying the `TagClaim`, which already contains a variable-sized `AccessPath` and will soon contain a `unique_ptr<Predicate>`.

markww:
This approach is probably too confusing to be the right way forward. Leaving as draft so we can use the code if some parts of it are useful for future work.

Issue #102, opened 2021-09-01
Resolve Difference Between Types in Taint Analysis and Auth Logic to allow integration


aferr:
The authorization logic does not use types in its souffle output (all arguments are of type `symbol`), but the taint analysis for Raksha does.

To make integration possible, we need to resolve the difference between types for the relation where they interface which is `says_downgrades`. In the `.dl` files produced by the authorization logic this is declared as

`.decl says_downgrades(x0: symbol, x1: symbol, x2: symbol)`

but in `src/analysis/souffle/authorization_logic.dl` this is declared as

`.decl says_downgrades(speaker: Principal, p: AccessPath, t: Tag)`

Here are a few ways to fix this:
1) Support types and type declarations in the authorization logic
2) (Temporarily) remove the types from the relevant parts of the taint analysis (this would ripple through to everything with AccessPath)
3) Write a script that just removes the type declarations for `says_downgrades` from the auth logic outputs so that just the declaration in `authorization_logic.dl` is used. Other relation declarations in the auth logic code should remain the same. I think that this should work, but would need to try it out.

Probably option 3) is easiest, but I figured I would ask what is preferred.

aferr:
@bgogul @markww

aferr:
@markww mentioned https://souffle-lang.github.io/types#type-conversion which seems potentially helpful for this
@bgogul also mentioned that it could be good to pass a file to auth logic that lists relations to skip declaring.

(wanted to just list these both before I forget)

bgogul:
> @bgogul also mentioned that it could be good to pass a file to auth logic that lists relations to skip declaring.

This might be the simplest? You pass a list of relations names and consult it before adding any declarations to the generated dl file?

Issue #8, opened 2021-07-01
[Authorization Logic] Look Into StructOpt for command line args


aferr:
[StructOp](https://docs.rs/structopt/0.3.21/structopt/)

This is also possibly a good opportunity to make the way inputs and outputs are specified a bit more conventional.

Suggested by: @conradgrobler @tiziano88

bMacSwigg:
I think this was done in #461 and #462 (as re-implementation of #245), so I'm marking this closed

aferr:
Thanks @bMacSwigg for both fixing this and properly taking managing the issue :)

PR #156, opened 2021-09-30
    Command line option to skip list of declarations

aferr:
This implements
    [#42](https://github.com/google-research/raksha/issues/102).
    There is now a fourth optional command line paramater which is a
    comma-separated list of strings that name relation declarations that the
    souffle emitter will skip when generating souffle code. If no list is
    given, this works as before. This exists for integration with the rest of
    Raksha. When we have
    [#8](https://github.com/google-research/raksha/issues/8), this should
    become a proper command-line option.

### DIFF
--- a/rust/tools/authorization-logic/src/compilation_top_level.rs
+++ b/rust/tools/authorization-logic/src/compilation_top_level.rs
@@ -38,9 +38,11 @@ fn source_file_to_ast(filename: &str, in_dir: &str) -> AstProgram {
     astconstructionvisitor::parse_program(&source[..])
 }
 
-pub fn compile(filename: &str, in_dir: &str, out_dir: &str) {
+pub fn compile(filename: &str, in_dir: &str, out_dir: &str,
+            decl_skip: &Vec<&str>) {
     let prog = source_file_to_ast(filename, in_dir);
     let prog_with_imports = import_assertions::handle_imports(&prog);
-    souffle_interface::ast_to_souffle_file(&prog_with_imports, filename, out_dir);
+    souffle_interface::ast_to_souffle_file(&prog_with_imports, filename,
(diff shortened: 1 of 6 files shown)


rust/tools/authorization-logic/src/compilation_top_level.rs
```diff
@@ -38,9 +38,11 @@ fn source_file_to_ast(filename: &str, in_dir: &str) -> AstProgram {
     astconstructionvisitor::parse_program(&source[..])
 }
 
-pub fn compile(filename: &str, in_dir: &str, out_dir: &str) {
+pub fn compile(filename: &str, in_dir: &str, out_dir: &str,
+            decl_skip: &Option<Vec<String>>) {
```
bgogul, line 42: It is straightforward to pass a string to a Rust library call from `C++`. However, I am not sure how we can do the same for rust vectors. 
Can we have a version where `decl_skip` is a comma-separated or newline-separated list of declarations?  e.g., 
`blah, some, foo`
or
```
blah
some
foo
```

bgogul (commented):
Looks good, except for a minor API change.

rust/tools/authorization-logic/src/compilation_top_level.rs
aferr: That's how the main function already works. It's just the compile function that takes a vector. I guess I can move the bit in the main that converts the string into compile.

aferr: OK fixed. PTAL @bgogul

bgogul (approved):
Tests are failing. You will need to update the corresponding `cc_library`  See https://github.com/google-research/raksha/blob/main/src/xform_to_datalog/authorization_logic.h

PR #157, opened 2021-09-30
Add predicate trees; check that expected output strs are valid Datalog.

markww:
This commit adds the Predicate class and its various subclasses, which
allow expressing complex predicates upon the presence of tags. It also
adds a ToDatalogRuleBody method for this class which prints out the
predicate as the rule body for a check.

The expected output strings for these Predicates are pretty complex and
somewhat difficult to make into valid Souffle datalog, so this commit
also adds a tool that will, given a vector of Datalog strings provided
to it via linking magic, print those strings out to a datalog file and
run Souffle on them to ensure that they are valid datalog. Although it
currently only works on rule bodies, it could easily be extended for any
kind of snippet of Datalog.


src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+namespace raksha::ir::proto {
+
+namespace ir = raksha::ir;
+namespace pred = raksha::ir::predicates;
+
+std::unique_ptr<pred::Predicate> Decode(
+    const arcs::InformationFlowLabelProto_Predicate &predicate_proto) {
+  switch (predicate_proto.predicate_case()) {
+    case arcs::InformationFlowLabelProto_Predicate::kLabel: {
+      const arcs::InformationFlowLabelProto &label = predicate_proto.label();
```
bgogul, line 32: `const auto &label`?

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+      CHECK(label.has_semantic_tag())
+        << "Found a label without required field tag.";
+      return std::make_unique<pred::TagPresence>(label.semantic_tag());
+    }
+    case arcs::InformationFlowLabelProto_Predicate::kAnd: {
+      const arcs::InformationFlowLabelProto_Predicate_And &and_predicate =
```
bgogul, line 38: `const auto &and_predicate` ?

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+          predicate_proto.and_();
+      CHECK(and_predicate.has_conjunct0())
+        << "Found an And predicate without required field conjunct0.";
```
bgogul, line 41: nit: an `And` predicate ...

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+      CHECK(and_predicate.has_conjunct1())
+        << "Found an And predicate without required field conjunct1.";
+      std::unique_ptr<pred::Predicate> conjunct0_predicate =
+          Decode(and_predicate.conjunct0());
+      std::unique_ptr<pred::Predicate> conjunct1_predicate =
+          Decode(and_predicate.conjunct1());
+      return std::make_unique<pred::And>(
+          std::move(conjunct0_predicate), std::move(conjunct1_predicate));
```
bgogul, line 49: I wonder if you can avoid having to specify `move` if you invoke Decode as part of a function argument:
```
return std::make_unique<pred::And>(
    Decode(and_predicate.conjunct0()), Decode(and_predicate.conjunct1()));
```

If so, change this an other places?

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+    }
+    case arcs::InformationFlowLabelProto_Predicate::kImplies: {
+       const arcs::InformationFlowLabelProto_Predicate_Implies
```
bgogul, line 52: `const auto & implies_predicate`?

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+           Decode(implies_predicate.antecedent());
+       std::unique_ptr<pred::Predicate> consequent_predicate =
+           Decode(implies_predicate.consequent());
+
+       return std::make_unique<pred::Implies>(
+           std::move(antecedent_predicate), std::move(consequent_predicate));
+
+    }
+    case arcs::InformationFlowLabelProto_Predicate::kNot: {
+      const arcs::InformationFlowLabelProto_Predicate_Not &not_predicate =
```
bgogul, line 68: `const auto &not_predicate`?

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,94 @@
...
+          predicate_proto.not_();
+      CHECK(not_predicate.has_predicate())
+        << "Found a Not predicate without required field predicate.";
+      return std::make_unique<pred::Not>(Decode(not_predicate.predicate()));
+    }
+    case arcs::InformationFlowLabelProto_Predicate::kOr: {
+      const arcs::InformationFlowLabelProto_Predicate_Or &or_predicate =
```
bgogul, line 75: `const auto & or_predicate`?

src/test_utils/dl_string_extractor/datalog_string.h
```diff
@@ -0,0 +1,43 @@
...
+// A string that is a snippet of datalog, marked with the grammatical part of
+// Datalog it represents.
+class DatalogString {
+ public:
+  DatalogString(DatalogStringKind kind, std::string dl_string)
+    : kind_(kind), dl_string_(std::move(dl_string)) { }
+
+  DatalogStringKind kind() const { return kind_; }
+
+  std::string dl_string() const { return dl_string_; }
```
bgogul, line 35: `const std::string&`

src/ir/predicates/BUILD
```diff
@@ -0,0 +1,58 @@
...
+    hdrs = ["predicate.h"],
+    deps = [
+        "//src/ir:access_path",
+        "@absl//absl/strings",
+        "@absl//absl/strings:str_format",
+    ],
+)
+
+cc_library(
+    name = "predicate_textproto_to_rule_body_testdata",
```
bgogul, line 34: testonly = True,

src/ir/predicates/predicate_test.cc
```diff
@@ -0,0 +1,88 @@
...
+  std::tuple<std::tuple<absl::string_view, absl::string_view>, AccessPath>> {
+ public:
+  ToDatalogRuleBodyTest() :
+    textproto_(std::get<0>(std::get<0>(GetParam()))),
+    access_path_(std::get<1>(GetParam())) {
+    absl::string_view rule_body_sub_string =
+        std::get<1>(std::get<0>(GetParam()));
+    expected_rule_body_ =
+        absl::Substitute(rule_body_sub_string, access_path_.ToString());
+  }
```
bgogul, line 42: nit: any way to make this more readable?

bgogul (commented):
Haven't looked at it fully. Just sending current set of comments.

src/ir/predicates/predicate.h
```diff
@@ -0,0 +1,155 @@
...
+#include "absl/strings/str_format.h"
+#include "absl/strings/string_view.h"
+#include "src/ir/access_path.h"
+
+// This file includes Predicate and its subclasses. These classes are boolean
+// expressions upon the presence of tags. We place them all here in one file
+// because they are small classes and related enough that it improves
+// readability to have them in one place.
+
+namespace raksha::ir::predicates {
```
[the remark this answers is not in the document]
markww: I guess not really; I can merge this into `ir`. I was initially thinking there would be more `.h` and `.cc` files, but it turned out to be just one header.

src/ir/predicates/predicate_textproto_to_rule_body_testdata.h
```diff
@@ -0,0 +1,99 @@
...
+namespace raksha::ir::predicates {
+// Note: StrFormat is careful not to allow dynamically constructed format
+// strings. In the few cases that it does allow format strings to come from
+// dynamic construction, it forces them to have a type indicating the format
+// specifiers at static time through a template. Instead, we use Substitute,
+// which is better suited for the purpose that we want. Therefore, access
+// path substitution points are indicated with $0 instead of %s.
+static constexpr std::tuple<absl::string_view, absl::string_view>
+predicate_textproto_to_rule_body_format[] = {
+    { R"(label: { semantic_tag: "tag1"})",
```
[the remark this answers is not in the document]
markww: Sure, that makes sense.

src/ir/proto/predicate.cc
markww: In general, I erred towards writing out the exact type rather than using type inference, as it seems like that's what the Google C++ style guide prefers. See https://google.github.io/styleguide/cppguide.html#Type_deduction .

src/ir/proto/predicate.cc
markww: See other comment about `auto` use and style guide.

src/ir/proto/predicate.cc
markww: See other comment about `auto` use and style guide.

src/ir/proto/predicate.cc
markww: See other comment about `auto` use and style guide.

src/ir/proto/predicate.cc
markww: See other comment about `auto` use and style guide.

src/ir/proto/predicate.cc
markww: Good point, I'll see what I can do there.

src/test_utils/dl_string_extractor/datalog_string.h
markww: ah, good catch. Might use an `absl::string_view` here instead.

src/ir/predicates/predicate_test.cc
markww: I'll see what I can do. Since I've basically been forced to write a constructor body anyway, I might do some more of this work there rather than in the per-field initializers.

src/ir/predicates/predicate.h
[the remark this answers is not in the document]
markww: Done.

src/test_utils/dl_string_extractor/datalog_string.h
markww: Done.

src/ir/predicates/BUILD
markww: Done.

src/ir/predicates/predicate_textproto_to_rule_body_testdata.h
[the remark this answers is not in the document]
markww: Done.

src/ir/proto/predicate.cc
markww: Done.

src/ir/proto/predicate.cc
markww: Do you mean that I should quote the `And` with single quotes?

src/ir/predicates/predicate_test.cc
markww: I think I improved it by moving most stuff out of the field initializers. I had to leave the one for `access_path_` there because it has no default constructor. I think it's a pretty big improvement though.

markww:
Addressed most of the comments you left so far, and am ready for the next round when your workday starts again.

src/ir/proto/predicate.cc
bgogul: SG. It is fine to have the type explicitly specified. However, `const auto &` would have been OK here. The case label `arcs::InformationFlowLabelProto_Predicate::kLabel` already provides an indication of what the type is. Therefore, it doesn't hinder readability. In fact, I think it would have improved readability by making the code concise.

src/test_utils/dl_string_extractor/dl_string_parsing_test_driver.cc
```diff
@@ -0,0 +1,41 @@
...
+//
+// This file is a simple driver to run datalog scripts generated from
+// extracting raw strings representing parts of a datalog program to ensure
+// that they parse correctly.
+//
+#include <memory>
+
+#include "souffle/SouffleInterface.h"
+
+int main(int argc, char **argv) {
```
bgogul, line 25: Not for this PR, but we should consolidate the different test drivers into one place (or even one binary). I have filed #160. Can you add a TODO here?

src/ir/predicate_textproto_to_rule_body_testdata.h
```diff
@@ -0,0 +1,122 @@
...
+// specifiers at static time through a template. Instead, we use Substitute,
+// which is better suited for the purpose that we want. Therefore, access
+// path substitution points are indicated with $0 instead of %s.
+static constexpr std::tuple<absl::string_view, absl::string_view>
+predicate_textproto_to_rule_body_format[] = {
+    // Each of the tests is preceded by a snippet of what the predicate
+    // might look like in an Arcs manifest file, with the AccessPath
+    // represented by ap.
+
+    // ap is tag1
```
bgogul, line 35: Thanks for reformatting this. This looks much more readable!

bgogul (approved):

build_defs/souffle.bzl
```diff
@@ -14,21 +14,35 @@
 # limitations under the License.
 #-----------------------------------------------------------------------------
 
-def souffle_cc_library(name, src, all_principals_own_all_tags = False, included_dl_scripts = [], visibility = None):
+def souffle_cc_library(
+        name,
+        src,
+        all_principals_own_all_tags = False,
```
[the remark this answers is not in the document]
markww: Acknowledged.

src/ir/BUILD
```diff
@@ -1,10 +1,32 @@
...
 package(default_visibility = ["//src:__subpackages__"])
 
+load("//build_defs:test_helpers.bzl", "extracted_datalog_string_test")
+
 cc_library(
     name = "ir",
     srcs = glob(
         ["*.cc"],
-        exclude = ["*_test.cc"],
+        exclude = [
```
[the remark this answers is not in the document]
markww: Yes, let's do a followup. Filed https://github.com/google-research/raksha/issues/164 about the general, whole-repo instance of the problem.

src/test_utils/dl_string_extractor/dl_string_parsing_test_driver.cc
markww: Yes, sounds good.

src/ir/predicate_textproto_to_rule_body_testdata.h
markww: You're welcome! I agree, definitely an improvement.

src/ir/proto/predicate.cc
```diff
@@ -0,0 +1,81 @@
...
+
+#include "src/ir/proto/predicate.h"
+
+#include <memory>
+
+#include "src/ir/predicate.h"
+
+namespace raksha::ir::proto {
+
+namespace ir = raksha::ir;
```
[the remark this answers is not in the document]
markww: ok, I'll clean this up.

src/ir/proto/predicate.cc
markww: Done.

src/ir/proto/predicate.cc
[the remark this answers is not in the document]
markww: Done.

src/test_utils/dl_string_extractor/dl_string_parsing_test_driver.cc
markww: Done.

PR #161, opened 2021-10-01
Add datalog facts from authorization logic into main entry point.

bgogul:


src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -59,6 +68,8 @@ INSTANTIATE_TEST_SUITE_P(
                       true,
                       "tag")},
                 {}, {}),
+            *(AuthorizationLogicDatalogFacts::create(
+              AuthorizationLogicTest::GetTestDataDir().c_str(), "simple_auth_logic")),
```
markww, line 72: Similar comment about `c_str` here; use implicit conversion or `native()` or change the signature to remove concern entirely?

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -16,23 +16,30 @@
...
   EXPECT_EQ(datalog_facts.ToDatalog(), expected_string);
 }
 
 INSTANTIATE_TEST_SUITE_P(
     DatalogFactsTest, DatalogFactsTest,
     testing::Values(
-        std::make_pair(ManifestDatalogFacts({}, {}, {}),
+        std::make_tuple(ManifestDatalogFacts({}, {}, {}),
+                        *(AuthorizationLogicDatalogFacts::create(
+                          AuthorizationLogicTest::GetTestDataDir().c_str(), "empty_auth_logic")),
```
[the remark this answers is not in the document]
markww: Actually, there might be an even better solution: perhaps `AuthorizationLogicDatalogFacts::create` should take a `std::sys::filepath` as its argument and prevent callers from having to consider this aspect at the callsite.

If you do take this option, there's a small nit within `create`: I belive that `absl::string_view` has an autoconversion to `const char *`, so at least one of the calls to `data()` in there may be unnecessary.

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -72,6 +83,12 @@ claimHasTag("particle", "recipe.particle.out", "tag").
 
 // Edges:
 
+.decl grounded_dummy(x0: symbol)
```
markww, line 86: Why are these authorization logic facts in a section commented as `Edges:`? Is this an unintentional result of putting these facts at the end without their own comment header?

src/xform_to_datalog/generate_datalog_program.cc
```diff
@@ -71,10 +72,11 @@ int main(int argc, char* argv[]) {
 
   auto datalog_facts = raksha::xform_to_datalog::DatalogFacts(
       raksha::xform_to_datalog::ManifestDatalogFacts::CreateFromManifestProto(
-          manifest_proto));
+          manifest_proto),
+      raksha::xform_to_datalog::AuthorizationLogicDatalogFacts(""));
```
markww, line 76: This will be filled in with a non-empty string in a follow-on PR?

src/xform_to_datalog/datalog_facts_test.cc
bgogul: Yes, there was no comment header for `Authorization Logic`. It should be fixed now.

src/xform_to_datalog/generate_datalog_program.cc
bgogul: Yes, I have a PR that will fill this in. Without this change, the tests won't pass.

src/xform_to_datalog/datalog_facts.h
```diff
@@ -32,17 +33,23 @@ namespace raksha::xform_to_datalog {
...
+      auth_logic_datalog_facts_(std::move(auth_logic_datalog_facts))
+  {}
 
   // Returns the datalog program with necessary headers.
   std::string ToDatalog() const {
     return absl::StrCat(
-        kDatalogFilePrefix, manifest_datalog_facts_.ToDatalog());
+        kDatalogFilePrefix,
+        manifest_datalog_facts_.ToDatalog(),
+        auth_logic_datalog_facts_.ToDatalog());
```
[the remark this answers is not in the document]
bgogul: Fixed this by using `absl::StrFormat` that ensures new lines are always present.

src/xform_to_datalog/datalog_facts_test.cc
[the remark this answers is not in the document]
bgogul: Good idea. See  #165  that changes the API.

src/xform_to_datalog/datalog_facts_test.cc
bgogul: `::create` takes an `fs::path` now.

src/xform_to_datalog/datalog_facts_test.cc
bgogul: Yes, there was no header. It should be fixed now.

markww (approved):

src/xform_to_datalog/generate_datalog_program.cc
```diff
@@ -71,10 +72,11 @@ int main(int argc, char* argv[]) {
...
 
-  std::ofstream datalog_file(
-      datalog_filepath, std::ios::out | std::ios::trunc | std::ios::binary);
+  std::ofstream datalog_file(datalog_filepath, std::ios::out | std::ios::trunc |
+                                                   std::ios::binary);
```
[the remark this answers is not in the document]
bgogul: this was the output of `clang-format`. So, I will leave it as is.

PR #163, opened 2021-10-01
Minor Changes for Demo

aferr:
This commit makes minor changes in preparation for a demo of the vision
pipeline example.
 - Some distracting comments are removed
 - Some distracting queries that are just there for tests are removed
 - The arcs manifest is moved back to where it can be built
 - A tag that is not mentioned in the auth logic part of the example is
   removed from the manifest

This still works. Instructions to run the demo to be given elsewhere.


third_party/arcs/examples/VisionPipeline.arcs
```diff
@@ -21,15 +21,14 @@
 // The "source" and "output" particles exist only because "claims" can
 // appear just on outputs and "checks" can appear only on inputs.
 
-particle VideoDataSource
```
aferr, line 24: Here I just changed the name from VideoDataSource to DataSource because this actually is a source of a "model" as well. 

Also, the only reason this particle exists is because arcs has a quirk where claims can only go on outputs, and I need to attach tags to these handles.

markww (approved):
LGTM

bgogul (approved):

PR #166, opened 2021-10-03
Fix (again) the sorting of declarations in souffle emitter. 

bgogul:
The sorting change introduced in #153 was lost when PR #156 was merged and conflicts were not handled correctly.

This PR doesn't have any rust tests, but tests it indirectly through the C interface tests. However, the C interface tests only check that the output is always sorted and iIt does not create an input with unordered decls and make sure they are sorted. We should add unit tests directly for souffle emitter to capture this sorting behavior.


aferr:
Sorry for breaking this :)

aferr (approved):

PR #168, opened 2021-10-04
Support dot in IDs

aferr:
Because antlr does not allow lexer rules that overlap, the PATH lexer
rule is replaced with ID which is now used in both positions. This means
that quotes can appear in the names of file paths (which is not desired), but since this code
is likely to be thrown out, this is probably OK.


bgogul:
> that quotes can appear in the names of file paths (which is not desired)

Why is undesirable? Shouldn't it be OK to say `import "some.file.auth"`?

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -60,22 +60,25 @@ assertion
     | fact ':-' flatFact (',' flatFact )* '.' #hornClauseAssertion
     ;
 
+// The IDs following "Export" are path names where JSON files containing
```
bgogul, line 63: Can you add a test?

bgogul (approved):

aferr:
> > that quotes can appear in the names of file paths (which is not desired)
> 
> Why is undesirable? Shouldn't it be OK to say `import "some.file.auth"`?

You're right I guess that's fine.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: There is already a test that checks that policy statements can be properly imported and exported and some tests about the related cryptographic signature checking.

I am guessing that I am over-indexing on the line you are commenting on, and that you want a test of the new functionality from this commit.

bgogul: I meant adding a test where access path has a dot (`.`) in it. I don't think any of the test_inputs meet the criteria?

markww (approved):

PR #169, opened 2021-10-04
Fix interface between AL/Raksha

aferr:
Our auth logic implementation appends "says_" to the beginning of
relations, but the convention in the Raksha impl has been to use camel
case. So to make this work `saysDowngrades` is renamed to
`says_downgrades` and similar for two other relations.


src/analysis/souffle/authorization_logic.dl
```diff
@@ -26,19 +26,23 @@
...
 
+// These relations need to be populated by the emitted authorization logic 
+// code. Because the emitted authorization logic code appends "says_" to the 
+// beginning of relations, these relations need to break the style 
+// conventions here.
+//
 // Some party asserts the claim that the label of an access path should be
 // downgraded by removing the referenced tag from the set.
-.decl saysDowngrades(speaker: Principal, p: AccessPath, t: Tag)
+.decl says_downgrades(speaker: Principal, p: AccessPath, t: Tag)
```
bgogul, line 36: Instead of messing with `taint.dl`, I was just thinking of just adding some rules like  `saysDowngrades(...) :- says_downgrades(...)`. We could add this in datalog_fact.h here: https://github.com/google-research/raksha/pull/161/files#diff-ac24c09d794c9c04efca399024ec64dfc514c4a37cac98e2460aaf5fd1b99d49R60

aferr: Yeah, that's probably better.

PR #170, opened 2021-10-04
Make will operate on handles

aferr:
Previously both will and may would operate on tags, but it makes more
sense for will to operate on handles. By allowing the "will" relation to
work on handles, the app behavior spec writer does not need to know
anything about the taint analysis, and can describe usgae in terms of
handles. This also seems more likely to prevent errors since the spec
writer does not have the option of forgetting to mention some tags in
the usage.

As a side-effect, some more comments are removed from the example.

### DIFF
--- a/rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
+++ b/rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
@@ -43,9 +43,6 @@
     "LocalDevice" may("do_pure_computation", "raw_video_tag").
     "LocalDevice" may("egress_to_shopping_service", "product_id_tag").
 
-    // Could also give different permissions to RemoteService for once it
-    // has "product_id_tag".
-
     // These are rules we should get rid of in the real thing.
     // The universe relations probably don't need to be decentralized. We could
     // have rules that populate a "centralized" universe and passes this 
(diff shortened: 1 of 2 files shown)


rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -75,36 +72,32 @@
...
-// (We can ostensibly do this as long as we can write
-// `claim <RelationName(...)>` in manifests)
-//
-// NOTE: maybe the "will" relations should refer to handles rather than tags, 
-// but the "may"s still refer to the tags, and we figure out which tags to
-// get permissions about based on the taint analysis (for now both still 
-// operate on the tags).
 "ImageDetector" says "ImageDetector"
-    will("do_pure_computation", "raw_video_tag").
+    will("do_pure_computation", "id_sensor_packet.camera_feed").
```
bgogul, line 76: Don't you still need to check some properties of the handle?

aferr: This is just a line that would go in the manifest (or app behavior spec) where the specifier says how the data on this handle is going to be used.

 Eventually, the check described in the comments at the bottom of this file will check the taint analysis on these handles to figure out which "may" facts it needs to check against. (Any "wills" done on handles need to come with a "may" by the owner of the taint tags on the handles, and more complicated proofs involving delegation can be used to satisfy the may part).

bgogul: By the Check, I believe you are referring to this:
```
//          `P says will <RELATION> using <HANDLE>`
//      prove
//          `ownsTag(Q, <TAG>)`
//          `Q says may <RELATION> using <TAG>`
```
How is the `<TAG>` for the `<HANDLE>` in `will` computed? I don't recall any datalog relations that connect the `<HANDLE>` with the `<TAG>`?

aferr: That's the check I am referring to, yes. The taint analysis in `taint.dl` connects taint tags to paths. We would need to apply the "will" facts to handles, then when the particle is instantiated, substitute the handle for the real path (just like we do/need to do for the claimHasTag facts).

aferr: (It might be that I'm not properly using the terms handle/paths, but hopefully this is close enough to get the idea across ?)

rust/tools/authorization-logic/raksha_examples/VisionPipelineAuthLogic
```diff
@@ -75,36 +72,32 @@
...
 "ImageSelector" says {
-    "ImageSelector" will("do_pure_computation", "raw_video_tag").
+    "ImageSelector" will("do_pure_computation", "is_detection_boxes").
     declassifies("is_select_image_id", "raw_video_tag").
     hasTag("is_select_image_id", "product_id_tag").
     objectSelected("EndUser", "is_selected_image_id").
 }
 "ProductIdOutput" says 
-    "ProductIdOutput" will("egress_to_shopping_service", "product_id_tag").
+    "ProductIdOutput" will("egress_to_shopping_service", "pio_remote_service").
```
markww, line 84: If this `will` annotation speaks of the `pio_remote_service` within the `VisionPipeline` recipe, shouldn't it have to indicate the recipe name as part of the `will` annotation? Perhaps this should be:

"ProductIdOutput" will("egress_to_shopping_service", "VisionPipeline.pio_remote_service").

or something similar.

PR #171, opened 2021-10-04
Auth logic args.len() off by 1 fix

aferr:


rust/tools/authorization-logic/src/main.rs
```diff
@@ -31,7 +31,7 @@ fn main() {
     let filename = &args[1];
     let in_dir = &args[2];
     let out_dir = &args[3];
-    let decl_skip = if(args.len() > 3) { &args[4] } else { "" };
+    let decl_skip = if(args.len() > 4) { &args[4] } else { "" };
```
bgogul, line 34: should this be 5 (assuming index of  `args` starts at `0`)?

aferr: It seems it should be 4 not 5 because it's `>` rather than `>=`. I also just checked and it works at 4, but not at 5.

bgogul: Never mind. I thought this was an assert.

bgogul (approved):

PR #176, opened 2021-10-06
A design pattern for non-unanimous decisions

aferr:
This shows a policy where a non-unanimous decision is made. It is a
contrived example rather than a particular application. It is simplified
from [PR #141](https://github.com/google-research/raksha/pull/141),
which includes a larger application. The particular choice of dispute
resolution policy is now based on whether a principal occupies a
particular role.


rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
```diff
@@ -0,0 +1,128 @@
...
+//    - the data owners each express their preference about what they'd like to 
+//    happen (as an attribute of the form "DataOwner" hasPreference())
+//    - The policy makes the decision using these preferences
+// Exactly how the policy makes the decision is orthogonal and can be 
+// application specific. In this case the policy is similar to role-based 
+// access control. The policy allows whichever principal occupies a specific 
+// role make the final decision.
+
+"DataOwner1" says {
+    owns("datum1").
```
markww, line 32: Is `owns` the same as `ownsTag`? Or is it something different?

rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
```diff
@@ -0,0 +1,128 @@
...
+    // owners will respect the outcome from the policy.
+  
+    // In this case, the preference of the device owner is the only one
+    // that matters for both pieces of data.
+    "ExampleParticle" may("send_remotely", "datum1") :-
+        PrincipalX canActAs "DeviceOwner",
+        PrincipalX hasPreference("do_send_remotely").
+    "ExampleParticle" may("send_remotely", "datum2") :-
+        PrincipalX canActAs "DeviceOwner",
+        PrincipalX hasPreference("do_send_remotely").
```
markww, line 98: I'm reading these examples as

"""
ExampleParticle may send `datum(1|2)` remotely if there exists an arbitrary principal, `PrincipalX`, that can act as `DeviceOwner` and that arbitrary principal has the `do_send_remotely` preference.
"""

Is that right? It seems like there's a disconnect in the `Principal` that the head and body of the rule are discussing. Should `ExampleParticle` be replaced with `PrincipalX` or `PrincipalX` be replaced with `ExampleParticle`?

rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
```diff
@@ -0,0 +1,128 @@
...
+}
+
+
+// This is a policy that collects the preferences of both data owners
+// on whether or not they are OK with data being sent. However, the 
+// actual decision depends only on the preference of the principal
+// in a specifc role (DeviceOwner).
+
+// This role-based decision policy is one way of resolving the conflict,
+// but the pattern of delegating
```
markww, line 76: Incomplete comment?

aferr: comment completed.

rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
aferr: I think your policy is definitely more robust for the case where there are multiple device owners (or in general multiple parties that can occupy the relevant role). 

Since this is a piece of code meant to demo a pattern, I prefer the simpler version, since it might be easier to understand. I added a comment that clarifies that here it's assumed there's just one device owner.

rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
markww: Did you commit but not push the changes here? I'm not seeing the completed comment (nor some other changes in this file).

rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
markww: Sounds good, simpler is probably better.

rust/tools/authorization-logic/raksha_examples/policy_patterns/NonUnanimousDecisions
```diff
@@ -0,0 +1,132 @@
...
+    isPrincipal("DataOwner1").
+    isPrincipal("DataOwner2").
+    isPrincipal("RoleAuthority").
+    isPreference("do_send_remotely").
+    isPreference("do_not_send_remotely").
+}
+
+// These are the checks that would be generated by a particle
+// called "ExampleParticle" with "will("send_remotely", "handleX")"
+// and also hasTag("handleX", "datum1"), hasTag("handleX", "datum2").
```
markww, line 121: Hold on, this comment says "will" but the recipe body says "may". Is this a disconnect or is this intentional?

aferr: This is intentional. This is meant to be an instance of the check described in the comment of [VisionPipelineAuthLogic](https://github.com/google-research/raksha/pull/170/files), which is repeated here:

```
// For each particle P, for each tag t in union over tags in read edges of P,
//      for each statement of the form:
//          `P says will <RELATION> using <TAG>`
//      prove
//          `ownsTag(Q, <TAG>)`
//          `Q says may <RELATION> using <HANDLE>`
```

The idea is that:
- data owners describe permitted us of data with `may`
- app specifiers describe usage that will happen on handles with `will`
- When the taint analysis says that some tags show up on a handle, we check that the intended use aligns with the permitted use (by using the "will"s to figure out some rules that check the "mays" that we generate)

Issue #164, opened 2021-10-01
Explicitly state headers and sources belong to Bazel packages rather than using globs and wildcards.


markww:
Currently, we have a number of places in the code where we use globs and wildcards to collect groups of source files that belong to that package. This was highly convenient when we had one package per directory, but is quickly becoming ugly. This issue tracks fixing this ugliness by having each package explicitly state its list of headers and sources.

PR #180, opened 2021-10-07
Eliminate globs from build targets; explicitly state srcs and hdrs.

markww:
Clean up build rules to elminate globs from build rules. Now all build
rules explicitly state the sources and headers that they use. This fixes
issue #164.


aferr (approved):

markww:
@aferr: the reason is that if you have multiple packages within a single directory, a wildcard can overmatch and pull in the source files of other packages. That produces very confusing build errors.

PR #181, opened 2021-10-07
Replace manual args with StructOpt

### LINKED ISSUE #8: [Authorization Logic] Look Into StructOpt for command line args

bMacSwigg:
Adapts the existing four command-line arguments into proper [clap::Arg](https://docs.rs/clap/2.33.1/clap/struct.Arg.html)s. Updates README documentation accordingly. This resolves #8.

In addition to the built-in parsing and typing, this also automatically transforms documentation into help text. For example:
```
$> cargo run -- -h
[...]
auth-logic 0.1.0
An auth logic compiler.

USAGE:
    auth-logic-prototype [OPTIONS] <filename> --in-dir <in-dir> --out-dir <out-dir>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -s, --skip <decl-skip>...    List of declarations to skip when generating Souffle code
    -i, --in-dir <in-dir>        Input directory where the auth logic program lives. If unspecified, it will be the
                                 current working directory [env: PWD=/home/mcswiggen/raksha/rust/tools/authorization-
                                 logic]
    -o, --out-dir <out-dir>      Output directory, where the compiled datalog program will be saved. If unspecified, it
                                 will be the current working directory [env:
                                 PWD=/home/mcswiggen/raksha/rust/tools/authorization-logic]

ARGS:
    <filename>    The name of the auth logic program to compile
```

Note that in the normal way of running the commands (with `cargo run`), you now have to append an extra `--` before the arguments, so that they get passed on to the auth-logic program rather than interpreted by `cargo`.

Now that `decl_skip` is always a `Vec<String>`, I think we could simplify the parameter types of the functions that use it. In particular, the `Option<Vec<String>>` (in `ast_to_souffle_file` and `emit_program`) doesn't seem to be used for anything, since it's always Some() when it's passed in. (And if no values are passed for `decl_skip` we get an empty Vec, which is what None is transformed into anyway.) But, for now I've opted to limit the scope of this change.

### DIFF
--- a/src/analysis/souffle/tags.dl
+++ b/src/analysis/souffle/tags.dl
@@ -28,4 +28,11 @@
 .type Tag <: symbol
 .decl isTag(tag: Tag)
 
+// An access path marked with an integrity tag (shortened here to IntegTag)
+// indicates that some property must be true of the data at the given
+// AccessPath. The conservative over-approximation for propagating an integrity
+// tag is that a node has an integrity tag only if all of its predecessors have
+// that integrity tag.
+.type IntegrityTag <: symbol
+
 #endif // SRC_ANALYSIS_SOUFFLE_TAGS_DL_
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -52,6 +52,27 @@
 // of the taint analysis).
 .decl claimNotEdge(principal: Principal, src: AccessPath, dst: AccessPath)
 
+// Indicates that an integrity tag is directly applied to an AccessPath. This
+// rule should only be populated with base facts with no derivation to prevent
+// us from getting tangled up in stratification issues. Propagation will happen
+// via the lacksIntegrityTag rule.
+.decl hasAppliedIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
+
+// Integrity tag universe.
+.decl isIntegrityTag(integTag: IntegrityTag)
+
+// Removes an integrity tag from some path.
+.decl removeIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
+
+// Indicates that a particular tag may lack an integrity tag, which
+// indicates that it should not have an integrity tag.
+.decl lacksIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
+
+// Used to test whether a particular path has a particular integrity tag.
+// Should be used as close to top-level checks as possible, and not for
+// propagation, to prevent stratification issues.
+.decl mustHaveIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
+
 //-----------------------------------------------------------------------------
 // Rules
 //-----------------------------------------------------------------------------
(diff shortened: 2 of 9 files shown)


rust/tools/authorization-logic/README.md
```diff
@@ -77,11 +77,12 @@ claim, and the recipient uses the public key to check the signature.
...
 ```
-cargo run FILENAME INPUT_DIR OUTPUT_DIR
+cargo run -- --in-dir INPUT_DIR --out-dir OUTPUT_DIR FILENAME
 ```
 
-which will result in a Souffle file called FILENAME.dl in OUTPUT_DIR. If Souffle
-is installed, it will be run on the generated code.
+which will result in a Souffle file called FILENAME.dl in OUTPUT_DIR. INPUT_DIR
+and OUTPUT_DIR are optional, and will use the current working directory if
+unspecified. If Souffle is installed, it will be run on the generated code.
```
markww, line 85: This might not be the right PR for this, but I think the interface would be cleaner if, instead of the command line arguments specifying a filename, input dir, and output dir, specifying just an input file and an output file. Feel free to defer to an issue or ignore if this is not the right spot for this.

markww: Also, should the `skip` option be documented here, or is that too internal to worry about in this file?

rust/tools/authorization-logic/src/main.rs
```diff
@@ -25,14 +25,38 @@ mod souffle;
...
-    let filename = &args[1];
-    let in_dir = &args[2];
-    let out_dir = &args[3];
-    let decl_skip = if(args.len() > 4) { &args[4] } else { "" };
-    compilation_top_level::compile(filename, in_dir, out_dir, &decl_skip);
-    souffle::souffle_interface::run_souffle(&format!("{}/{}.dl", out_dir,
-                                                    filename), out_dir);
+    let opt = Opt::from_args();
+    compilation_top_level::compile(&opt.filename, &opt.in_dir, &opt.out_dir,
+        &opt.decl_skip);
```
markww, line 59: As long as we are messing with options, it would be nice to have an option that allowed this binary to stop after compilation and just print the datalog file rather than always running Souffle afterwards. Again, feel free to ignore if this is the wrong commit for this.

rust/tools/authorization-logic/README.md
bMacSwigg: I like the idea of input file and output file, but it's a little complicated because we also use the output directory for Souffle files. Certainly solvable, but I think I'll wait to do it in another PR.

`skip` seems internal/specialized enough that it's probably not worth highlighting here. It seems simpler to have this be the minimal basic example, and rely on `cargo run -- --help` (or `-h`) for the detailed, living documentation. But with that in mind, I've also added a mention of the help command here.

rust/tools/authorization-logic/src/main.rs
bMacSwigg: Definitely agreed. That's pretty easy to add, so I included it here.

aferr:
> Updated the tests here. Not sure what I need to add to the Bazel BUILD file for structop; I'm not familiar with Bazel for Rust.

I was trying to help with this for a bit. What I tried is here: https://github.com/google-research/raksha/commit/d468a2f5af1b3cedb80cbc371dbfadfb05397a74 . I can get it to produce a new error, but can't figure out how to get past this new error with proc-macro2.

Maybe @bgogul does since he is the one that set up blaze/rust integration.

bgogul:
> > Updated the tests here. Not sure what I need to add to the Bazel BUILD file for structop; I'm not familiar with Bazel for Rust.
> 
> I was trying to help with this for a bit. What I tried is here: [d468a2f](https://github.com/google-research/raksha/commit/d468a2f5af1b3cedb80cbc371dbfadfb05397a74) . I can get it to produce a new error, but can't figure out how to get past this new error with proc-macro2.
> 
> Maybe @bgogul does since he is the one that set up blaze/rust integration.

Sorry for the trouble you are having. As far as I can see, the dependencies for structopt are not being setup for bazel. For that you will have to do the following from `rust/tools/authorization-logic` directory:
 $ cargo raze

This should update the `Cargo.lock` file and also add some files in `//third_party/cargo/...`. Usually, things would just work, but in some cases, you will have to pass along addition information to the package spec in `Cargo.toml`. Can you put the error output to a gist and I will be happy to take a look.

aferr:
> This should update the `Cargo.lock` file and also add some files in `//third_party/cargo/...`. Usually, things would just work, but in some cases, you will have to pass along addition information to the package spec in `Cargo.toml`. Can you put the error output to a gist and I will be happy to take a look.

Here's the error: https://gist.github.com/aferr/3555ef4299a57a65f3e38d7139c60574

I was trying various things in Cargo.toml by basically pattern matching on what you had. There error is about `proc_macro2`, which I see you also have something for in `Cargo.toml`. It also doesn't seem like proc_macro2 is a dependency for structopt.

bMacSwigg:
FYI, I've updated this PR with the Cargo and BUILD updates. I'm still seeing the proc-macro2 errors that Andrew was.

PR #183, opened 2021-10-08
Negated predicates

aferr:


rust/tools/authorization-logic/test_inputs/negation
```diff
@@ -0,0 +1,8 @@
+"TestPrin" says foo("x") :- !bar("x").
+"TestPrin" says baz("x") :- bar("x").
+
+// yes
+q1 = query "TestPrin" says foo("x")?
+// no
+q2 = query "TestPrin" says baz("x")?
+
```
markww, line 8: In this test, `q1` should be true because `bar` does not have `"x"` (or indeed, anything) in it, right? This test would be a bit better controlled if we also checked that it did the right thing for a symbol that was present in a relation, either as an extension to this test or a separate one.

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -28,6 +28,7 @@ pub struct AstPrincipal {
 /// implemented manually.
 #[derive(Debug, Hash, Clone, Serialize, Deserialize)]
 pub struct AstPredicate {
+    pub sign: bool, // false if negatedm, true otherwise
```
markww, line 31: This seems like a source of confusion. There are two ways to naively read this field when you're far away from the field's definition. One is "Is this value signed (ie, negative when true)" the other is "What is the polarity of this sign (ie, positive when true)". To prevent bugs, it might be worth creating an `enum` with two values: `Negated` and `NotNegated` to make the intent clear at the various usage sites.

Also, Nit: typo in comment.

rust/tools/authorization-logic/test_inputs/negation
aferr: ok fixed.

rust/tools/authorization-logic/src/ast.rs
aferr: ok fixed.

PR #185, opened 2021-10-11
A simpler vision pipeline example

aferr:
This PR simplifies the vision pipeline example by removing the local
devices group, and removing the condition that the product ID needs to
be selected by the end user.


rust/tools/authorization-logic/raksha_examples/VisionPipelineSimple
```diff
@@ -0,0 +1,146 @@
...
+        isAccessPath(pathX).
+}
+
+"ImageDetector" says "ImageDetector"
+    will("in_memory_computation", "raw_video_tag").
+"ImageSelector" says {
+    "ImageSelector" will("in_memory_computation", "raw_video_tag").
+    declassifies("is_select_image_id", "raw_video_tag").
+    hasTag("is_select_image_id", "product_id_tag").
+}
```
bgogul, line 78: Can we replace this with a particle spec? If so, let us document what that would look like as a comment here.

aferr: Maybe if we extend the syntax of particle specs we could. I added a comment.

By the way this does correspond to a manifest which is almost the same as the one we had for the un-simplified version.  The only difference is that for the simpler version there is one fewer tag. I checked in a separate .arcs file with that change as `third_party/arcs/examples/VisionPipelineSimple.arcs`

PR #186, opened 2021-10-13
Make multimic2 compile

aferr:
This commit gets multimic2 to compile. It tries to change as little else
about it as possible other than to get it to compile. (It's not tested
in any other way).

Co-authored-by:markww


aferr:
Don't feel obligated to pull this into your branch if you don't want. (Or also you can, if you do want). This is just another way I can think of to help get you up and running with auth logic.

aferr:
@markww @bgogul This now just makes new files rather than modifying Marks. This PR is the base for #187 and #197. If you don't want me to merge this in, I can also just delete these files and make the base of #187 main. Let me know what you want me to do.

bgogul:
> @markww @bgogul This now just makes new files rather than modifying Marks. This PR is the base for #187 and #197. If you don't want me to merge this in, I can also just delete these files and make the base of #187 main. Let me know what you want me to do.

Is my understanding that this just makes a copy of the `multimic` directly and #187 and #197 use them? If so, let us drop this PR and continue the discussion on the dependent PRs? Are the other PRs ready for review? Or, were you planning to make it testable with your `may`, `will`, `allowedUsage` changes?

aferr:
> Is my understanding that this just makes a copy of the `multimic` directly and #187 and #197 use them? If so, let us drop this PR and continue the discussion on the dependent PRs? Are the other PRs ready for review? Or, were you planning to make it testable with your `may`, `will`, `allowedUsage` changes?

It changes multimic to get it to compile. OK I can drop it, though. 187 should be ready for review. I can maybe just make  tests out of the queries to make it more incremental.

markww:
Yeah, I think that this can just be dropped. I'm going to close this so it's out of the PR view, but leave the backing branch intact so it can be restored if we want it.

PR #187, opened 2021-10-15
A variant of the multimic policy.

aferr:
This is an example of the multimic policy using a combination of IFC and
auth logic using the may/will check. Because in this scenario, a lot of
the information needed to make policy decisions is not available until
runtime, this example assumes that there is an interpreter and a runtime
that uses the interpreter to decide if particles can run (or not). (We
do not have an interpreter or a reference monitor, yet).


src/analysis/souffle/examples/multimic/multimic-variant.arcs
```diff
@@ -0,0 +1,69 @@
...
+particle MicCompute
+  audioIn: reads AudioData { }
+  asrNeuralNet: reads ASRNeuralNet { }
+  asrStorageConn: writes ASRData { }
+  // Need to evaluate this at runtime since information about user's
+  // policies are not known until then.
+  // MicCompute says {
+  //    "MobileDeviceRuntime" canSay isInRoom(UserX)
+  //        :- isUser(UserX). // we would need to figure out how to get rid of 
+  //                          // this isUser part
```
markww, line 39: One thought that I've had about Universes: are Universes in our policy language the same thing as types? Not user-defined types, but the basic types that our policy language understands.

One way to look at a type is that it is an overapproximation of the set of possible values that may be associated with a particular name. A Universe in the Datalog sense is just the exact set of data values that a particular family of names used in recipes can have. I don't think it's a coincidence that we have almost a one-to-one correspondence between Datalog base types and Datalog universes. Perhaps in the policy language that we are building the two concepts should be unified.

src/analysis/souffle/examples/multimic/multimic-variant.arcs
```diff
@@ -0,0 +1,69 @@
...
+// Storage credentials can be in the form of authorization logic facts
+// and guarded by the may/will check. So I didn't include credentials checking 
+// here.
+
+particle Notifier
+  audioIn: reads AudioData { }
+  networkChannelsToUserDevices: writes Notification { }
+  // (This could be more like "claims { ... } " to keep it more in Arcs style.)
+  // Notifier says {
+  //    will("deliver_privacy_notifications", "audioIn").
```
markww, line 25: I don't know that keeping Arcs style is a goal. We're using it as a starting point, but we're building something new.

One thing I do notice here is that we have a `Notifier says` block while enclosed within a `Notifier` particle. I wonder, in the future, if it would be more helpful to have an implicit speaker of the particle while enclosed in a particle's context. That said, super early for that kind of decision, just something to think about.

src/analysis/souffle/examples/multimic/multimic-variant.arcs
```diff
@@ -0,0 +1,69 @@
...
+  // identified users from "MobileDeviceRuntime" (see also comment in 
+  // multimic.authlogic).
+
+// Storage credentials can be in the form of authorization logic facts
+// and guarded by the may/will check. So I didn't include credentials checking 
+// here.
+
+particle Notifier
+  audioIn: reads AudioData { }
+  networkChannelsToUserDevices: writes Notification { }
```
markww, line 22: I was thinking, upon reviewing this line, that we should probably have an indication that this does not derive from anything, as it would likely be pre-generated, static strings. But I'm not sure if the current `derives from` syntax admits of an output that does not derive from any input. I guess that's another reason to switch to @bgogul's suggested `does not derive from` syntax.

markww (commented):
This is great. This is a very imaginative and natural-feeling way of expressing some of the concerns that I was trying to go for. I have some quibbles, but they're all in the dataflow graph side of things; the policy part looks great.

src/analysis/souffle/examples/multimic/multimic-variant.arcs
aferr: I thought derives from was more about removing edges between two handles from the graph. Isn't it enough to just not put any tags on the handle? Is the point to catch accidentally leaving out tags?

src/analysis/souffle/examples/multimic/multimic-variant.arcs
aferr: I think making the speaker implicit is probably better like you said.

I also think syntax changes for aesthetics/convenience/readability are eventually going to be important, but like you said it is early to nail down those things down while we are still figuring out how to get it to work functionally. I left it totally verbose just so that what it means and how it connects with everything else mechanically is clear.

src/analysis/souffle/examples/multimic/multimic-variant.arcs
aferr: I'm not totally sure I completely follow your point. I think the universe relation pattern we have is just our way of making sure our code is stratifiable. I do think that once we have the key system design details figured out, having types in our policy language will be useful. I remember you also had some ideas about subtypes for principals, and I think that still sounds potentially useful.

Actually, is this your point:
 - programmer-introduced types tend to correspond to the universe relations that we make
 - if we want to eventually generate and populate universe relations "behind-the-scenes" so that our developer-facing language is clearer and less verbose, then the user-defined types are probably exactly the universe relations we need to create. (And the constants that get used for each user-defined type are probably how we populate the universe relations).
 
 If that's close to your point, then I think that's on to something.

markww:
Don't merge this into the "compiles" branch yet. I had a conversation with Gogul that clarified my thinking about this case considerably.

markww:
Is this still up to date, or has it been superseded by later PRs?

aferr:
@markww leave it for now, I can just turn this and the other one into test on just the rust side. This and the other one are different scenarios.

aferr:
@bgogul I turned this into a test on just the rust / authorizaiton logic side, and changed the test infrastructure there a bit. Technically this was approved already, but I'll wait for you to have a change to give it a look before pushing it through.

aferr:
Or if Gogul can't get to it, can @markww ?

rust/tools/authorization-logic/src/souffle/souffle_interface.rs
```diff
@@ -40,6 +40,7 @@ pub fn emit_souffle(filename: &str) {
 /// translates it to datalogIR, and emits souffle code to a new file with the
 /// same name as the input but with the .dl extension in the out_dir
 pub fn input_to_souffle_file(filename: &str, in_dir: &str, out_dir: &str) {
+    println!("{}", &format!("indir: {}/{}", in_dir, filename));
```
markww, line 43: The use of `format!` here is redundant, as `println!` already accepts `format!`-style formatting for its first argument. Also, is this supposed to be here, or is this a leftover debug print>

markww (approved):
LGTM

rust/tools/authorization-logic/src/souffle/souffle_interface.rs
aferr: Thanks for catching that :) . This was not meant to be here, and it's written in that redundant way because I wanted to just lazily copy/paste it while debugging. It's gone now.

Issue #48, opened 2021-07-23
Setup Kokoro/GCB for CI Testing


bgogul:
We have github actions now, but would be good to migrate to kokoro to get more resources and better logs.

markww:
@bgogul , I think you already fixed this. I've reassigned it to you, but will let you hit the "close with comment" button so you get the associated feeling of accomplishment :)

bgogul:
> @bgogul , I think you already fixed this. I've reassigned it to you, but will let you hit the "close with comment" button so you get the associated feeling of accomplishment :)

Thank you for the honors, @markww! :)

Fixed by #212 and #213.

PR #190, opened 2021-10-21
Fix an error in Kokoro build due to using an old version of Bazel.

markww:
Makes progress towards fixing https://github.com/google-research/raksha/issues/48


markww:
This PR is dead, we have a GCB CI set up instead.

PR #192, opened 2021-10-21
Skip declaration of `says_ownsTag` when generating authorization facts.

bgogul:
Part of the fix for #78

### DIFF
--- a/src/xform_to_datalog/authorization_logic_datalog_facts.cc
+++ b/src/xform_to_datalog/authorization_logic_datalog_facts.cc
@@ -32,7 +32,7 @@ AuthorizationLogicDatalogFacts::create(
   const std::filesystem::path &program_dir, absl::string_view program) {
   auto result_dir = std::filesystem::temp_directory_path();
   int res = generate_datalog_facts_from_authorization_logic(
-    program.data(), program_dir.c_str(), result_dir.c_str(), "");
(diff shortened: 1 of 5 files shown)


src/xform_to_datalog/authorization_logic_datalog_facts.cc
```diff
@@ -32,7 +32,7 @@ AuthorizationLogicDatalogFacts::create(
   const std::filesystem::path &program_dir, absl::string_view program) {
   auto result_dir = std::filesystem::temp_directory_path();
   int res = generate_datalog_facts_from_authorization_logic(
-    program.data(), program_dir.c_str(), result_dir.c_str(), "");
+    program.data(), program_dir.c_str(), result_dir.c_str(), "says_ownsTag");
```
markww, line 35: I notice in PR #191, you treat `says_ownsTag` and `says_hasTag` specially in similar ways. Should `says_hasTag` also be addressed here?

bgogul: Good catch. Actually, I should remove `says_hasTag` from #191, because it is not clear if we need a principal component for `hasTag`. Currently, we don't have it in `taint.dl`, but it is likely that we may need it. We can address this thing separately.

Issue #195, opened 2021-10-26
`says_hasTag` is disconnected from relations considered by Raksha, making it ineffective.


markww:
The authorization logic allows speaking of whether a principal can say that something has a tag. For instance, the following is valid authorization logic:

```
"EndUser" says ownsTag("EndUser", "userSelection").
"EndUser" says "P1" canSay hasTag(accessPathX, "userSelection") :- isAccessPath(accessPathX).
```

This says that while "EndUser" maintains ownership over "userSelection", it is allowing "P1" to `claim` that access paths have the userSelection tag. Currently, however, the above auth logic results in the following datalog:

```
.decl grounded_dummy(x0: symbol)
.decl says_canSay_hasTag(x0: symbol, x1: symbol, x2: symbol, x3: symbol)
.decl says_hasTag(x0: symbol, x1: symbol, x2: symbol)
.decl says_isAccessPath(x0: symbol, x1: symbol)
says_ownsTag("EndUser", "EndUser", "userSelection").
says_hasTag("EndUser", accessPathX, "userSelection") :- says_hasTag(x___1, accessPathX, "userSelection"), says_canSay_hasTag("EndUser", x___1, accessPathX, "userSelection").
says_canSay_hasTag("EndUser", "P1", accessPathX, "userSelection") :- says_isAccessPath("EndUser", accessPathX).
grounded_dummy("dummy_var").
```

which means that `says_canSay_hasTag` is used in populating `says_hasTag`. `says_hasTag` is not connected to the analogous relation in `authorization_logic.dl` `saysHasTag`. In contrast, `ownsTag` is:

```
// Rules for linking generated relations with the ones in the dl program.
.decl says_ownsTag(speaker: Principal, owner: Principal, tag: Tag)
saysOwnsTag(x, y, z) :- says_ownsTag(x, y, z).
```

This means that only the owner can claim a tag is present on some access path and `canSay hasTag` is currently useless. The issue runs deeper than just the code that is generated, however, as this snippet in `taint.dl` (written by yours truly :( ) indicates:

```
hasTag(tgt, tag) :- claimHasTag(principal, tgt, tag), ownsTag(principal, tag).
```

there is no analogous rule for `saysHasTag` in the file.

We should fix this or decide that `canSay hasTag` is not what we want and that we only want to have owners claim the presence of tags.

PR #199, opened 2021-10-28
Delegate using canSay for asrDisallowed tag rather than ownership.

markww:
In the multimic example, we have previously used changing ownership
settings to allow UserCTainter to claim that its output has the
asrDisallowed tag because of issue #195. Because PR #198 fixes that,
this PR can now make the multimic example perform a lesser-powered
delegation by allowing UserCTainter to say that an access path has
asrDisallowed while still causing MicCompute to retain ownership over
the tag.


markww:
I think this branch has served its purpose.

PR #200, opened 2021-10-28
Change `downgrade` to `removeTag`.

bgogul:
Downgrade has a specific meaning in IFC. In our implementation, `downgrade` is mainly used to remove a `tag`. So, changing the name from `downgrade` to `removeTag` to make the intention more explicit.

### DIFF
--- a/src/analysis/souffle/tests/taint_with_logic/test_taint_with_auth_logic.dl
+++ b/src/analysis/souffle/tests/taint_with_logic/test_taint_with_auth_logic.dl
@@ -82,7 +82,7 @@ claimHasTag("ImageSelectionSource", "image_selection", "image_selection_tag").
 claimHasTag(
     "DetectionModelSource", "detection_model", "image_detection_model_tag").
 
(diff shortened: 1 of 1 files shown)


aferr:
Removing a tag actually does do a downgrade. A downgrade means moving a label down in the lattice. You can think of our labels as using the lattice over sets where subset is the lattice operator. Removing a tag makes the set smaller, so it does do a downgrade. This name is also fine though. It could also make the implementation more clear if we made labels distinct objects.

bgogul:
> Removing a tag actually does do a downgrade. A downgrade means moving a label down in the lattice. You can think of our labels as using the lattice over sets where subset is the lattice operator. Removing a tag makes the set smaller, so it does do a downgrade. This name is also fine though. It could also make the implementation more clear if we made labels distinct objects.

Yes, I agree removing a tag is a downgrade. I renamed it to be very explicit so that it is not confusing to users. Also, we might want to reuse the `downgrade` relation to capture something high-level as well.

PR #203, opened 2021-10-29
Make ir package no longer depend on Arcs proto library.

markww:
This is a small cleanup PR that makes the ir package no longer depend
upon the Arcs protos. It should be merged only after PRs #201 and #202.
This is part of the effort to fix issue #118.


bgogul (approved):
Nice! Thanks for cleaning this up.

PR #205, opened 2021-11-01
Move HandleConnectionSpec into :ir, split proto fns into ir:proto.

markww:
Prior to this PR, HandleConnectionSpec and ParticleSpec were placed in
//src/xform_to_datalog/arcs_manifest_tree. It has become more clear as
we have worked on the IR that these classes are part of the general
Raksha IR rather than being special to the xform_to_datalog tool, so
I have begun moving them to the IR package.

This PR moves the HandleConnectionSpec class to ir. As part of this
move, I have also split out the proto functions into a separate proto
library. I will handle ParticleSpec in a follow-on PR.


src/ir/handle_connection_spec.h
```diff
@@ -0,0 +1,78 @@
...
+    : name_(std::move(name)), reads_(reads), writes_(writes),
+      type_(std::move(type)) {}
+
+  const std::string &name() const { return name_; }
+
+  bool reads() const { return reads_; }
+
+  bool writes() const { return writes_; }
+
+  const types::Type &type() const { return *type_; }
```
bgogul, line 44: nit: I would remove the whitespace separating them, but leave it up to you.

src/ir/handle_connection_spec.h
```diff
@@ -0,0 +1,78 @@
...
+
+  // Get all of the AccessPaths rooted at the associated ParticleSpec
+  // (indicated through the provided name) through leaf fields of type_.
+  std::vector<AccessPath> GetAccessPaths(
+      std::string particle_spec_name) const {
+    std::vector<AccessPath> result_paths;
+    AccessPathSelectorsSet selectors_set =
```
bgogul, line 51: `const &` or we really making a copy here?

bgogul (approved):

src/ir/handle_connection_spec.h
markww: Neither; `GetAccessPathSelectorsSet` returns a temporary and thus this is a guaranteed copy elision.

src/ir/handle_connection_spec.h
markww: OK, I'll take out the whitespace here.

markww: Done.

PR #206, opened 2021-11-02
Untangle Predicate arena from Predicate decoder. 

markww:
This PR fixes a nasty little tangle between the IR and the protos. Due to
our desire to copy Predicates between spec and impl version of TagChecks,
we place them in an arena to prevent from having to use std::uniqute_ptr
(which restricts copying) while still ensuring that the predicates lived
long enough to be printed out as datalog. The initial way I did this was
by having the PredicateDecoder act as the arena. This is fine if the
only way you are constructing predicates is by parsing protos, but bad
otherwise.

This PR separates the concepts of the PredicateDecoder from the
PredicateArena. It also simplifies the ownership model for Predicates:
while the PredicateArena initially held only the root Predicate pointed
to by a TagCheck, with Predicates pointing to their sub-Predicates with
std::unique_ptrs, now all Predicates are owned by the PredicateArena and
pointers between Predicates are just regular pointers.

To ensure that Predicates are constructed only on arenas, I have made
their constructors private and given them a public static Create factory
method which takes a PredicateArena as an argument. To allow const
global Predicates to be constructed in tests (handy for test
parameterization) I have created a FakePredicateArena which just leaks
the std::unique_ptr it is handed rather than storing it. This allows
creating a global arena for tests that is effectively const due to its
lack of internal state.

This PR makes progress on fixing #118.


src/ir/predicate_test.cc
```diff
@@ -86,73 +93,118 @@ INSTANTIATE_TEST_SUITE_P(ToDatalogRuleBodyTest, ToDatalogRuleBodyTest,
...
+static const And *kAndTag1Tag3_2 =
+    And::Create(kArena, kTag1Present, kTag3Present);
+static const And *kAndTag2Tag3_2 =
+    And::Create(kArena, kTag2Present, kTag3Present);
+
+static const Or *kOrTag1Tag2 = Or::Create(kArena, kTag1Present, kTag2Present);
+static const Or *kOrTag1Tag3 = Or::Create(kArena, kTag1Present, kTag3Present);
+static const Or *kOrTag2Tag3 = Or::Create(kArena, kTag2Present, kTag3Present);
+
+static const Or *kOrTag1Tag2_2 = Or::Create(kArena, kTag1Present, kTag2Present);
```
bgogul, line 147: Why do we need this in addition to `kOrTag1Tag2`? (Same question for other variables as well.)

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
```diff
@@ -135,11 +135,11 @@ class ParticleSpec {
   // A map of HandleConnectionSpec names to HandleConnectionSpecs.
   absl::flat_hash_map<std::string, raksha::ir::HandleConnectionSpec>
     handle_connection_specs_;
-  // The PredicateDecoder object which was used to construct the predicates
+  // The PredicateArena object which was used to construct the predicates
   // on all of the checks on this ParticleSpec. This also owns those
   // predicates, so holding this field here causes the checks to live as long
   // as the ParticleSpec.
-  raksha::ir::proto::PredicateDecoder predicate_decoder_;
+  std::unique_ptr<raksha::ir::PredicateArena> predicate_arena_;
```
bgogul, line 142: It doesn't feel quite right to have an object just for the sole purpose of owning these predicates. There should be a way to avoid such work around. Once the protos and ir is detangled, we should look into cleaning up these structures.

src/ir/fake_predicate_arena.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef SRC_IR_FAKE_PREDICATE_ARENA_H_
+#define SRC_IR_FAKE_PREDICATE_ARENA_H_
+
+#include "src/ir/predicate_arena.h"
+
+namespace raksha::ir {
+
+// An "arena" that just returns the objects it should capture, releasing them
+// from the std::unique_ptr, which leaks them. Allows avoiding mutable global
```
bgogul, line 25: Even for tests, leaks seem like a bad idea. How about we make the arena part of the `Test` context and use it?

bgogul (commented):
I had a couple of concerns or questions that I left in the PR:
 - Can we avoid FakePredicateArena from leaking data?
 - Why are there duplicate predicates in tests.

src/ir/predicate_test.cc
markww: This checks that we have the right behavior for values that are structurally equal, but not pointer equal. My earlier draft of this test created a bunch of structurally-distinct objects and ensured that they only compared equal if they were pointer-equal. I realized that, because I early-exist on pointer equality when comparing predicates, I wasn't actually exercising the case where structural equality compares to true. These duplicates and the numbered equivalence classes fix that issue with the test.

src/xform_to_datalog/arcs_manifest_tree/particle_spec.h
markww: Definitely. This is very much a consequence of the lack of class-level distinction between spec and impl object versions, and would go away if we fix that issue.

src/ir/fake_predicate_arena.h
markww: You mean like a member of a test fixture? The issue here is that I'm using this arena to create global test objects to place into a list of parameters for a `TEST_P` pattern. As such, they need to be constructed before the test fixture.

In a previous draft, I tried making these test objects be lambdas that would yield the test object when provided the arena. This allowed me to parameterize the `TEST_P` with the lambdas and place the arena inside the test fixture. That was nice from a statelessness perspective, but it was a pretty deep change to the organization of the tests. I can switch back to that approach.

src/ir/predicate_arena.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace raksha::ir {
+
+class PredicateArena {
+  public:
+  virtual ~PredicateArena() {}
+
+  // CapturePredicate takes in an owning pointer for some predicate base
+  // class, takes ownership of it, and returns a pointer to that same
+  // predicate base class. This
```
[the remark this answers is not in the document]
markww: Done.

src/ir/fake_predicate_arena.h
bgogul: Yes, I meant in a text fixture. Getting rid of the fake arena would be nice, but not the end of the world. I will let you make the call on whether it is worth the change in the organization of the tests.

bgogul (approved):

src/ir/fake_predicate_arena.h
markww: I've come up with a lower-cost way of doing it: make a test-only class that contains a predicate and the single-use arena used to allocate it. This will allow easy construction of const test objects in the static lifetime while also avoiding mutable state. Going to do that.

markww: Fixed.

PR #207, opened 2021-11-02
Move ParticleSpec to IR, separate out protos, delete arcs_manifest_tree.

### LINKED ISSUE #118: Disentangle the close intertwining of the IR classes and the protos.

markww:
This PR moves the ParticleSpec class from the arcs_manifest_tree
directory/package into the ir directory/package. Because this was the
last class in arcs_manifest_tree, this allows us to delete that
subdirectory. As part of the move, we split the proto functionality out
of the ParticleSpec class and into a separate function in the ir::proto
directory. This disentangles Arcs proto code from the last IR class; as
such, this PR fixes #118.


bgogul (approved):
Nice!

PR #209, opened 2021-11-03
Implement may/will check

aferr:
This is meant to implement the may/will check that was previously
described in comments.

Another potentially useful part of this that is not implemented yet is
to throw an error when a particle reads from a handle that has at least
one tag but does not describe how this handle is used with "will". To
implement this, we would need a notion of a particle in the IR and in
the datalog, so this has been put off.


src/xform_to_datalog/testdata/ok_claim_propagates.dl
```diff
@@ -1,6 +1,7 @@
 // GENERATED FILE, DO NOT EDIT!
```
aferr, line 1: @bgogul I had to change your test data, so check that this is OK.

src/analysis/souffle/may_will.dl
```diff
@@ -0,0 +1,45 @@
+#ifndef MAY_WILL_DL_
+#define MAY_WILL_DL_
+
+#include "taint.dl"
+
+// .decl hasInputPath(particle: Principal, path: AccessPath)
```
markww, line 6: Is this supposed to be here? It looks like it's some experimental code that got commented out but never deleted.

src/analysis/souffle/may_will.dl
```diff
@@ -0,0 +1,45 @@
...
+// than occasionally disbelieve that a request is being made.
+
+// For each particle P, for each tag t in union over tags in read edges of P,
+//      for each statement of the form:
+//          `P says will <RELATION> using <HANDLE>`
+//      prove
+//          `ownsTag(Q, <TAG>)`
+//          `Q says P may <RELATION> using <TAG>`
+
+.decl permittedUsage(actor: Principal, tag: Tag, usage: symbol)
```
markww, line 24: Here, you have `tag` preceding `usage`, while in `saysMay`, you have `usage` preceding `tag`. To prevent confusion, it would be nice to have a consistent parameter order between `may` and other relations.

src/analysis/souffle/may_will.dl
```diff
@@ -0,0 +1,45 @@
...
+
+permittedUsage(consumer, tag, usage) :- ownsTag(owner, tag),
+    saysMay(owner, consumer, usage, tag).
+
+.decl disallowedUsage(dataConsumer: Principal, usage: symbol, tag: Tag)
+
+disallowedUsage(dataConsumer, usage, tag) :-
+    saysWill(dataConsumer, usage, path), mayHaveTag(path, tag),
+    !permittedUsage(dataConsumer, tag, usage),
+    isPrincipal(dataOwner).
```
markww, line 34: Why is this here? It appears that `dataOwner` is not referenced anywhere else in the rule. Was this supposed to be `dataConsumer`?

src/xform_to_datalog/datalog_facts.h
```diff
@@ -66,13 +67,17 @@ class DatalogFacts {
 .decl duplicateTestCaseNames(testAspectName: symbol)
 .output duplicateTestCaseNames
 
+.output disallowedUsage
```
markww, line 70: Perhaps this should have `(IO=stdout)` to prevent the issue we addressed in #208 ?

src/analysis/souffle/may_will.dl
```diff
@@ -0,0 +1,45 @@
...
+//      prove
+//          `ownsTag(Q, <TAG>)`
+//          `Q says P may <RELATION> using <TAG>`
+
+.decl permittedUsage(actor: Principal, tag: Tag, usage: symbol)
+
+permittedUsage(consumer, tag, usage) :- ownsTag(owner, tag),
+    saysMay(owner, consumer, usage, tag).
+
+.decl disallowedUsage(dataConsumer: Principal, usage: symbol, tag: Tag)
```
markww, line 29: We should introduce a new `Usage` (or similar) base type rather than using a bare `symbol`.

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//-------------------------------------------------------------------------------
+
+"EndUser" says {
+    ownsTag("EndUser", "userSelection").
+    "P1" canSay hasTag(PathX, "userSelection") :- isAccessPath(PathX).
+    may("P2", "doLocalComputation", "userSelection").
```
bgogul, line 21: Isn't something missing here? Don't we want to say that "P2" is allowed to do `"localComputation"` on the data `"EndUser"` owns as long as it has the tag `"userSelection"`?

aferr: It's the other way: the taint analysis tells us if data _might_ be influenced by user selection data, not that it definitely was.  The idea behind the check is to check for consent for usage whenever data may have influenced something.

Ah although, the example is probably confusing because "userSelection" is meant to be a secret, but that doesn't seem intuitively like a secret. I changed it to "rawVideo" so that the example makes more sense conceptually.

I left this open to see if it actually addresses your question/comment.

src/analysis/souffle/may_will.dl
aferr: oops. thanks.

src/analysis/souffle/may_will.dl
aferr: Can we put that off actually? The authorization logic frontend and backend both lack types for now and the usages come from auth logic. So we would need to add a bunch of extra rules to deal with this and the types don't really buy us much safety because they are connected to untyped things anyway. If we like the may/will check, we might also want to add may/will as keywords that apply to predicates rather than just using them as ordinary predicates.

src/analysis/souffle/may_will.dl
aferr: oops removed.

src/analysis/souffle/may_will.dl
aferr: fixed.

src/xform_to_datalog/datalog_facts.h
aferr: Right. fixed. Thanks again.

src/analysis/souffle/may_will.dl
markww: Can't we just add the base types in the shim layer like we do between `says_ownsTag` and `saysOwnsTag` and similar?

aferr: OK

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
```diff
@@ -0,0 +1,31 @@
...
+    may("P2", "doLocalComputation", "rawVideo").
+}
+
+// It would be better if we had control over the manifest syntax and could just 
+// extend it so we could include:
+//
+// particle P2
+//  
+"P2" says {
+    will("doLocalComputation", "R.P2#1.foo").
```
markww, line 30: Having `"R.P2#1.foo"` as the access path indicates to me that this `will` claim is attaching itself to a particular particle's access path rather than attaching to all particles that implement this `ParticleSpec`. Is that right? If the `ParticleSpec` represents a particular class of device or a particular version of a piece of software, I would expect that it `will` do the same thing in all of its instantiations.

If what is currently written is what we want, I need to re-work the numbering of particle instances. It's currently not as local to a recipe nor a particular `ParticleSpec` as one might like.

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -55,13 +56,17 @@ INSTANTIATE_TEST_SUITE_P(
 .decl duplicateTestCaseNames(testAspectName: symbol)
 .output duplicateTestCaseNames(IO=stdout)
 
+.output disallowedUsage(IO=stdout)
```
markww, line 59: Why is the `.output disallowedUsage(IO=stdout)` declaration in two locations?

src/xform_to_datalog/datalog_facts.h
```diff
@@ -66,13 +67,17 @@ class DatalogFacts {
 .decl duplicateTestCaseNames(testAspectName: symbol)
 .output duplicateTestCaseNames(IO=stdout)
 
+.output disallowedUsage(IO=stdout)
```
markww, line 70: Why is the `.output disallowedUsage(IO=stdout)` declaration in two locations?

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
aferr: I want it to work the way you want it to work. The comment on line 27 was just not finished. I fixed it now. See also the comment in the other file `simple_may_will_fail.authlogic`. The auth logic code is just what should be produced from one specific instantiation of P2 from the recipe.

In general, I think you and I are very much on the same page about how applying annotations to handles in particles should work. So if it seems like I am doing a different thing It's probably either: 1) me making a typo, or 2) me not having a way to hook it up to the way it really works just yet.

Actually, how easy is it to extend the Arcs manifest language or to find some other bit of syntax that we can repurpose for this kind of annotation? (IIRC, we can't change the actual arcs manifest language, and when I looked at the parser for Arcs it was a manually written one rather than one produced by a parser generator).

src/xform_to_datalog/datalog_facts_test.cc
aferr: You mean a few different files? Two of these files (like the one in this comment) contain a string that is used to compare against the output of generated code. So we need to change both the code generation in `datalog_facts.h` and the code used for tests (twice in `datalog_facts_test.cc`, `ok_claim_propagates.dl`). Cloudbuild is sad if we don't do this.

src/xform_to_datalog/datalog_facts.h
aferr: See other comment.

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
bgogul: Sorry for reopening this thread.  Where did we end up on this? Ideally, we should just want to be able to specify at the level of particle spec, e.g., `"P2.foo"` instead of `"R.P2#1.foo"`. Is the plan to use "R.P2#1.foo" for now and move to "P2.foo" later?

(This will also be useful for the ownership PR #215.)

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
bgogul: I probably did not word my comment well.  In this example, we are hard coding "P2" in the authorization logic. If so, doesn't that require "EndUser" to be aware of all particles in the system? Ideally, the "EndUser" will only be aware of the ingestion particle at the edge of the data flow graph. Therefore, "EndUser" will allow the ingestion particle to add tags to the data that capture the data usage restrictions. In this case, we might want to say that the tag `"userSelection"` (or `"rawVideo"`) would mean only "doLocalComputation" is allowed. As the data flows through the system, the policy engine should make sure that the tags on the data are compatible with the operations being performed. I am not sure how that would be modeled or captured.

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
aferr: I want the annotation in the particle spec to be `P2.foo` and then the backend generates `R.P2#1.foo` for this specific instance. To do this we need to change the arcs manifest syntax, though.

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
aferr: > In this example, we are hard coding "P2" in the authorization logic. If so, doesn't that require "EndUser" to be aware of all particles in the system? 

This is true in this simple test, but that's not true in general. In a real example, "EndUser" would delegate to a party it trusts to apply the tags properly, say "ThirdPartyPrivacyOrg". Then "ThirdPartyPrivacyOrg" would audit the interfaces of the actual application (in other words the ingestion particles) and apply the tags to the particles (by delegating to the particles)**. Concretely this would look like:

```
"EndUser" says "ThirdPartyPrivacyOrg" canSay hasTag(pathX, "rawVideo") :- isPath(pathX).
"ThirdPartyPrivacyOrg" says {
     "P2" canSay hasTag(pathX, "rawVideo") :- isPath(pathX).
     "P3" canSay hasTag(pathX, "rawVideo") :- isPath(pathX).
     // ..etc
}
```
As a result, the user's policy only needs to talk about the abstraction of the data and top-level organzations. The third party privacy org's policy talks about the actual particles because it audits the application. More layers of delegation could also be introduced as needed (e.g. user delegates to SomeCompnay, and SomeCompany delegates to PrivacySubOrg). 

Although this one is simple because it's a test, these two are meant to be more fully fleshed out examples with a delegation chain that, like the one described in this thread, does not require the user to know about the application: [1](https://github.com/google-research/raksha/tree/main/rust/tools/authorization-logic/raksha_examples), [2](https://github.com/google-research/raksha/pull/197).


> As the data flows through the system, the policy engine should make sure that the tags on the data are compatible with the operations being performed. I am not sure how that would be modeled or captured.

This PR is meant to implement exactly this check. Just like you said, the tags are added to the ingestion points. Then the taint analysis (in `taint.dl`) propagates these tags. The check implemented here (mainly in `may_will.dl` and `datalog_facts.h`) is meant to make sure the operations done on data tainted with the tags matches what the data owners say is allowed. The auditors of the application behavior spec write what particles will do and this populates `saysWill(...)`. The rule that defines `disallowedUsage` finds the cases where there is a mismatch, which happens when a particle `will` do something with a handle (saysWill on the RHS), that has been tainted with a tag (mayHaveTag on the RHS), and the owner of the tag does not allow it (!permittedUsage) on the RHS.

**Delegation  with canSay does already work like this -- if X delegates to Y, Y can delegate to Z. If we want add another form of delegation that prevents re-delegation, SecPal also has a design for this we can just pick up.

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
bgogul: Sounds good. So, the summary is that it is not possible to do it as is and needs some code changes.
In any case, I don't think we will need to change arcs syntax. We might be able to just add some inference rules in datalog that instantiate the particles themselves.

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
bgogul: Sounds good.  My other question is if using tags in the  `will` and `may` abstractions are sufficient for capturing all the desired operations in a particle. In any case, we can solve that later and let us go with this for the time being.

bgogul (approved):

src/analysis/souffle/examples/simple_may_will/simple_may_will_pass.authlogic
aferr: The goal here is to check prescriptive policies about data usage. 

BTW this idea is mostly made up. It's based on the check done in [This policy language paper](https://www.microsoft.com/en-us/research/publication/s4p-a-generic-language-for-specifying-privacy-preferences-and-policies/) (that you sent out), except that in our implementation, the check involves a static (taint) analysis, whereas in the paper, this may/will check happens when a client talks to a remote service.

markww (approved):

PR #211, opened 2021-11-03
Mention the included files  explicitly in `souffle_cc_library`.

bgogul:
If the files are not specified, the builds don't work when executed remotely.


markww (approved):
I'm surprised it worked without this. LGTM.

bgogul:
> I'm surprised it worked without this. LGTM.

I *think* by default `#include` looks for files in the same directory. It seems to work for local builds as all files are in WORKSPACE, but for remote builds that assumption probably does not hold.

PR #212, opened 2021-11-04
Configuration for google cloud build to run tests.

bgogul:
All the necessary configuration is under `gcb/` directory. I went back and forth between kokoro and gcb, but decided to stick with gcb in the end for a variety of reasons:
  - The setup is working now (modulo providing view permissions for others contributors).
  - Integrating ResultsUI and RBE on Kokoro requires further setup and seems more involved.
  - With GCB all the necessary configuration is in the open source code itself. 
 
One of the advantages of Kokoro is that it posts nice status messages with an embedded link to the test ResultsUI. However, I figured out a way to do the same in GCB. It just requires wrapping the bazel runs in a script, extracting the sponge links from the output and posting as status. I will do that in a follow up URL.


gcb/rbe/configs/path/cc/tools/cpp/empty.cc
```diff
@@ -0,0 +1 @@
+int main() {}
```
markww, line 1: No newline at the end of file.

gcb/rbe/configs/path/cc/cc_wrapper.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# Ship the environment to the C++ action
+#
+set -eu
+
+# Set-up the environment
+
+
+# Call the C++ compiler
+/usr/bin/clang "$@"
```
markww, line 25: Is this a sample file? Should there be an environment setup?

.dockerignore
```diff
@@ -0,0 +1,6 @@
+.vscode
+.git
+bazel-out
+bazel-arcs
+bazel-bin
+
```
markww, line 6: Extraneous blank line. Ignore if this is just a copy of a sample file.

gcb/rbe/configs/path/cc/BUILD
```diff
@@ -0,0 +1,151 @@
...
+    "-fstack-protector",
+    "-Wall",
+    "-Wthread-safety",
+    "-Wself-assign",
+    "-fcolor-diagnostics",
+    "-fno-omit-frame-pointer"],
+    opt_compile_flags = ["-g0",
+    "-O2",
+    "-D_FORTIFY_SOURCE=1",
+    "-DNDEBUG",
```
markww, line 113: Should we be turning assertions off? Usually I like to have them on at runtime, but if this is a file we got from elsewhere, this is fine.

markww (commented):
Found some basic style stuff, but nothing serious. Then again, this is a big pile of stuff and I'm not sure what parts I should look at in depth and what I should skim. It looks to me like much of this are copies of standard files; is there a particular directory pattern I should skim rather than look at in depth?

bgogul:
> Found some basic style stuff, but nothing serious. Then again, this is a big pile of stuff and I'm not sure what parts I should look at in depth and what I should skim. It looks to me like much of this are copies of standard files; is there a particular directory pattern I should skim rather than look at in depth?

Everything under the `gcb/rbe` is generated. Feel free to skip them. I will probably avoid incorporating any of the style changes there. It might be worthwhile looking at the other files in some detail. Hopefully, there are no surprises there.

gcb/presubmit.sh
```diff
@@ -0,0 +1,24 @@
...
+# Run all the tests.
+${BAZEL_TEST} //src/...
+# Verifies that the following targets build fine:
+#  - Arcs parser and proto works.
+#  - Arcs manifest tests that we do not yet handle parse correctly.
+#  - Authorization logic code with bazel.
+#  - Arcs manifest examples.
+${BAZEL_BUILD} //third_party/arcs/examples:consume third_party/arcs/proto:manifest_cc_proto \
+   //src/analysis/souffle/tests/arcs_manifest_tests_todo/... \
+   //rust/tools/authorization-logic:authorization_logic \
```
markww, line 21: Now that we are actually using the authorization logic in tests, this should probably be removed.

Actually, should we swap the order of the build and test invocations? In a sense, the build checks are "more basic" in that they just make sure that everything still builds without invoking any deep Raksha functionality. Therefore, we would expect these build commands to pass more frequently than the test commands, making them more cacheable.

markww (approved):

gcb/rbe/configs/path/cc/tools/cpp/empty.cc
bgogul: Auto generated. Will leave this as is.

.dockerignore
bgogul: Done.

gcb/rbe/configs/path/cc/BUILD
bgogul: Auto generated. I will leave this as is for now.

gcb/rbe/configs/path/cc/cc_wrapper.sh
bgogul: Auto generated.

gcb/presubmit.sh
bgogul: done.

PR #213, opened 2021-11-05
Extract the results url from cloud build runs and post it as status

bgogul:
This should make it easier to access the results of the bazel test runs. Specifically, two different status messages are posted:
  - Required builds
  - Required tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN apt-get update && apt-get install -y \
     bison \
     build-essential\
     clang \
+    curl \
     doxygen \
     flex \
     g++ \
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -28,12 +28,21 @@ steps:
 
 # Run presubmit tests.
 - id: 'presubmit'
+  secretEnv: ['GITHUB_COMMIT_STATUS_TOKEN']
   waitFor: ['build-image']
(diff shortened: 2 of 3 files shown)


markww (approved):
This looks good to me. Also, kind of fun to see the proposed results right there in the CI check results, makes it easy to take a look!

markww:
A thought: since we can see right here that the cloud builds are working, want to get rid of the Github Actions stuff in the same PR?

bgogul:
> A thought: since we can see right here that the cloud builds are working, want to get rid of the Github Actions stuff in the same PR?

I considered it, but I think it is prudent to leave the GitHub Actions in place until we have enough confidence in the new set up. We could remove the "required" constraint on that test run though.

markww:
Fair point on the confidence thing. IDK, if we're going to keep it around, I think it might as well be blocking. With a time difference of 2m vs 25m, we'll be tempted to just ignore it if it's not required anyway. Keeping it required encourages us to keep considering while it is still around.

bgogul:
> Fair point on the confidence thing. IDK, if we're going to keep it around, I think it might as well be blocking. With a time difference of 2m vs 25m, we'll be tempted to just ignore it if it's not required anyway. Keeping it required encourages us to keep considering it while it is still around.

Makes sense. Let us leave it in place for now. We can remove it in two weeks or so if we don't run into any major issues with the new set up.

PR #214, opened 2021-11-06
Replace claimHasTag relation with saysHasTag. 

bgogul:
These are equivalent relations. `saysHasTag` is more aligned with authorization logic claims and we should just use that going forward.

### DIFF
--- a/src/analysis/souffle/authorization_logic.dl
+++ b/src/analysis/souffle/authorization_logic.dl
@@ -46,9 +46,7 @@
 
 .decl saysOwnsTag(speaker: Principal, owner: Principal, tag: Tag)
 
-// claimHasTag and saysHasTag are synonyms. They indicate that the `speaker`
-// claims that `path` definitely has `tag` taint.
-.decl claimHasTag(speaker: Principal, path: AccessPath, tag: Tag)
+// The `speaker` says that `path` definitely has `tag` taint.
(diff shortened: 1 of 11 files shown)


aferr:
I think you should now be able to get rid of `claimHasTag` in [authorization_logic.dl](https://github.com/google-research/raksha/blob/main/src/analysis/souffle/authorization_logic.dl). You might also want to look around for other uses of `claimHasTag` in rules in `../analyis/souffle/*` to see if you can simplify them away now as well.

aferr (changes requested):
Can you remove `claimHasTag` from `authorizaiton_logic.dl` now? Maybe double check for other uses in `analysis/souffle` as well? 

(sorry for the duplicate comment. this was just bad use of github on my part).

markww (approved):
LGTM

bgogul:
> Can you remove `claimHasTag` from `authorizaiton_logic.dl` now? Maybe double check for other uses in `analysis/souffle` as well?
> 
> (sorry for the duplicate comment. this was just bad use of github on my part).

@aferr I don't see any references to claimHasTag in `authorization_logic.dl`. I searched all the code base too. Am I missing something?

aferr (approved):
@bgogul Actually you did get all the uses of claimHasTag (including the one in `authorization_logic.dl`). I must have hallucinated that it was missed.

PR #217, opened 2021-11-09
Delete and simplify

aferr:
This removes some files and/or relations we don't actually use anywhere.

The vision pipeline example is src/analysis/souffle/examples is also
deleted because there is a more complete one under the authorization
logic subdirectory.


markww (approved):
LGTM.

bgogul (approved):

PR #218, opened 2021-11-12
Rename `says` relations in datalog to match facts from auth logic.

### LINKED ISSUE #102: Resolve Difference Between Types in Taint Analysis and Auth Logic to allow integration

bgogul:
Closes #102. Ultimately, we should change authorization logic to generate camel case relations, but let us defer this to the time when we rewrite the parser.

### DIFF
--- a/src/analysis/souffle/authorization_logic.dl
+++ b/src/analysis/souffle/authorization_logic.dl
@@ -28,11 +28,11 @@
 
 // Some party asserts the claim that the label of an access path should
 // remove the referenced tag from the set.
-.decl saysRemoveTag(speaker: Principal, path: AccessPath, tag: Tag)
+.decl says_removeTag(speaker: Principal, path: AccessPath, tag: Tag)
 
 // The `speaker` says that `delegatee` is allowed to indicate that a `path` does
 // not have a `tag`.
-.decl saysCanSayRemoveTag(
+.decl says_canSay_removeTag(
   speaker: Principal, delegatee: Principal, path: AccessPath, tag: Tag)
 
 // Indicates that a speaker is allowed to say removeTag.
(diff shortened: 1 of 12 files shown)


src/analysis/souffle/authorization_logic.dl
```diff
@@ -78,29 +78,29 @@ canSayRemoveTag(speaker, path, tag) :-
 // If the owner says a speaker can say removeTag, the speaker can say
 // removeTag.
 canSayRemoveTag(speaker, path, tag) :-
```
aferr, line 80: I think we shouldn't need to have `saysCanSayHasTag` or `says_canSay_hasTag` in the raksha code. I think we only need the ones that are prefixed with `says` (but don't involve `canSay`). The rules involving canSay duplicate the logic that should already be working in the generated authorization logic code. It should be enough to just have a raksha-auth-logic interface file with:

```
.decl saysHasTag(speaker: Principal, path: AccessPath, tag: Tag)
saysHasTag(s, p t) :- says_hasTag(s, p, t).

.decl saysRemoveTag(speaker: Principal, path: AccessPath, tag: Tag)
saysRemoveTag(s, d, p, t) :- says_removeTag(s, p, t).
```

and so on for the ones beginning with says. To be complete, I think we need these rules for:
```
says_removeTag
says_ownsTag
says_hasTag
```

and I think we can delete the rules about 

```
says_canSay_removeTag
says_canSay_hasTag
```

aferr: (And if this doesn't work you can push it back to me and I can try to debug it)

bgogul: These are being removed in the ownership PR #215. I am not sure if we can remove these relations before that without some tests failing.

markww: I think @aferr is right. I added the `says_canSay_removeTag` and `says_canSay_hasTag` because I didn't understand the echoing pattern for delegation. You might have to remove some rules that use them, but everything should work without them. That said, I'm not sure this is the right PR for this; this PR just renames the existing rules and makes a simple, understandable transformation on the code. That would be complicated by mixing in the more complex removal of the `says_canSay` relations from our checked-in dl files.

markww (approved):
This renaming LGTM.

aferr (approved):

PR #219, opened 2021-11-12
Remove GitHub CI workflow as GCB setup working well so far.

bgogul:


markww (approved):
Awesome work cutting us over to GCB, LGTM!

PR #221, opened 2021-11-17
Create a SystemSpec class in the place of ParticleSpecRegistry.

bgogul:
This is the place where all information about the system being modeled will be kept.This is class that frontends should build into.Further, all the information in the manifest.proto should be decoded to this class.

### DIFF
--- a/src/ir/particle_spec_test.cc
+++ b/src/ir/particle_spec_test.cc
@@ -37,39 +37,34 @@ struct ParticleSpecProtoAndExpectedInfo {
 class ParticleSpecFromProtoTest :
    public testing::TestWithParam<ParticleSpecProtoAndExpectedInfo> {
  protected:
-  ParticleSpecFromProtoTest()
-    : particle_spec_registry_(),
-      particle_spec_(CreateParticleSpec(particle_spec_registry_)) {}
+  ParticleSpecFromProtoTest() : particle_spec_(CreateParticleSpec()) {}
 
-  static const ParticleSpec &CreateParticleSpec(
-      ParticleSpecRegistry &particle_spec_registry) {
+  static std::unique_ptr<ParticleSpec> CreateParticleSpec() {
     arcs::ParticleSpecProto particle_spec_proto;
     std::string textproto = GetParam().textproto;
     CHECK(google::protobuf::TextFormat::ParseFromString(
         textproto, &particle_spec_proto))
         << "Particle spec textproto did not parse correctly.";
-    return *proto::Decode(
-        particle_spec_registry, std::make_unique<PredicateArena>(),
-            particle_spec_proto);
+    return proto::Decode(std::make_unique<PredicateArena>(),
+                         particle_spec_proto);
   }
 
-  ParticleSpecRegistry particle_spec_registry_;
-  const ParticleSpec &particle_spec_;
+  std::unique_ptr<ParticleSpec> particle_spec_;
 };
 
 TEST_P(ParticleSpecFromProtoTest, ParticleSpecFromProtoTest) {
   const ParticleSpecProtoAndExpectedInfo &param = GetParam();
 
-  EXPECT_EQ(particle_spec_.name(), param.expected_name);
+  EXPECT_EQ(particle_spec_->name(), param.expected_name);
   EXPECT_THAT(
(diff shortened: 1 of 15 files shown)


src/ir/particle_spec_test.cc
```diff
@@ -37,39 +37,34 @@ struct ParticleSpecProtoAndExpectedInfo {
...
  protected:
-  ParticleSpecFromProtoTest()
-    : particle_spec_registry_(),
-      particle_spec_(CreateParticleSpec(particle_spec_registry_)) {}
+  ParticleSpecFromProtoTest() : particle_spec_(CreateParticleSpec()) {}
 
-  static const ParticleSpec &CreateParticleSpec(
-      ParticleSpecRegistry &particle_spec_registry) {
+  static std::unique_ptr<ParticleSpec> CreateParticleSpec() {
     arcs::ParticleSpecProto particle_spec_proto;
```
markww, line 43: It looks like you've undone the pattern that I had where constructing a `ParticleSpec` required passing in the context object that would hold it (formerly `ParticleSpecRegistry`, now `SystemSpec`) so that its `unique_ptr` could not be held by any object other than the context object. I wrote this pattern to preserve the guarantee that it would definitely be held by an object with a long enough lifetime, but I do recognize that it was a little awkward to work with in tests. Is that why you've refactored it to make the `Create` method just return the owning pointer?

src/xform_to_datalog/manifest_datalog_facts.cc
```diff
@@ -85,8 +75,8 @@ ManifestDatalogFacts ManifestDatalogFacts::CreateFromManifestProto(
       // Find the ParticleSpec referenced by this Particle. The information
       // contained in the spec will be needed for all facts produced within a
       // Particle.
-      const ir::ParticleSpec &particle_spec =
-          particle_spec_registry.GetParticleSpec(particle_spec_name);
+      const ir::ParticleSpec *particle_spec =
+          CHECK_NOTNULL(system_spec->GetParticleSpec(particle_spec_name));
```
markww, line 79: If we know this to be non-null, we should use a `const ir::ParticleSpec &` for `particle_spec` to let the type system in on that information.

src/xform_to_datalog/manifest_datalog_facts.h
```diff
@@ -24,16 +24,16 @@
...
 #include "third_party/arcs/proto/manifest.pb.h"
 
 namespace raksha::xform_to_datalog {
 
 class ManifestDatalogFacts {
  public:
   static ManifestDatalogFacts CreateFromManifestProto(
-      const arcs::ManifestProto &manifest_proto,
-      ir::ParticleSpecRegistry &particle_spec_registry);
+      const ir::SystemSpec* system_spec,
```
markww, line 35: Why are we using a `const ir::SystemSpec *` instead of a `const ir::SystemSpec &`? The latter would ensure that we cannot be passed a `nullptr`.

src/xform_to_datalog/manifest_datalog_facts.cc
```diff
@@ -33,24 +33,14 @@ namespace types = raksha::ir::types;
 //  traversal. This works, but it is hard to read and hard to test. We should
 //  break this up.
 ManifestDatalogFacts ManifestDatalogFacts::CreateFromManifestProto(
-    const arcs::ManifestProto &manifest_proto,
-    ir::ParticleSpecRegistry &particle_spec_registry) {
+    const ir::SystemSpec* system_spec,
+    const arcs::ManifestProto &manifest_proto) {
```
markww, line 37: See other comments elsewhere, but I think we should use a `const ir::SystemSpec &` to ensure `system_spec` is non-null.

src/xform_to_datalog/manifest_datalog_facts.h
bgogul: Good point. Will do.

bgogul:
@markww addressed your comments. PTAL.

src/ir/particle_spec_test.cc
markww: OK, sounds good, that sounds fine to me.

markww (approved):

PR #223, opened 2021-11-17
Early start at refactoring AccessPath and AccessPathSelectorSet.

markww:
Rather than having an AccessPath be a root combined with a list of
selectors and an AccessPathSelectorSet be a straightforward set of
those AccessPaths, this PR reimagines an AccessPathSelectorSet to be
a tree of selectors with both parent->child pointers and child->parent
pointers. Now, an AccessPath can be viewed as a pointer to a particular
leaf within a tree, and can be printed out by recursing from the leaf to
the root and printing on return. This means that AccessPaths are either
a single pointer or a pair of an AccessPathRoot and a single pointer.

I didn't figure out exactly where these AccessPathSelectorSets should
live (probably somewhere long lived so that we don't have use after
frees when following the pointers in AccessPaths) nor many other
details. But I think this will make copies of AccessPaths less painful
and will make AccessPaths take up a lot less space.

### DIFF
--- a/src/ir/access_path_selector_tree.h
+++ b/src/ir/access_path_selector_tree.h
@@ -1,0 +1,62 @@
+#ifndef SRC_IR_ACCESS_PATH_SELECTOR_TREE_H_
+#define SRC_IR_ACCESS_PATH_SELECTOR_TREE_H_
+
+#include <memory>
+
+#include "src/common/logging/logging.h"
+#include "src/ir/selector.h"
+
+namespace raksha::ir {
+
+// A tree of AccessPathSelectors. Used to describe the AccessPaths that
+// derive from a particular type. This contains both links to children and
+// links to parents to support different use cases: the upwards links so that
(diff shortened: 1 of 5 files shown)


src/ir/access_path_selector_tree.h
```diff
@@ -0,0 +1,62 @@
...
+
+  class AccessPathSelectorTreeIter {
+   private:
+    std::vector<const AccessPathSelectorTree *> path_stack_;
+  };
+
+ private:
+  Selector selector_;
+  const AccessPathSelectorTree *parent_;
+  std::vector<std::unique_ptr<AccessPathSelectorTree>> children_;
```
bgogul, line 57: why vector? I was thinking more in the lines of the following:
```
AccessPath ::= AccessPathRoot SelectorTree
SelectorTree ::= Selector SelectorTree
Selector :== Field | \epsilon
```

Any reason we don't want that?

markww:
This draft has accomplished its purpose. Closing.

Issue #220, opened 2021-11-17
Ownership of access paths based on particle spec or handles


bgogul:
PR #215 introduces a notion of ownership of access paths. However, they refer to access paths in an actual instance of a particle in a recipe. Therefore, we end up having to make statements like `ownsAccessPath("UserA", "R.P1#2.foo")`, where `P1#2` is an instance of particle `P1` in recipe. We should make it possible to specify ownership at the level of access path for store handles or access path of a particle spec. e.g., `ownsAccessPath("UserA", "P1.foo")`, ​`ownsAccessPath("UserA", "store1")`.

markww:
This relates to the Raksha Arcs manifest frontend, which is now deprecated.

PR #225, opened 2021-11-18
Typing interface

aferr:
This fixes the typing interface between authorization logic and the
raksha taint analysis and IR. 

This will conflict with
[#215](https://github.com/google-research/raksha/pull/215), so I will
likely need to redo this later. Even after 215, I think this fix is
still needed in order to handle cases like the following, where in the
authorization logic there is a user-defined relation on the RHS that is used to
prove one of the facts beginning with "says" in Raksha:

```
"P1" saysHasTag("some.path", "someTag") :-
    someUserDefinedRelation("some.path", "someTag").
```

which produce the declaration:

```
.decl speaker_someUserDefinedRelation(x1: symbol, x2: symbol,
    x3: symbol)
```

which will cause a type error without an interface fix like this one.

### DIFF
--- a/src/analysis/souffle/auth_logic_type_interface.dl
+++ b/src/analysis/souffle/auth_logic_type_interface.dl
@@ -1,0 +1,49 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+
+// This file provides an interface that applies types to the untyped relations 
+// from authorization logic.
+
+
+#ifndef SRC_ANALYSIS_SOUFFLE_AUTHORIZATION_LOGIC_TYPE_INTERFACE_DL_
+#define SRC_ANALYSIS_SOUFFLE_AUTHORIZATION_LOGIC_TYPE_INTERFACE_DL_
+
+#include "dataflow_graph.dl"
+#include "authorization_logic.dl"
+#include "tags.dl"
+
+.decl says_removeTag(speaker: symbol, path: symbol, tag: symbol)
+.decl says_ownsTag(speaker: symbol, path: symbol, tag: symbol)
+.decl says_hasTag(speaker: symbol, path: symbol, tag: symbol)
+.decl says_may(speaker: symbol, actor: symbol, usage: symbol, tag: symbol)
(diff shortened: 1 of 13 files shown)


aferr:
@bgogul @markww I started fixing this because I was trying to finally get the may-will check versions of multimic connected. This will conflict with the ownership PR, so don't approve it yet. I think a version of this will still be needed after that PR (for the reason described above), so maybe it's still useful to give this early feedback, and I can make this work after this PR.

aferr:
The main thing to look at is `src/analysis/souffle/auth_logic_type_interface.dl` and to notice that the delegation examples that mark had from before still pass.

src/analysis/souffle/auth_logic_type_interface.dl
```diff
@@ -0,0 +1,49 @@
...
+
+
+#ifndef SRC_ANALYSIS_SOUFFLE_AUTHORIZATION_LOGIC_TYPE_INTERFACE_DL_
+#define SRC_ANALYSIS_SOUFFLE_AUTHORIZATION_LOGIC_TYPE_INTERFACE_DL_
+
+#include "dataflow_graph.dl"
+#include "authorization_logic.dl"
+#include "tags.dl"
+
+.decl says_removeTag(speaker: symbol, path: symbol, tag: symbol)
```
bgogul, line 28: Thanks! I was planning to create something like this after merging in the ownership PR.

Can you wait for that PR to be merged in?

src/analysis/souffle/auth_logic_type_interface.dl
```diff
@@ -0,0 +1,49 @@
...
+.decl says_may(speaker: symbol, actor: symbol, usage: symbol, tag: symbol)
+.decl says_will(speaker: symbol, usage: symbol, path: symbol)
+
+saysRemoveTag(as(speaker, Principal), as(path, AccessPath), as(tag, Tag)) :-
+    says_removeTag(speaker, path, tag).
+
+saysOwnsTag(as(speaker, Principal), as(owner, Principal), as(tag, Tag)) :-
+    says_ownsTag(speaker, owner, tag).
+
+saysHasTag(as(speaker, Principal), as(path, AccessPath), as(tag, Tag)) :-
```
bgogul, line 40: Do we need to add these rules if we are have the `.decl` above and use `says_` versions in the taint.dl? (See #218). 

Also, adding this can have some undesirable consequences as I found out when working on #220. Specifically, the rules from manifest and rules from authorization logic influence each other. Therefore, we have to make sure all grounded facts use the `says_` version. e.g., for a `claim x is foo`, we will still need to generate `says_hasTag` and not `saysHasTag`. Otherwise, some inferences will not be triggered.

src/xform_to_datalog/testdata/ok_claim_propagates.dl
```diff
@@ -2,6 +2,7 @@
 
 #include "taint.dl"
 #include "may_will.dl"
+#include "auth_logic_type_interface.dl"
```
markww, line 5: There are a number of `BUILD` files in the project where we list included `dl` files for certain tests. Even if the build system is managing to locate this file without it being explicitly named there, we should probably add it in those places so that the build system correctly re-triggers builds if this file were to change. I believe you will want to update the `BUILD` file under `arcs_fact_tests`, maybe others.

PR #228, opened 2021-11-19
Print datalog without instantiation of access paths. 

bgogul:
In this PR, instead of creating instantiated versions of the various data structures, we simply fix the access path at the time of datalog generation. This required passing along the `instantiation_map` as a part of the `DatalogPrintContext`. This is a very temporary hack. When all the classes are fleshed out, we will implement `ToDatalog` as a visitor, where the necessary context can be cleanly passed along in the visitor.

### DIFF
--- a/src/ir/access_path_root.cc
+++ b/src/ir/access_path_root.cc
@@ -1,0 +1,34 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+#include "src/ir/access_path_root.h"
+
+#include "src/ir/datalog_print_context.h"
+
+// The classes in this file describe the root of an AccessPath. At the moment,
+// we have only two types of roots: a HandleConnectionSpecAccessPathRoot and
+// a HandleConnectionAccessPathRoot, describing a HandleConnectionSpec in a
+// ParticleSpec and a Handle in a Particle, respectively.
+namespace raksha::ir {
+
+std::string AccessPathRoot::ToDatalog(const DatalogPrintContext &ctxt) const {
+  const auto *instantiation_map = ctxt.instantiation_map();
+  if (instantiation_map == nullptr) return this->ToString();
+  auto find_res = instantiation_map->find(*this);
+  return (find_res != instantiation_map->end()) ? find_res->second.ToString()
+                                                : this->ToString();
+}
+
+}  // namespace raksha::ir
--- a/src/ir/datalog_print_context.h
+++ b/src/ir/datalog_print_context.h
@@ -17,7 +17,9 @@
 #ifndef SRC_IR_DATALOG_PRINT_CONTEXT_H_
 #define SRC_IR_DATALOG_PRINT_CONTEXT_H_
 
+#include "absl/container/flat_hash_map.h"
 #include "absl/strings/str_cat.h"
+#include "src/ir/access_path_root.h"
 
 namespace raksha::ir {
 
@@ -26,7 +28,7 @@ namespace raksha::ir {
 // unique labels.
 class DatalogPrintContext {
  public:
-  DatalogPrintContext() : check_counter_(0) {}
+  DatalogPrintContext() : check_counter_(0), instantiation_map_(nullptr) {}
 
   // DatalogPrintContext is not copyable, as we need a single copy to be the
   // source of truth on creating unique labels. It is, however, movable.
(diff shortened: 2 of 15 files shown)


src/ir/datalog_print_context.h
```diff
@@ -37,8 +39,22 @@ class DatalogPrintContext {
   std::string GetUniqueCheckLabel() {
     return absl::StrCat("check_num_", check_counter_++);
   }
+
+  void set_instantiation_map(
```
markww, line 43: According to style guide, functions should be in CapsCamelCase.

src/ir/datalog_print_context.h
```diff
@@ -26,7 +28,7 @@ namespace raksha::ir {
 // unique labels.
 class DatalogPrintContext {
  public:
-  DatalogPrintContext() : check_counter_(0) {}
+  DatalogPrintContext() : check_counter_(0), instantiation_map_(nullptr) {}
```
markww, line 31: Rather than setting the instantiation map to nullptr initially then resetting it for each particle instance, it might be nicer to wrap the `check_counter_` in its own simple `CheckCounter` class. A pointer to `CheckCounter` can then be passed in in the ctor alongside the instantiation map. Then, a new `DatalogPrintContext` can be constructed for each new `Particle`, preventing `instantiation_map_` from ever being `nullptr` and removing one piece of mutable state.

That said, I hear your point about this being a temporary fix until we have a full-fledged visitor, so I'm not strongly opposed to the pattern you have here right now.

src/ir/predicate.h
```diff
@@ -111,16 +113,17 @@ class Implies : public Predicate {
...
   virtual ~Implies() { }
 
-  std::string ToDatalogRuleBody(const AccessPath &ap) const override {
+  std::string ToDatalogRuleBody(
+      const AccessPath &ap, const DatalogPrintContext &ctxt) const override {
     // To make it so that Souffle has first class information that our
     // antecedent is the same in both places it is used, we assign it to a
     // variable within the rule.
-    std::string antecedent_str = antecedent_->ToDatalogRuleBody(ap);
+    std::string antecedent_str = antecedent_->ToDatalogRuleBody(ap, ctxt);
```
markww, line 121: Ugh, the comment above this line is wrong and it is my fault. I initially wanted to assign the `antecedent_str` to some kind of temporary within the rule, but that isn't possible (or I couldn't figure it out). Could you remove the above comment? I can fix it later if you do not want to clean up my mess, but since you have altered this file already, thought I'd ask.

src/ir/access_path_root.cc
```diff
@@ -0,0 +1,34 @@
...
+
+// The classes in this file describe the root of an AccessPath. At the moment,
+// we have only two types of roots: a HandleConnectionSpecAccessPathRoot and
+// a HandleConnectionAccessPathRoot, describing a HandleConnectionSpec in a
+// ParticleSpec and a Handle in a Particle, respectively.
+namespace raksha::ir {
+
+std::string AccessPathRoot::ToDatalog(const DatalogPrintContext &ctxt) const {
+  const auto *instantiation_map = ctxt.instantiation_map();
+  if (instantiation_map == nullptr) return this->ToString();
```
markww, line 28: Isn't this a case where we would like to assert via a `CHECK`? It seems most likely to be the case that if the `instantiation_map` is `nullptr` while `ToDatalog` is called that we intended to set it but forgot.

src/ir/access_path_root.cc
```diff
@@ -0,0 +1,34 @@
...
+  auto find_res = instantiation_map->find(*this);
+  return (find_res != instantiation_map->end()) ? find_res->second.ToString()
+                                                : this->ToString();
```
markww, line 31: Also, this seems somewhat overly permissive. Within a `Particle`, we currently expect to find a match for every `AccessPath` present. This will not always be the case (for instance, stores will not be affected by the instantiation map) , but we don't currently handle the cases that would be exceptions. For now, I think it might be more appropriate to `CHECK` that the result is not `end()`.

src/ir/datalog_print_context.h
```diff
@@ -37,8 +39,22 @@ class DatalogPrintContext {
...
   }
+
+  void set_instantiation_map(
+      const absl::flat_hash_map<ir::AccessPathRoot, ir::AccessPathRoot>
+          *instantiation_map) {
+    instantiation_map_ = instantiation_map;
+  }
+
+  const absl::flat_hash_map<ir::AccessPathRoot, ir::AccessPathRoot>
+      *instantiation_map() const {
```
markww, line 50: If you follow my (optional) suggestion above to instantiate the `instantiation_map_` in the constructor with a non-null pointer, then you can dereference the pointer upon return of this function and return a `const flat_hash_map<...> &`, removing the possibility of nullness from the return type.

Also, please fix casing of method name.

src/ir/datalog_print_context.h
bgogul: If you are worried about `instantiation_map_` from being null, we could fix it by making `instantiation_map_` mandatory and the `check_counter` an optional argument. We can reinitialize the counter between particles. We will have to do something like that with `CheckCounter` as well.

In any case, let us leave this as is for now. When we have a visitor and the core data structures, this will become cleaner.

src/ir/access_path_root.cc
bgogul: Ideally, yes. However, there are tests where we cannot create a meaningful instantiation map as they already create the instantiated paths. Shoehorning a map in those tests will make them harder to read.

src/ir/access_path_root.cc
bgogul: See comment above. In the case you are referring to, the problem will be caught here:
https://github.com/google-research/raksha/blob/a24d1baeac7f6c107f6d7321d9e373369c1527fa/src/ir/access_path_root.h#L45

src/ir/predicate.h
bgogul: Done in https://github.com/google-research/raksha/pull/230. (I generally try to keep PRs single purpose. It makes it easier to cherry-pick, roll back, etc.)

src/ir/datalog_print_context.h
bgogul: Unfortunately, it is not possible to guarantee this is never null due to the way some tests are set up. :(

bgogul (commented):
I have replied to your comments. Unfortunately, it is not possible to incorporate some of your suggestions. PTAL.

src/ir/datalog_print_context.h
markww: So it is!

src/ir/predicate.h
markww: Thanks.

markww (approved):
LGTM

PR #229, opened 2021-11-19
Add auth logic tests to CI

aferr:


Dockerfile
```diff
@@ -30,3 +30,26 @@ ENV JAVA_HOME /usr/lib/jvm/default-java
...
+## Set the right version of rust
+ARG RUST_VERSION=nightly-2021-06-30
+RUN rustup toolchain install ${RUST_VERSION} \
+    && rustup default ${RUST_VERSION}
+
+# Souffle is needed to run the tests in authorization logic
+RUN curl -s https://packagecloud.io/install/repositories/souffle-lang/souffle/script.deb.sh | bash
+RUN apt-get update && apt-get install -y souffle
+
+## TODO / WIP delete these. Used for local debugging with docker
```
aferr, line 53: Note to self

Dockerfile
```diff
@@ -30,3 +30,22 @@ ENV JAVA_HOME /usr/lib/jvm/default-java
 # Set up baselisk.
 RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.10.1/bazelisk-linux-amd64 \
     -O /usr/local/bin/bazelisk && chmod a+x /usr/local/bin/bazelisk
+
+# Rust is needed for authorization logic
+ARG rustup_dir=/usr/local/cargo
+ENV RUSTUP_HOME ${rustup_dir}
+ENV CARGO_HOME ${rustup_dir}
+ENV PATH "${rustup_dir}/bin:${PATH}"
```
markww, line 38: In general, it is best practice to add items to the end of the `PATH` rather than the beginning. The reason for this is that the OS searches for binaries in `PATH` order and usually places more common locations at the beginning of the `PATH`. Only rust tools will be found in `${rustup_dir}/bin`, so it will slow down lookup of most tools to place it first.

Dockerfile
```diff
@@ -30,3 +30,22 @@ ENV JAVA_HOME /usr/lib/jvm/default-java
...
+RUN curl --location https://sh.rustup.rs > /tmp/rustup \
+  && sh /tmp/rustup -y --default-toolchain=none \
+  && chmod a+rwx ${rustup_dir} \
+  && rustup --version
+
+## Set the right version of rust
+ARG RUST_VERSION=nightly-2021-06-30
```
markww, line 45: This is a very old nightly version; would something more recent work?

Dockerfile
```diff
@@ -30,3 +30,22 @@ ENV JAVA_HOME /usr/lib/jvm/default-java
...
+RUN rustup toolchain install ${RUST_VERSION} \
+    && rustup default ${RUST_VERSION}
+
+# Souffle is needed to run the tests in authorization logic
+RUN curl -s https://packagecloud.io/install/repositories/souffle-lang/souffle/script.deb.sh | bash
+RUN apt-get update && apt-get install -y souffle
```
markww, line 51: This doesn't look right. Our project has a copy of Souffle that it already builds and uses as part of the Raksha tests; shouldn't we use the same Souffle version here rather than fetching another version from `apt`?

gcb/presubmit.sh
```diff
@@ -68,3 +68,21 @@ bazel_run build \
...
+    if cargo test
+    then
+        post_commit_status "${STATUS_MESSAGE}" "success" ${INVOCATION_ID}
+    else
+        post_commit_status "${STATUS_MESSAGE}" "failure" ${INVOCATION_ID}
+    fi
+}
+
+# Run cargo test for authorization logic
+auth_logic_cargo_test
```
markww, line 88: I don't see the utility in placing the auth logic portion of this script in a function and invoking it with no arguments. In other places in this script, stages are placed in functions but take some parameters so that they can be reused in different ways; this can only be used in one way. I think there are two ways you can go about fixing this:

1. Decide that this function can be generalized in a useful way and do so, passing it some arguments. For instance, this can become `cargo_test`, take a parameter that gets assigned to a variable like `CARGO_TEST_DIR` and pass `AUTH_LOGIC_DIR` as a n argument to that parameter.

2. Decide that this function is not worth generalizing and inline it into the script.

gcb/presubmit.sh
```diff
@@ -68,3 +68,21 @@ bazel_run build \
 
 # Run all the tests.
 bazel_run test //src/...
+
```
markww, line 71: You might want to fix the comment above this line, as with the addition of the invocation for auth logic it does not run all of the tests.

gcb/presubmit.sh
bgogul: The whole logic around posting commit status was used to extract that results URL of a bazel run. Given that we don't use bazel here, this should simply be `cargo test`. Otherwise, the post_commit_status will post an incorrect URL. 

In any case, it might be better to run this as another stage in [cloudbuild.yaml](https://github.com/google-research/raksha/blob/main/cloudbuild.yaml). Add these after the `presubmit` stage in cloudbuild.yaml and remove the code you added in presubmit.sh:
```
# Run authorization logic tests 
- id: 'auth_logic_tests'
  waitFor: ['build-image']
  name: '${_DOCKER_IMAGE}'
  entrypoint: 'bash'
  args: ['cargo', 'test']
```

These tests will run in parallel. Further, the output of the tests will aloso be nicely separated out in the google cloud build console.

Dockerfile
aferr: Isn't that version of souffle a part of bazel/BUILD, which I think doesn't put it into the docker environment? I need to run these tests with cargo which does not use bazel. I also ran this without adding souffle here first, and it did not work, so I think it must be neeeded.

bgogul: I think it is OK to install another version of souffle for now. In the current setup now, it is not straightforward to run the tool we built from the command line as it would be in the bazel's cache. We will need to add a `sh_binary` rule to wrap the souffle binary and change how it is invoked from rust here:
https://github.com/google-research/raksha/blob/60fd88a02232b5d2199e0a8d9fa31634137977a5/rust/tools/authorization-logic/src/souffle/souffle_interface.rs#L67

In any case, I agree that we might run into consistency issues. e.g., some feature not working on downloaded version but working on our local version (or vice versa). Hopefully that should not cause major problems. Let us file an issue to add a sh_binary rule for souffle and fix this later.

Dockerfile
aferr: done.

Dockerfile
aferr: I just confirmed that the latest nightly (November 1) really does not work because some cargo feature breaks ("raw" ?). Gogul and I did a short search for the latest-ish nightly that works when he was getting rust to build with blaze, and this is what worked then :)

gcb/presubmit.sh
aferr: Done! :)

gcb/presubmit.sh
aferr: Done

bgogul (commented):
Thanks @aferr  for taking care of this. The way you are posting commit status will not work for `cargo test`. I have left an alternative fix.

aferr:
> Thanks @aferr for taking care of this. The way you are posting commit status will not work for `cargo test`. I have left an alternative fix.

Where is the alternative fix? I don't see a relevant comment of yours. Are you pushing something to this branch / making another PR for this fix or am I just missing it?

Dockerfile
aferr: Ah, OK I guess it is fixable by changing how souffle is invoked from within the rust code. Sounds good.

aferr:
> > Thanks @aferr for taking care of this. The way you are posting commit status will not work for `cargo test`. I have left an alternative fix.
> 
> Where is the alternative fix? I don't see a relevant comment of yours. Are you pushing something to this branch / making another PR for this fix or am I just missing it?

@bgogul sorry found it. I think I posted in and resolved that thread in parallel with your reply.

gcb/presubmit.sh
aferr: Errr the fix for what mark said is done, but not what gogul said.

cloudbuild.yaml
```diff
@@ -37,6 +37,17 @@ steps:
     - 'REPO_NAME=${REPO_NAME}'
     - 'COMMIT_SHA=${COMMIT_SHA}'
 
+# Run authorization logic tests 
+- id: 'auth_logic_tests'
+  waitFor: ['build-image']
+  name: '${_DOCKER_IMAGE}'
+  entrypoint: 'bash'
+  args: [
+    'cd rust/tools/authorization-logic &&',
```
bgogul, line 46: It is failing with the following error: 

`bash: cd rust/tools/authorization-logic &&: No such file or directory`

You might want to use manifest-path argument: https://doc.rust-lang.org/cargo/commands/cargo-test.html. Change args as follows:
```
args: ['cargo', 'test', '--manifest-path=rust/tools/authorization-logic']
```

Another alternative is to create a `gcb/cargo_test.sh` that changes to the directory and runs `cargo test`. This will also help us add some better status messages later if needed. If you go down the route, then this cloudbuild.yaml rule would change as follows:

```
- id: 'auth_logic_tests'
  waitFor: ['build-image']
  name: '${_DOCKER_IMAGE}'
  entrypoint: 'bash'
  args: [ '.gcb/cargo-test.sh']
  ]

Comment (aferr) [reply cloudbuild.yaml]: Thanks. Sorry I had to step away since it was evening in my time zone.

I ran into [this issue](https://stackoverflow.com/questions/61055324/docker-cannot-execute-binary-file), so it looks like it also needed `-l -c`. At least when I launch the gcb job manually from the CLI, it works. So it should be OK now.

File: cloudbuild.yaml [hunk 8]
```diff
@@ -37,6 +37,16 @@ steps:
     - 'REPO_NAME=${REPO_NAME}'
     - 'COMMIT_SHA=${COMMIT_SHA}'
 
+# Run authorization logic tests 
+- id: 'auth_logic_tests'
+  waitFor: ['build-image']
+  name: '${_DOCKER_IMAGE}'
+  entrypoint: 'bash'
+  args: [
+    '-l', '-c', 'cargo', 'test', '--manifest-path=/rust/tools/authorization-logic/'
```
bgogul, line 46: Add a comment as to why `-l` and `-c` are needed?

cloudbuild.yaml
```diff
@@ -37,6 +37,16 @@ steps:
     - 'REPO_NAME=${REPO_NAME}'
     - 'COMMIT_SHA=${COMMIT_SHA}'
 
+# Run authorization logic tests 
+- id: 'auth_logic_tests'
```
bgogul, line 41: The tests don't seem to be running. This is what I see in the output:
```
Already have image: us-central1-docker.pkg.dev/google.com/raksha-ci/gcb-docker-images/gcb:latest
Rust's package manager

USAGE:
    cargo [+toolchain] [OPTIONS] [SUBCOMMAND]

OPTIONS:
    -V, --version                  Print version info and exit
        --list                     List installed commands
        --explain <CODE>           Run `rustc --explain CODE`
    -v, --verbose                  Use verbose output (-vv very verbose/build.rs output)
    -q, --quiet                    No output printed to stdout
        --color <WHEN>             Coloring: auto, always, never
        --frozen                   Require Cargo.lock and cache are up to date
        --locked                   Require Cargo.lock is up to date
        --offline                  Run without accessing the network
        --config <KEY=VALUE>...    Override a configuration value (unstable)
    -Z <FLAG>...                   Unstable (nightly-only) flags to Cargo, see 'cargo -Z help' for details
    -h, --help                     Prints help information

Some common cargo commands are (see all commands with --list):
    build, b    Compile the current package
    check, c    Analyze the current package and report errors, but don't build object files
    clean       Remove the target directory
    doc         Build this package's and its dependencies' documentation
    new         Create a new cargo package
    init        Create a new cargo package in an existing directory
    run, r      Run a binary or example of the local package
    test, t     Run the tests
    bench       Run the benchmarks
    update      Update dependencies listed in Cargo.lock
    search      Search registry for crates
    publish     Package and upload this package to the registry
    install     Install a Rust binary. Default location is $HOME/.cargo/bin
    uninstall   Uninstall a Rust binary

See 'cargo help <command>' for more information on a specific command.
```

bgogul (commented):
@aferr, the tests are not running?

cloudbuild.yaml
aferr: Not sure how I missed this... I couldn't figure out how to get it to work using a step that calls cargo test directly, so I made it run a script. It is running the tests now.

cloudbuild.yaml
aferr: It's a script that doesn't use these options now.

aferr:
@bgogul @markww This is just a gentle re-review reminder coming back from break. Step 4 output now says `test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.82s` among other things.

gcb/auth_logic.sh
```diff
@@ -0,0 +1,3 @@
+AUTH_LOGIC_DIR="rust/tools/authorization-logic"
```
bgogul, line 1: Add the following to the beginning? Add at least `set -e` so that the script exits on error.
```
#!/bin/bash
#
set -e
```

bgogul (approved):
Thanks! The tests takes around 7m. Most of that is related to downloading and compiling crates. Is it possible do this setup as part of creating the Docker image?

markww (approved):

aferr:
@bgogul I'll look into prefetching the rust crates in another PR.

PR #232, opened 2021-11-19
Remove BulkInstantiation and Instantiate methods.

bgogul:
This part of the clean up of getting rid of the concept of instantiation.


bgogul:
@markww are you OK with this direction as you have the most context on this code? I also plan to send  a PR removing the `IsInstantiated` methods.

markww (approved):

markww:
I'm ok with this. I tried to do something similar (although less cleanly) with https://github.com/google-research/raksha/pull/155. I'm very happy to see copy-based instantiation going away.

PR #236, opened 2021-11-22
Add datalog rules for integrity tags.

markww:
This PR adds rules for adding, removing, and propagating integrity tags.
This is a little more complex than regular tags, as integrity tags are
only propagated to an AccessPath if all edges pointing to that
AccessPath come from AccessPaths having that integrity tag. This
intersection property cannot be directly expressed in Datalog. Instead,
we propagate the lack of an integrity tag, which we call an inverse
integrity tag, or an InvIntegTag. Propagating the InvIntegTag allows us
to use union operations (ie, a destination access path has an
InvIntegTag if any of its incoming edges have an InvIntegTag) which is
easy to express in Datalog.

Note that this PR covers only the propagation of integrity tags that do not
have principals attached. A follow up PR will attach these facts to
authorization logic.


src/analysis/souffle/tags.dl
```diff
@@ -28,4 +28,21 @@
 .type Tag <: symbol
 .decl isTag(tag: Tag)
 
+// An access path marked with an integrity tag (shortened here to IntegTag)
+// indicates that some property must be true of the data at the given
+// AccessPath. The conservative over-approximation for propagating an integrity
+// tag is that a node has an integrity tag only if all of its predecessors have
+// that integrity tag. Checking this property requires an intersection, which
+// unfortunately cannot be directly expressed in datalog.
+.type IntegTag <: symbol
```
bgogul, line 37: Let us call it `IntegrityTag`?

src/analysis/souffle/tags.dl
```diff
@@ -28,4 +28,21 @@
...
+
+// To deal with that, we instead propagate the opposite of the integrity tag:
+// the absence of an integrity tag, which we call an inverse integrity tag, or
+// an InvIntegTag. An inverse integrity tag is introduced any node with no
+// predecessor edges and without an integrity tag added. It is also introduced
+// at any point where an integrity tag is removed. It is removed at any point
+// where an integrity tag is added. Textually, an InvIntegTag symbol is just the
+// associated IntegTag with a '!' prepended.
+.type InvIntegTag <: symbol
```
bgogul, line 46: Why do we need this? Can't we capture this in a relation like `lacksIntegrity(ap: AccessPath, owner: Principal, tag: IntegrityTag)`?

src/analysis/souffle/taint.dl
```diff
@@ -52,6 +52,30 @@
 // of the taint analysis).
 .decl claimNotEdge(principal: Principal, src: AccessPath, dst: AccessPath)
 
+// Indicates that an AccessPath is given an integrity tag directly. This rule
+// should only be populated with base facts with no derivation to prevent us
+// from getting tangled up in stratification issues. Propagation will happen
+// via the mayHaveInvIntegTag rule.
+.decl directHasIntegTag(accessPath: AccessPath, integTag: IntegTag)
```
bgogul, line 59: what about `ownership`?

src/analysis/souffle/taint.dl
```diff
@@ -71,6 +95,37 @@ mayHaveTag(tgt, owner, tag) :- hasTag(tgt, owner, tag).
...
+// The inverse integrity tag is the same text as the integrity tag except with
+// a '!' in front of it.
+integTagToInv(integTag, cat("!", integTag)) :- isIntegTag(integTag).
+
+// Introduce an inverse integrity tag at any access point that has no incoming
+// edges and is not directly given the corresponding integrity tag.
+mayHaveInvIntegTag(ap, invIntegTag) :-
+  isAccessPath(ap), integTagToInv(integTag, invIntegTag), !edge(_, ap),
+  !directHasIntegTag(ap, integTag).
+
```
bgogul, line 111: Might be simpler to do it as follows?

```suggestion
.decl lacksIntegrity(ap: AccessPath, owner: Principal, tag: Tag)

lacksIntegrity(ap, owner, tag) :-
  isAccessPath(ap), isPrincipal(owner), edge(_, ap),
  lacksIntegrity(ap, owner, tag).
```

src/analysis/souffle/tags.dl
markww: That is also fine, and semantically equivalent to what I've done with the inverse integrity tags here. The reason I made the inverse integrity tags was to hopefully prevent two different types of confusion:

1. Improve readability by making the relations not contain negatives (that is, speaking about the positive presence of an inverse integrity tag rather than the absence of an integrity tag)

2. Use a separate base type to prevent propagating an integrity tag from a place where it was supposed to be absent to a place where it was supposed to be present.

That said, if this is just confusing in a different way, happy to just use `lacksIntegrity`.

src/analysis/souffle/taint.dl
markww: I didn't handle it in this PR because it wasn't necessary to show simple propagation behavior and I wanted to get some feedback on this approach first. I can add that in if you'd like, however.

bgogul: We should think about it as confidentiality labels have ownership. Also, unlike confidentiality tags, adding an integrity tag is a privileged operation.

src/analysis/souffle/taint.dl
bgogul: We may also need to scope out the `tag` in the second rule:

```
lacksIntegrity(ap, owner, tag) :- 
	 isAccessPath(ap), isPrincipal(owner), isIntegrityTag(tag), !directHasInteg(ap, owner, tag).
```

markww: I don't think the above will work. The second rule:

```Datalog
lacksIntegrity(ap, owner, tag) :- isAccessPath(ap), isPrincipal(owner), !directHasInteg(ap, owner, tag).
```

will make every `AccessPath` that is not directly given an `IntegrityTag` lack that `IntegrityTag`. We want `AccessPath`s not directly given an `IntegrityTag` but for which all inputs have an `IntegrityTag` to have the `IntegrityTag`. This is why my current rules only introduce inverse integrity tags on `AccessPath`s not directly given an `IntegrityTag` that have no incoming edges: there is no way for them to get an `IntegrityTag`, and thus the integrity tag must be lacking. From there, I can propagate the lack of integrity tags like confidentiality tags.

src/analysis/souffle/taint.dl
```diff
@@ -71,6 +95,37 @@ mayHaveTag(tgt, owner, tag) :- hasTag(tgt, owner, tag).
...
+// Introduce inverse integrity tag wherever a corresponding integrity tag is
+// removed.
+mayHaveInvIntegTag(ap, invIntegTag) :-
+  isAccessPath(ap), integTagToInv(integTag, invIntegTag),
+  removeIntegTag(ap, integTag).
```
aferr, line 116: I find it a bit more intuitive to handle removal of integrity tags in the rule for `mustHaveIntegTag` directly rather than using the removal to introduce an inverse integ tag (or to populate the lackOfIntegTag relation).

src/analysis/souffle/taint.dl
aferr: /\ I think as I was typing that comment just the first message was in the thread, so what I wrote might be out of sync.

bgogul: From another comment thread from @markww :
> We want AccessPaths not directly given an IntegrityTag but for which all inputs have an IntegrityTag to have the IntegrityTag

Ah, now that you call this out, I am not sure if this is always desirable. Consider something like `TimestampMillis` and a particle with two inputs (say 'a' and 'b') and one output (say 'a + b'). Suppose that the inputs have `TimestampMillis` doesn't mean that the output should have the tag. That said, if we we think of integrity as provenance (say "screentext"), it would make sense to propagate integrity to all the derived outputs. I think we will want something like `edgePreservesIntegrity("screentext")` and use that in propagation.

In any case, as to your comment about `lacksIntegrity` not working, you are both right. We will have to combine them into one. Here is what I had written up before seeing @aferr's comment: 
```
lacksIntegrity(ap, owner, tag) :-
  isAccessPath(ap), isPrincipal(owner), isIntegrityTag(tag), edge(src, ap),
  lacksIntegrity(src, owner, tag), !directHasInteg(ap, owner, tag).

```

markww: Another way we could approach this is by making `IntegrityTag`s *never* propagate automatically. That is, for each `IntegrityTag`, you must specify explicitly the rules that cause, for some set of input nodes, an integrity tag to appear on an output node.

Removing automatic propagation would remove the ambient double-negation that occurs due to propagation and would allow us to only consider `IntegrityTag`s in the affirmative. If we do that, all of the stratification requirements evaporate and we can easily handle the SQL cases we are considering elsewhere.

bgogul: I am supportive of this idea, but we will still need to deal with intersection. Suppose we have a rule like `edgePreservesIntegrity(src, tgt, tag)`. We will still need to make sure that _all_ `edgePreservesIntegrity(src, tgt, tag)` to be true before we set that the integrity of `tgt` to be true.

markww: This may not be the right approach, but my idea was that we wouldn't even have a fact like `edgePreservesIntegrity`. Instead, we would always use policy rules that may match on certain subsections of the Raksha graph to propagate integrity tags.

I think we could, at least in most cases, sidestep intersection due to specific knowledge about the shape of the graph encoded within these propagation rules. Consider, for instance, the example where we remove the lower bits of a timestamp to prevent fingerprinting a user. We could write such a rule like this:

```Datalog
hasIntegrityTag(tgt, "TimestampSeconds") :-
  isOperation(tgt, "/"), // Indicates that this access path holds the result of a divide
  isDividendOf(src1, tgt), // Indicates that this access path holds the value being divided at tgt
  isShiftWidthOf(src2, tgt), // Indicates that this access path holds the value being divided by at tgt
  hasIntegrityTag(src1, "TimestampMillis"), // Is this a millisecond value.
  hasIntegrityTag(src2, "DivisibleBy1000"), // Do we know this definitely has three 0s in the low bits
  count : { edge (_, tgt) } = 2. // Only two edges enter this node.
```

The above rule effectively performs an intersection for the specific case of a division. It manages to do this without negation upon dynamic facts because it knows that a division should only take two inputs. We check that this is the case with the `count` operation; this is currently fine because we don't dynamically add edges, so the contents of the `edge` relation is static (we could even assert this property during IR translation to keep it out of Datalog entirely).

A generic propagation rule must use negation because it has to be generic and thus not know the number of inputs that it must intersect. This requires it to state the intersection as no input having the negation of the property. However, if we know that there are `X` inputs, we can exhaustively say `x1, x2, x3, ...` `X` times.

Of course, this does not work for circumstances where the user does not have specific knowledge of the number of inputs going into an `AccessPath` that is relevant to an integrity property. But perhaps those circumstances are rare. All Arcs `Particle`s know the number of input handles that they have. Most SQL expressions and statements also know the exact number of incoming inputs (the list in `IN` being a notable exception). We need a constraint to make things stratify, and perhaps this is a least-onerous one to use.

src/analysis/souffle/tags.dl
```diff
@@ -28,4 +28,11 @@
 .type Tag <: symbol
 .decl isTag(tag: Tag)
 
+// An access path marked with an integrity tag (shortened here to IntegTag)
+// indicates that some property must be true of the data at the given
+// AccessPath. The conservative over-approximation for propagating an integrity
+// tag is that a node has an integrity tag only if all of its predecessors have
+// that integrity tag.
+.type IntegrityTag <: symbol
```
bgogul, line 36: I am not sure if there is extra benefit with keeping a separate type. We should simply use the `Tag` type?

markww: The reason why I think it makes sense to have two types is because it is entirely possible that someone, while hacking on the Datalog or code that generates Datalog, could accidentally write a rule body that propagates an integrity tag based upon a fact involving a confidentiality tag or vice-versa. This separate type helps prevent that sort of error from occuring.

src/analysis/souffle/tags.dl
markww: Fixed.

src/analysis/souffle/tags.dl
markww: I've moved to using `lacksIntegrity` as suggested.

src/analysis/souffle/taint.dl
markww: Added ownership aspect.

src/analysis/souffle/taint.dl
markww: I've removed the `InvIntegTag` entirely. However, we have to make `removeIntegrityTag` introduce the lack of an `IntegrityTag` rather than directly influence `mustHaveIntegrityTag` to preserve stratification (as `mustHaveIntegrityTag` uses `!lacksIntegrityTag on the RHS of one of its rules).

src/analysis/souffle/taint.dl
```diff
@@ -71,6 +95,37 @@ mayHaveTag(tgt, owner, tag) :- hasTag(tgt, owner, tag).
...
+
+// Propagate inverse integrity tags along edges to any destination that did not
+// have an integrity tag directly added.
+mayHaveInvIntegTag(dst, invIntegTag) :-
+  isAccessPath(src), isAccessPath(dst), edge(src, dst), mayHaveInvIntegTag(src, invIntegTag),
+  integTagToInv(integTag, invIntegTag), !directHasIntegTag(dst, integTag).
+
+// An access path must have an integrity tag if it may not have the
+// corresponding inverse integrity tag.
+mustHaveIntegTag(ap, integTag) :-
```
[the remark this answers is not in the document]
markww: They do, just indirectly. `directHasIntegrityTag` blocks the introduction and propagation of `lacksIntegrityTag`-tracked facts. As such, any `AccessPath` with `directHasIntegrityTag` must also be `mustHaveIntegTag` - unless the same `AccessPath` is also the subject of a `removeIntegrityTag`, in which case the removal "wins".

src/analysis/souffle/taint.dl
markww: @bgogul and I had a video call in which he convinced me that automatic propagation is needed in at least some circumstances. We sort of stumbled upon an idea together of having a special binary `phi` operation which actually performs the merging of confidentiality state and integrity state for two incoming `AccessPath`s. A `phi` result has a confidentiality tag when at least one of its inputs has that confidentiality tag, and has an integrity tag when both of its inputs have that integrity tag. `phi`s can be chained together to capture multiple inputs.

What does this do for us? Well, it means that we can perform intersection without negation: checking that an integrity tag is present on all inputs of a `phi` is equivalent to ensuring that it is present on both of its operands.

When combined with a constraint that each `AccessPath` is the target of at most one incoming edge, this would force all state merging to occur in operations (`phi` may soon be followed by less exotic operations such as `+`, `/`, etc). That would be a radically different approach than what we currently have in this PR, and would necessitate changes elsewhere in the system, so it's probably not in scope for this particular PR.

src/analysis/souffle/tags.dl
aferr: Isn't the propagation of tags internal to the Raksha implementation? I don't see how that mistake would happen. The programmer writes rules that add or remove tags through a predicate that mentions whether the tag is added as confidentiality or integrity. So which predicate is used makes it clear.

src/analysis/souffle/taint.dl
```diff
@@ -71,6 +92,32 @@ mayHaveTag(tgt, owner, tag) :- hasTag(tgt, owner, tag).
...
+
+// Propagate lack of integrity tag along edges to any destination that did not
+// have an integrity tag directly added.
+lacksIntegrityTag(dst, owner, integTag) :-
+  isAccessPath(src), isAccessPath(dst),
+  edge(src, dst), lacksIntegrityTag(src, owner, integTag),
+  !directHasIntegrityTag(dst, owner, integTag).
+
+// An access path must have an integrity tag if it does not lack the integrity tag.
+mustHaveIntegrityTag(ap, owner, integTag) :-
```
aferr, line 117: This still has the problem that `directHasIntegrity` does not imply `mustHaveIntegrityTag`.

aferr: Sorry, I missed your comment in the other thread because I only looked at the `Files Changed` tab. Looks like this does work.

src/analysis/souffle/taint.dl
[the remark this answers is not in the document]
aferr: OK looks like it works as of latest version.

src/analysis/souffle/tags.dl
bgogul: I see your point about unintentional errors,. However, I think we should be able to treat a tag as either a confidentiality or an integrity tag. Introducing a new type keeps the namespaces of confidentiality and integrity separate, which I think is not quite desirable.

That said, I guess we could start with a separate type and change it later if needed. It might be harder (but not impossible) to go the other way. So, OK to leave it for now while we understand the problem. However, I think we will eventually want to unify them.

src/analysis/souffle/taint.dl
```diff
@@ -71,6 +92,32 @@ mayHaveTag(tgt, owner, tag) :- hasTag(tgt, owner, tag).
...
+
+// Any access point that has no incoming edges and is not directly given an
+// integrity tag lacks that integrity tag, as there is no way for it to
+// propagate to that AccessPath.
+lacksIntegrityTag(ap, owner, integTag) :-
+  isAccessPath(ap), isPrincipal(owner), isIntegrityTag(integTag),
+  !edge(_, ap), !directHasIntegrityTag(ap, owner, integTag).
+
+// Any point where an integrity tag is removed lacks an integrity tag.
+lacksIntegrityTag(ap, owner, integTag) :- isAccessPath(ap), removeIntegrityTag(ap, owner, integTag).
```
bgogul, line 107: Do we need `isAccessPath(ap)` because `removeIntegrityTag` already scopes it out?

src/analysis/souffle/tests/arcs_fact_tests/integrity_simple_test.dl
```diff
@@ -0,0 +1,178 @@
+#include "taint.dl"
```
bgogul, line 1: Could we split this up into multiple tests? Or, is it too much work?

bgogul (approved):

src/analysis/souffle/taint.dl
markww: PR #243 contains a prototype of the `phi`-based solution.

src/analysis/souffle/tests/arcs_fact_tests/integrity_simple_test.dl
markww: This can be split up, and probably should be. It's gotten rather large.

src/analysis/souffle/taint.dl
```diff
@@ -52,6 +52,27 @@
 // of the taint analysis).
 .decl claimNotEdge(principal: Principal, src: AccessPath, dst: AccessPath)
 
+// Indicates that an AccessPath is given an integrity tag directly. This rule
+// should only be populated with base facts with no derivation to prevent us
+// from getting tangled up in stratification issues. Propagation will happen
+// via the lacksIntegrityTag rule.
+.decl directHasIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
```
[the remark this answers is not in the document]
markww: My main issue with that is, to someone naive to the codebase, what is the difference between `hasIntegrityTag` and `mustHaveIntegrityTag`?

With confidentiality tags, there is a major difference between `hasTag` and `mayHaveTag`: the level of certainty is higher in the first than in the second, which is enough to clue a new reader into the difference in semantics. For integrity tags, `hasIntegrityTag` and `mustHaveIntegrityTag` are at the same level of certainty, and could invite confusion. I'm open to renaming this, I also think `directHasIntegrityTag` is awkward, but I haven't come up with a better human-readable way to indicate the distinction yet.

src/analysis/souffle/taint.dl
markww: Fixed.

src/analysis/souffle/tests/arcs_fact_tests/integrity_simple_test.dl
markww: Done.

src/analysis/souffle/taint.dl
markww: With the realization of the `phi`-based solution in a different PR, @bgogul convincing me that automatic propagation is necessary, and me switching us to a `lacksIntegrityTag`-based propagation approach, I think everything in this thread has been addressed.

src/analysis/souffle/tags.dl
markww: Sounds good, will leave it for now, and will remove distinction if it causes unwanted friction.

src/analysis/souffle/tags.dl
markww: Leaving distinction for now, as mentioned elsewhere, and will remove if we get unwanted friction.

aferr (approved):

src/analysis/souffle/taint.dl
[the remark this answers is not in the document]
markww: I could have sworn I left a comment here yesterday about this.

My big problem with `hasIntegrityTag` is that the difference between `hasIntegrityTag` and `mustHaveIntegrityTag` is not clear to a human coming into the repo and trying to make sense of it. Contrast this with `hasTag` and `mayHaveTag` where there is an obvious difference in certainty.

In addition, `directHasIntegrityTag` is different from `mustHaveIntegrityTag` because `directHasIntegrityTag` is the base case, where `IntegrityTag`s are directly added, while `mustHaveIntegrityTag` includes derived facts as well.

I also am not a fan of `directHasIntegrityTag`, but I would like a name that indicates that this is just the `IntegrityTag`s that are initially added. Perhaps `baseCaseHasIntegrityTag` or `initialHasIntegrityTag`?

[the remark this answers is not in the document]
markww: On the aspect of `says_hasIntegrityTag`, I think it might be a bit premature to do that. I don't think there is functionality in the Authorization Logic parser to connect to, and I'd prefer to add that bit only when that is there (so in a follow-up PR).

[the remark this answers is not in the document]
bgogul: I see your point about the name. Ideally, there should be only one rule `hasIntegrityTag` or `mustHaveIntegrityTag`, but I see that can cause issues with stratification. Other choices for the name: `hasAppliedIntegrityTag`, `hasAssignedIntegrityTag`.  I don't have a strong preference as long as we document. 

As for the ownership, nothing new is needed? All we have to do is add `says_hasIntegrityTag` relation to the declaration and have the following rule just like `hasTag`:
```
hasIntegrityTag(ap, owner, tag) :- says_hasIntegrityTag(owner, ap, owner, tag).`
```

Isn't that the case?

bgogul (approved):

markww:
Sounds good, I'll update to `hasAppliedIntegrityTag`.

On the `says_` rule, it's not that nothing new is needed, it's that I don't think the name we're connecting to exists in the parser yet. Regardless, I can add it now, not a big deal.

src/analysis/souffle/taint.dl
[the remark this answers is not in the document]
bgogul: > Sounds good, I'll update to `hasAppliedIntegrityTag`.
> 
> On the `says_` rule, it's not that nothing new is needed, it's that I don't think the name we're connecting to exists in the parser yet. Regardless, I can add it now, not a big deal.

We don't need any support in the parser? In any case, do it in a separate PR.

markww:
On the comment that @bgogul left that I saw in email but it looks like GitHub dropped (!?):

```
We don't need any support in the parser? In any case, do it in a separate PR.
```

We do need support in the parser eventually, definitely. But in this PR I'm trying to keep this feature contained to the DL side only, as getting it hooked up to the AuthLogic parser and the IR translator requires the feature to be more "settled" than it currently is (with our discussions about the `phi` alternative, how we handle dynamic rules with propagation, etc).

I'll get this in with the `hasAppliedIntegrityTag` change.

src/analysis/souffle/taint.dl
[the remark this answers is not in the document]
markww: Fixed name, will defer connecting to `says_` to separate PR.

PR #239, opened 2021-11-29
AST Nodes for authorization logic

aferr:
This PR includes some c++ AST nodes for authorization logic. The structure of the AST exactly follows the one in `//rust/tools/authorization-logic/src/ast.rs`, but it is reimplemented in C++. The AST nodes (and other parts) related to generating/checking cryptographic signatures or importing/exporting facts have been left out because these can be implemented as a separate feature later.


aferr:
You can ignore the failing test on cloudbuild and take a look at this now. The failing test does not do anything at the moment :)

src/ir/auth_logic/ast.cc
```diff
@@ -0,0 +1,73 @@
...
+  // AstFact
+  std::unique_ptr<AstFact>
+  AstFact::CreateFact(AstFlatFact flat_fact) {
+    return std::unique_ptr<AstFact>(new AstFact(std::move(flat_fact)));
+  }
+  
+  std::unique_ptr<AstFact>
+  AstFact::CreateFact(AstPrincipal prin, AstFact* fact) {
+    CanSayFact* can_say_fact = new CanSayFact(std::move(prin), std::move(fact));
+    return std::unique_ptr<AstFact>(new AstFact(std::move(*can_say_fact)));
```
markww, line 56: Another memory leak, of the same style as the above. Also, you do not need to use `std::move` to move `fact`, which is of type `AstFact *`. All pointers are efficiently copiable and you don't need to worry about preventing copies for them.

src/ir/auth_logic/ast.cc
```diff
@@ -0,0 +1,73 @@
...
+  AstAssertion::CreateAssertion(AstFact fact) {
+    return std::unique_ptr<AstAssertion>(new AstAssertion(std::move(fact)));
+  }
+
+  std::unique_ptr<AstAssertion>
+  AstAssertion::CreateAssertion(AstFact lhs, std::vector<AstFlatFact> rhs) {
+    CondAssertion* cond_assertion = new CondAssertion(std::move(lhs),
+        std::move(rhs));
+    return std::unique_ptr<AstAssertion>(new AstAssertion(
+        std::move(*cond_assertion)));
```
markww, line 70: Another memory leak, same pattern of misuse of `new`.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+
+class AstPredicate {
+  public:
+    AstPredicate(
+        std::string name,
+        std::vector<std::string> args,
+        AstSign sign)
+      : name_(std::move(name)),
+        args_(std::move(args)),
+        sign_(std::move(sign)) {}
```
markww, line 46: No need to `std::move` `sign`; it is an `enum` and will therefore be efficiently copiable.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+// To encapsulate which form this takes, it:
+//    - uses factory methods to make either form
+//    - uses a private std::variant member that takes either of these two forms
+//    - uses std::visit to do operations on the actual state
+class AstVerbPhrase {
+  public:
+    std::unique_ptr<AstVerbPhrase>
+      CreateVerbPhrase(AstPredicate pred);
+    std::unique_ptr<AstVerbPhrase>
+      CreateVerbPhrase(AstPrincipal right_prin);
```
markww, line 66: Factory methods that create an object must be declared `static`. Otherwise, you need to call them on a particular instance, but you don't yet have an instance upon which to call them.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+  private:
+    // The principal variant of the internal form means this represents
+    // "canActAs principal".
+    using InternalFormType = std::variant<AstPredicate, AstPrincipal>;
+    InternalFormType internal_form_;
+    AstVerbPhrase(InternalFormType internal_form)
+      : internal_form_(std::move(internal_form)) {}
```
markww, line 73: Although I am sympathetic to this use of an alias to ensure that the same type is used for the field and the constructor parameter, I believe it is a violation of the rule [not to use aliases in the public API](https://google.github.io/styleguide/cppguide.html#Aliases).

In addition, I believe that, under style guide declaration order rules, the [constructor must come before the field](https://google.github.io/styleguide/cppguide.html#Aliases).

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+//   - <principal> <verb_phrase>  (* a principal applied to a verbphrase *)
+//   - <predicate>                (* just a predicate *)
+// A similar variant-based pattern is used for this as for VerbPhrase. Here an 
+// inner struct is used for the first form (principal applied to verb_phrase).
+class AstFlatFact {
+  public:
+    std::unique_ptr<AstFlatFact>
+      CreateFlatFact(AstPrincipal prin, AstVerbPhrase verb_phrase);
+    std::unique_ptr<AstFlatFact>
+      CreateFlatFact(AstPredicate prin);
```
markww, line 86: As mentioned elsewhere, factory methods must be `static`.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+  private:
+    typedef struct VerbPhraseFlatFact {
+      AstPrincipal prin;
+      AstVerbPhrase verb_phrase;
+      VerbPhraseFlatFact(AstPrincipal prin, AstVerbPhrase verb_phrase)
+        : prin(std::move(prin)),
+          verb_phrase(std::move(verb_phrase)) {}
+    } VerbPhraseFlatFact;
```
markww, line 94: This pattern of `typedef struct <definition> StructName` is a very C way of defining a struct. In C, type aliases and struct names do not occupy the same name universe and you must qualify every `struct` name with `struct`. The pattern here introduces the `struct` name in both universes so that the `struct` qualifier is unnecessary.

C++ does not have this issue: you can describe `struct` and `class` type names without the additional `struct` qualifier. Therefore, this is unnecessary and non-idiomatic. In addition, it violates the rule about i[nternal aliases in headers](https://google.github.io/styleguide/cppguide.html#Aliases).

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+    std::unique_ptr<AstFact> CreateFact(AstFlatFact flat_fact);
+    std::unique_ptr<AstFact> CreateFact(AstPrincipal prin, AstFact* fact);
+  private:
+    typedef struct CanSayFact {
+      AstPrincipal prin;
+      AstFact *fact;
+      CanSayFact(AstPrincipal prin, AstFact* fact)
+        : prin(std::move(prin)),
+          fact(fact) {}
+    } CanSayFact;
```
markww, line 115: Fix C-style struct declaration, as mentioned elsewhere.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+    using InternalFormType = std::variant<AstFlatFact, CanSayFact>;
```
markww, line 116: [Internal alias in interface](https://google.github.io/styleguide/cppguide.html#Aliases), as mentioned elsewhere.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+      AstFact lhs;
+      std::vector<AstFlatFact> rhs;
+      CondAssertion(AstFact lhs, std::vector<AstFlatFact> rhs)
+        : lhs(std::move(lhs)), rhs(std::move(rhs)) {}
+    } CondAssertion;
+    using InternalFormType = std::variant<AstFact, CondAssertion>;
+    InternalFormType internal_form_;
+    AstAssertion(InternalFormType internal_form)
+      : internal_form_(std::move(internal_form)) {};
+};
```
markww, line 142: Fix C-style struct definition, as mentioned elsewhere. Also remove internal-only alias.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+    AstPrincipal prin_;
+    AstFact fact_;
+};
+
+class AstProgram {
+  public:
+    AstProgram(std::vector<AstSaysAssertion> says_assertions,
+        std::vector<AstQuery> queries)
+      : says_assertions_(says_assertions),
+        queries_(queries) {};
```
markww, line 174: Use `std::move` for these initializations to allow copy elision.

markww (changes requested):

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+
+#include <string>
+#include <vector>
+#include <variant>
+
+namespace raksha::ir::auth_logic {
+
+class AstPrincipal {
+  public:
+    AstPrincipal(std::string name) : name_(std::move(name)) {}
```
markww, line 30: Do we have any restrictions on what characters/format are allowed in a principal's `name`? If so, this constructor might be a good place to `CHECK` that they are honored, as it will allows us to trust that those constraints are in place for the rest of the object's lifetime.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+  private:
+    std::string name_;
+};
+
+// Used to represent whether a predicate is negated or not
+enum AstSign { negated, positive };
```
markww, line 36: Currently, these are not following the [style guide's decrees](https://google.github.io/styleguide/cppguide.html#Enumerator_Names) for naming enums.

In addition, you should consider [using `enum class` instead of `enum`](https://en.cppreference.com/w/cpp/language/enum). In C++, the classical behavior of `enum`s was that the constants declared therein would be added as members in the enclosing class or namespace. This is super weird to reason about, and so they added `enum class` in C++11, which causes the enum's type name to itself act as a name scope, to which the members belong.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+
+class AstPredicate {
+  public:
+    AstPredicate(
+        std::string name,
```
markww, line 41: Similar to what I said above, if there are restrictions on the contents or format of this `name`, the body of this constructor might be a good place to `CHECK` them.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+        std::vector<std::string> args,
```
markww, line 42: What are these `args`? Should they have a bit more AST structure added to them?

src/ir/auth_logic/ast.cc
```diff
@@ -0,0 +1,73 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#include "src/ir/auth_logic/ast.h"
+
+namespace raksha::ir::auth_logic {
```
bgogul, line 19: Can you run `clang-format` on these files? It is intended incorrectly.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,182 @@
...
+#ifndef SRC_IR_AUTH_LOGIC_AST_H_
+#define SRC_IR_AUTH_LOGIC_AST_H_
+
+#include <string>
+#include <vector>
+#include <variant>
+
+namespace raksha::ir::auth_logic {
+
+class AstPrincipal {
```
bgogul, line 28: nit: Do we need the `Ast` prefix in the name? Isn't the namespace `raksha::ir::auth_logic`specific enough?

src/ir/auth_logic/ast.h
bgogul: Also, move this into `AstPredicate` if it is only related to that class.

src/ir/auth_logic/ast.h
bgogul: Why not create public constructors for the class?

```C++
explicit AstVerbPhrase(AstPredicate predicate) {...}
explicit AstVerbPhrase(AstPrincipal predicate) {...}
```

src/ir/auth_logic/ast.h
bgogul: Do we need the factory methods or can we just have constructors as mentioned below?

Also, why do you need to return a `unique_ptr` and not an object itself?

src/ir/auth_logic/ast.cc
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // AstFlatFact
+  std::unique_ptr<AstFlatFact>
+  AstFlatFact::CreateFlatFact(AstPrincipal prin,
+      AstVerbPhrase verb_phrase) {
+    VerbPhraseFlatFact* verb_phrase_flat_fact = new VerbPhraseFlatFact(
+        prin, verb_phrase);
+    return std::unique_ptr<AstFlatFact>(new AstFlatFact(
+      std::move(*verb_phrase_flat_fact)));
```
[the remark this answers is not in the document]
bgogul: Using [`std::make_unique`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) is even better:

```C++
return std::make_unique<AstFlatFact>(VerbPhraseFlatFact(prin, verb_phrase));
```

bgogul (commented):
Can you address the comments that Mark and I left before we do another round of reviews?

src/ir/auth_logic/ast.cc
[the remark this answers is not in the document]
aferr: It looks like `make_unique` doesn't work here because the constructors are private and `make_unique` involves calling them from a different namespace.

src/ir/auth_logic/ast.h
aferr: I used them so that these are all created with smart pointers, and so that we always get the memory management features from smart pointers (prevention of leaks and double frees). It also looked like this pattern was used in a few cases in the other IR code for Raksha.

For now I kept the factory methods and made them `static`. Let me know if you'd rather I get rid of these and just make the constructors public. I have no opinion about this.

src/ir/auth_logic/ast.h
aferr: fixed.

src/ir/auth_logic/ast.h
aferr: done.

src/ir/auth_logic/ast.h
aferr: done.

src/ir/auth_logic/ast.h
aferr: done.

src/ir/auth_logic/ast.h
aferr: oops, fixed.

src/ir/auth_logic/ast.h
aferr: I might leave this around for a moment longer because it happens to prevent an unused private field `-Werr` about `sign_` but I left a TODO

src/ir/auth_logic/ast.h
aferr: I have no idea yet if using these variants to write more substantial code becomes unwieldy, though. I guess I/we'll find out shortly.

src/ir/auth_logic/ast.h
aferr: The parser should already check that the principal names are in a particular format, so we shouldn't need to replicate that here, I think.

src/ir/auth_logic/ast.h
aferr: I think removing it is better, too. Done.

src/ir/auth_logic/ast.h
aferr: - Fixed style
- Looks like I can't make Sign an inner class because it is used as a parameter for the constructor of Predicate

src/ir/auth_logic/ast.h
aferr: This predicate is `pred_name(arg_1, ..., arg_n)`, and the args are the `arg_1, ..., arg_n` bit. I don't see how more structure could be usefully added.

src/ir/auth_logic/ast.h
aferr: resolving this one, see the other one.

src/ir/auth_logic/ast.cc
aferr: done.

aferr:
@bgogul @markww see unresolved conversations

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,200 @@
...
+//   - <principal> <verb_phrase>  (* a principal applied to a verbphrase *)
+//   - <predicate>                (* just a predicate *)
+// A similar variant-based pattern is used for this as for VerbPhrase. Here an
+// inner struct is used for the first form (principal applied to verb_phrase).
+class FlatFact {
+ public:
+  static std::unique_ptr<FlatFact> CreateFlatFact(Principal prin,
+                                                  VerbPhrase verb_phrase) {
+    return std::unique_ptr<FlatFact>(
+        new FlatFact(VerbPhraseFlatFact(prin, verb_phrase)));
```
markww, line 89: You should probably `std::move` `prin` and `verb_phrase` here so that they have no live names when being passed into the `VerbPhraseFlatFact` constructor and can be copy elided.

src/ir/auth_logic/ast.h
bgogul: Why would `move` suppress the unused field error?

src/ir/auth_logic/ast.h
bgogul: In this case, defining public constructors would be more versatile. We can have both value types as well as unique_ptr if needed. See my comment related to `VerbPhrase` definition.

src/ir/auth_logic/ast.h
bgogul: Regarding my comment about constructors what I meant was define constructors as follows:

```C++
class VerbPhrase {
 public:
   explicit VerbPhrase(Predicate pred): internal_form_(pred) {}
   explicit VerbPhrase(Principal right_prin): internal_form_(right_prin) {}
 private:
  // The principal variant of the internal form means this represents
  // "canActAs principal".
  std::variant<Predicate, Principal> internal_form_;
};
```

bgogul (commented):
Left some comments. PTAL.

src/ir/auth_logic/ast.h
aferr: I thought the `std::move`s that the constructor for VerbPhraseFlatFact already does would be enough. In any case, I will add them.

aferr:
Since this was already reviewed, is this OK to merge as is (if the comments have been addressed), and then I can add a ToString and some tests as a follow-up PR? Or would you rather I do that in this PR?

src/ir/auth_logic/ast.h
aferr: I have no idea why it suppresses that error because this doesn't look like a use of `sign_` at all, but it does suppress the error!

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,175 @@
...
+class Fact {
+ public:
+  explicit Fact(FlatFact flat_fact) : internal_form_(std::move(flat_fact)) {}
+  explicit Fact(Principal prin, Fact* fact)
+    : internal_form_(CanSayFact(std::move(prin), fact)) {}
+
+ private:
+  struct CanSayFact {
+    Principal prin;
+    Fact* fact;
```
markww, line 108: Weren't you going to make this bare pointer a `std::unique_ptr` to remove the lifetime concern for `Fact`?

If using a `std::unique_ptr` doesn't work here, it is probably best to make this a `const Fact *`, as you usually don't expect AST nodes to be altered after construction.

markww (approved):
Looks good, other than the one remaining question about the bare `Fact` pointer.

bgogul:
> Since this was already reviewed, is this OK to merge as is (if the comments have been addressed), and then I can add a ToString and some tests as a follow-up PR? Or would you rather I do that in this PR?

As practice, let us always add tests when creating new functionality, unless there is a strong reason not to. However, I am not sure what tests you will add here as these are just defining the data structures. 

That said,  I have some suggestions and questions (working on it). So, let us not merge it in yet.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+  Principal(std::string name) : name_(std::move(name)) {}
+
+ private:
+  std::string name_;
+};
+
+// Used to represent whether a predicate is negated or not
+enum Sign { kNegated, kPositive };
+
+class Predicate {
```
bgogul, line 39: A brief documentation for this class? Does this effectively correspond to a datalog relation?

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+  std::vector<std::string> args_;
+  Sign sign_;
+};
+
+// This represents an expression that applies either a predicate or a
+// "canActAs" to a principal, and it is a sub-term of FlatFact
+//    - pred_name(arg_1, ..., arg_n)           (a predicate)
+//    - principal_1 canActAs principal_2       (a canActAs)
+// To encapsulate which form this takes, it:
+//    - uses factory methods to make either form
```
bgogul, line 59: stale documentation?

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+//    - uses a private std::variant member that takes either of these two forms
+//    - uses std::visit to do operations on the actual state
+class VerbPhrase {
```
bgogul, line 62: curious. Why is this called a `VerbPhrase`? Also, this seems to combine seemingly unrelated concepts into one variant?

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+      : internal_form_(std::move(right_prin)) {}
+
+ private:
+  // The principal variant of the internal form means this represents
+  // "canActAs principal".
+  std::variant<Predicate, Principal> internal_form_;
+};
+
+// This the AST Node for FlatFact which takes on one of two forms:
+//   - <principal> <verb_phrase>  (* a principal applied to a verbphrase *)
```
bgogul, line 75: What would this correspond to in the authorization logic syntax? I am asking because none of the following are valid syntax:
  - `"UserX" "UserY"`
  - `"UserX" claimHasTag("...")`

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+//   - <predicate>                (* just a predicate *)
+// A similar variant-based pattern is used for this as for VerbPhrase. Here an
+// inner struct is used for the first form (principal applied to verb_phrase).
+class FlatFact {
```
bgogul, line 79: This question is mostly for my education: what does `Flat` signify? I am wondering if there is a different prefix (like `Base`) that will be more relevant to use here.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+    std::vector<FlatFact> rhs;
+    CondAssertion(Fact lhs, std::vector<FlatFact> rhs)
+        : lhs(std::move(lhs)), rhs(std::move(rhs)) {}
+  };
+  std::variant<Fact, CondAssertion> internal_form_;
+};
+
+// AST node for saysAssertion, which prepends an assertion with:
+// "<principal> says"
+class SaysAssertion {
```
bgogul, line 137: Do we generally want to group all assertions made by a `Principal`? If so, would it be better to add a map to the `Program` data structure below?

src/ir/auth_logic/ast.h
aferr: @markww oops, I just forgot, but I pushed a fix now.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+ public:
+  explicit Fact(FlatFact flat_fact) : internal_form_(std::move(flat_fact)) {}
+  explicit Fact(Principal prin, Fact* fact)
+      : internal_form_(CanSayFact(std::move(prin), std::move(fact))) {}
+
+ private:
+  struct CanSayFact {
+    Principal prin;
+    std::unique_ptr<Fact> fact;
+    CanSayFact(Principal prin, Fact* fact)
```
bgogul, line 109: This should be `std::unique_ptr<Fact>` and not `Fact*`.

src/ir/auth_logic/ast.h
bgogul: We agreed to get rid of this and make `FlatFact` to have three variants: `Predicate`, `PrincipalAttribute`,  `CanActAs`.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+//  - <flat_fact>
+//  - <principal> canSay <flat_fact>
+class Fact {
+ public:
+  explicit Fact(FlatFact flat_fact) : internal_form_(std::move(flat_fact)) {}
+  explicit Fact(Principal prin, Fact* fact)
+      : internal_form_(CanSayFact(std::move(prin), std::move(fact))) {}
+
+ private:
+  struct CanSayFact {
```
bgogul, line 106: Move this out.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+//  - <fact>                                      (unconditional)
+//  - <fact> :- <flat_fact_1> ... <flat_fact_n>   (conditional)
+class Assertion {
+ public:
+  explicit Assertion(Fact fact) : internal_form_(std::move(fact)) {}
+  explicit Assertion(Fact lhs, std::vector<FlatFact> rhs)
+      : internal_form_(CondAssertion(std::move(lhs), std::move(rhs))) {}
+
+ private:
+  struct CondAssertion {
```
bgogul, line 126: Use full name `ConditionalAssertion`.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+
+// AST node for saysAssertion, which prepends an assertion with:
+// "<principal> says"
+class SaysAssertion {
+ public:
+  explicit SaysAssertion(Principal prin, std::vector<Assertion> assertions)
+      : prin_(std::move(prin)), assertions_(std::move(assertions)){};
+
+ private:
+  Principal prin_;
```
bgogul, line 143: `principal_`

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+class Query {
+ public:
+  explicit Query(std::string name, Principal prin, Fact fact)
+      : name_(std::move(name)),
+        prin_(std::move(prin)),
+        fact_(std::move(fact)) {}
+
+ private:
+  std::string name_;
+  Principal prin_;
```
bgogul, line 158: `principal_`

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+  Fact fact_;
+};
+
+class Program {
+ public:
+  explicit Program(std::vector<SaysAssertion> says_assertions,
+                   std::vector<Query> queries)
+      : says_assertions_(std::move(says_assertions)),
+        queries_(std::move(queries)){};
```
bgogul, line 167: `;` is redundant.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,176 @@
...
+
```
bgogul, line 168:
```C++
const std::vector<SaysAssertion> says_assertion()  const { return says_assertions_; }
```

src/ir/auth_logic/ast.h
aferr: done

src/ir/auth_logic/ast.h
aferr: fixed.

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,175 @@
...
+//    - uses std::visit to do operations on the actual state
+class VerbPhrase {
+ public:
+  explicit VerbPhrase(Predicate pred): internal_form_(std::move(pred)) {}
+  explicit VerbPhrase(Principal right_prin): internal_form_(std::move(right_prin)) {}
+
+ private:
+  // The principal variant of the internal form means this represents
+  // "canActAs principal".
+  std::variant<Predicate, Principal> internal_form_;
```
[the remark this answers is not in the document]
aferr: yeah

src/ir/auth_logic/ast.h
aferr: we discussed by video

src/ir/auth_logic/ast.h
aferr: name is now base as discussed

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,175 @@
...
+// inner struct is used for the first form (principal applied to verb_phrase).
+class FlatFact {
+ public:
+  explicit FlatFact(Principal prin, VerbPhrase verb_phrase) 
+    : internal_form_(VerbPhraseFlatFact(std::move(prin),
+          std::move(verb_phrase))) {}
+  explicit FlatFact(Predicate pred) : internal_form_(std::move(pred)) {}
+
+ private:
+  struct VerbPhraseFlatFact {
```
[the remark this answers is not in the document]
aferr: fixed.

src/ir/auth_logic/ast.h
aferr: done

src/ir/auth_logic/ast.h
aferr: I made these shared pointers so that copies can be made for accessors, which seems better than having accessors take ownership of the memory.

src/ir/auth_logic/ast.h
aferr: done

src/ir/auth_logic/ast.h
aferr: I think there is not a use-case for any compiler passes the Rust verison has (so we won't need them here either). If we need this later, I thing we'd be better off doing it in a separate pass (and implementing this just when we need it).

src/ir/auth_logic/ast.h
aferr: done

src/ir/auth_logic/ast.h
aferr: done.

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h
aferr: i made accessors everywhere

bgogul (commented):
You should return `const &` for the accessors.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+ public:
+  Predicate(std::string name, std::vector<std::string> args, Sign sign)
+      : name_(std::move(name)),
+        args_(std::move(args)),
+        // TODO move surpresses an unused private field error about sign_
+        // for now, get rid of this eventually
+        sign_(std::move(sign)) {}
+
+  const std::string name() const { return name_; }
+  const std::vector<std::string> args() const { return args_; }
```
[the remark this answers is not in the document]
aferr: done.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+};
+
+// Attribute corresponds to an attribute of a principal with the form
+// <Principal> <pred_name>(arg_1, ..., arg_n)
+class Attribute {
+ public:
+  explicit Attribute(Principal principal, Predicate predicate)
+    : principal_(principal), predicate_(predicate) {}
+  const Principal principal() const { return principal_; }
+  const Predicate predicate() const { return predicate_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+// CanSay corresponds to an expression of the form <principal> canSay Fact
+class CanSay {
+  public:
+    explicit CanSay(Principal principal, std::shared_ptr<Fact>& fact)
+      : principal_(principal), fact_(std::move(fact)) {}
+    const Principal principal() const { return principal_; }
+    const std::shared_ptr<Fact> fact() const { return fact_; }
+  private:
+    Principal principal_;
+    std::shared_ptr<Fact> fact_;
```
markww: Google's style guide all but bans `std::shared_ptr`. Every pointer should either be a `unique_ptr` or a regular old `Type *` or `Type &`.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+
+// CanActAs corresponds to a canActAs expression of the form
+// <principal_1> canActAs <principal_2>
+class CanActAs {
+ public:
+  explicit CanActAs(Principal left_principal, Principal right_principal)
+    : left_principal_(left_principal),
+      right_principal_(right_principal) {}
+  const Principal left_principal() const { return left_principal_; }
+  const Principal right_principal() const { return right_principal_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+//  - An attribute
+//  - A canActAs expression
+// The particular choice of form is represented with a variant which is 
+// encapsulated.
+class BaseFact {
+ public:
+  explicit BaseFact(Predicate predicate) : value_(std::move(predicate)) {};
+  explicit BaseFact(Attribute attribute) : value_(std::move(attribute)) {};
+  explicit BaseFact(CanActAs canActAs) : value_(std::move(canActAs)) {};
+  const std::variant<Predicate, Attribute, CanActAs> value() const {
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+    std::shared_ptr<Fact> fact_;
+};
+
+// Fact corresponds to either a base fact or a an expression of the form 
+// <principal> canSay 
+class Fact {
+ public:
+  explicit Fact(BaseFact base_fact) : value_(std::move(base_fact)) {}
+  explicit Fact(std::shared_ptr<CanSay>& can_say)
+    : value_(std::move(can_say)) {}
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+// CanSay and Fact are forward-declared because they are mutually recursive.
+class CanSay;
+class Fact;
+
+// CanSay corresponds to an expression of the form <principal> canSay Fact
+class CanSay {
+  public:
+    explicit CanSay(Principal principal, std::shared_ptr<Fact>& fact)
+      : principal_(principal), fact_(std::move(fact)) {}
+    const Principal principal() const { return principal_; }
```
markww: Should also return a reference or a pointer to `Principal` rather than making a copy.

src/ir/auth_logic/ast_test.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "src/common/testing/gtest.h"
+#include <iostream>
+
+namespace raksha::ir::auth_logic {
+
+TEST(AstTest, PrincipalAccessors) {
+  // Principal principal = Principal("TestPrincipal");
+  // std::cout << "test principal name" << principal.name() << std::endl;
+  EXPECT_TRUE(true);
+  // EXPECT_EQ(principal.name(), "TestPrincipal");
```
markww, line 27: This commented-out test code should either be resurrected or deleted.

markww (changes requested):

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+};
+
+// Fact corresponds to either a base fact or a an expression of the form 
+// <principal> canSay 
+class Fact {
+ public:
+  explicit Fact(BaseFact base_fact) : value_(std::move(base_fact)) {}
+  explicit Fact(std::shared_ptr<CanSay>& can_say)
+    : value_(std::move(can_say)) {}
+  const std::variant<BaseFact, std::shared_ptr<CanSay>> value() const {
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+    return value_;
+  }
+
+ private:
+  std::variant<BaseFact, std::shared_ptr<CanSay>> value_;
```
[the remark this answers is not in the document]
aferr: This one shouldn't be a unique_ptr? I guess if this is not a unique ptr, the constructor for Fact should have `const CanSay*` as the parameter instead of `std::unique_ptr<CanSay>` ?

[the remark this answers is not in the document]
aferr: If I leave the parameter of the constructor for Fact as a `unique_ptr<CanSay>` but change the value_ to be a `const CanSay*` I get:
```
./src/ir/auth_logic/ast.h:134:7: error: no matching constructor for initialization of 'std::variant<BaseFact, const CanSay *>'
```

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+
+// ConditionalAssertion the particular form of assertion that can have 
+// conditions which is:
+// <fact> :- <flat_fact_1> ... <flat_fact_n>
+class ConditionalAssertion {
+  public:
+    explicit ConditionalAssertion(Fact lhs, std::vector<BaseFact> rhs)
+      : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
+    const Fact lhs() const { return lhs_; }
+    const std::vector<BaseFact> rhs() const { return rhs_; }
```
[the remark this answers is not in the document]
aferr: done.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+// Assertions can have two forms:
+//  - Facts
+//  - Conditional assertions
+class Assertion {
+ public:
+  explicit Assertion(Fact fact) : value_(std::move(fact)) {}
+  explicit Assertion(ConditionalAssertion conditional_assertion)
+    : value_(std::move(conditional_assertion)) {}
+
+  const std::variant<Fact, ConditionalAssertion> value() const {
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+};
+
+// SaysAssertion prepends an assertion with <principal> says"
+class SaysAssertion {
+ public:
+  explicit SaysAssertion(Principal principal, std::vector<Assertion> assertions)
+      : principal_(std::move(principal)),
+        assertions_(std::move(assertions)) {};
+  const Principal principal() const { return principal_; }
+  const std::vector<Assertion> assertions() const { return assertions_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+// given the rules and base facts.
+class Query {
+ public:
+  explicit Query(std::string name, Principal principal, Fact fact)
+      : name_(std::move(name)),
+        principal_(std::move(principal)),
+        fact_(std::move(fact)) {}
+  const std::string name() const { return name_; }
+  const Principal principal() const { return principal_; }
+  const Fact fact() const { return fact_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+
+// This is a top-level program which can contain 
+class Program {
+ public:
+  explicit Program(std::vector<SaysAssertion> says_assertions,
+                   std::vector<Query> queries)
+      : says_assertions_(std::move(says_assertions)),
+        queries_(std::move(queries)) {}
+  std::vector<SaysAssertion> says_assertions() { return says_assertions_; }
+  std::vector<Query> queries() { return queries_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+class CanSay;
+class Fact;
+
+// CanSay corresponds to an expression of the form <principal> canSay Fact
+class CanSay {
+  public:
+    explicit CanSay(Principal principal, std::shared_ptr<Fact>& fact)
+      : principal_(principal), fact_(std::move(fact)) {}
+    const Principal principal() const { return principal_; }
+    const std::shared_ptr<Fact> fact() const { return fact_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,223 @@
...
+#include <string>
+#include <variant>
+#include <vector>
+
+namespace raksha::ir::auth_logic {
+
+class Principal {
+ public:
+  explicit Principal(std::string name) : name_(std::move(name)) {}
+  std::string name() { return name_; }
```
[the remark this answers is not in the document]
aferr: done

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,225 @@
...
+  Predicate predicate_;
+};
+
+// CanActAs corresponds to a canActAs expression of the form
+// <principal_1> canActAs <principal_2>
+class CanActAs {
+ public:
+  explicit CanActAs(Principal left_principal, Principal right_principal)
+    : left_principal_(left_principal),
+      right_principal_(right_principal) {}
```
markww, line 80: These initializations should use `std::move`.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,225 @@
...
+    Principal principal_;
+    std::unique_ptr<Fact> fact_;
+};
+
+// Fact corresponds to either a base fact or a an expression of the form 
+// <principal> canSay 
+class Fact {
+ public:
+  explicit Fact(BaseFact base_fact) : value_(std::move(base_fact)) {}
+  explicit Fact(const CanSay* can_say)
```
markww, line 133: Is this `CanSay` not owned by the `Fact`? Could it be made into a `std::unique_ptr<CanSay>`, or is it pointed to elsewhere?

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,225 @@
...
+    : value_(std::move(can_say)) {}
+  const std::variant<BaseFact, const CanSay*>& value() const {
+    return value_;
+  }
```
markww, line 137: I find this accessor interesting. It's not necessarily wrong, but usually if a class has a tagged union, it's that class's job to unpack the union.

Do we expect that the client will have no expectation of which of the possibilities they will get, or will they have a strong idea of what they are expecting and we could get away with two different accessors, each of which asserts if the wrong choice is made? If it is the latter, we should provide two accessors which check the client's mental model and, if correct, unpack the value from the variant and return.

But a deeper question: is returning this field the right choice? Usually, ASTs are traversed using a Visitor, and `std::variant`s have a very good mechanism for visiting all of their possibilities. If we expect to be using that idiom, then this field should not be made visible to the outside world at all. We should simply expose a `visit` method that takes a visitor and does the "right thing".

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,225 @@
...
+
+// ConditionalAssertion the particular form of assertion that can have 
+// conditions which is:
+// <fact> :- <flat_fact_1> ... <flat_fact_n>
+class ConditionalAssertion {
+  public:
+    explicit ConditionalAssertion(Fact lhs, std::vector<BaseFact> rhs)
+      : lhs_(std::move(lhs)), rhs_(std::move(rhs)) {}
+    const Fact& lhs() const { return lhs_; }
+    const std::vector<BaseFact>& rhs() const { return rhs_; }
```
markww, line 151: Similar to my comment above, should we be exposing these members or should we be visiting them?

markww (commented):
The high level theme of most of my comments here: I was distracted by the accessors making copies, but I do wonder whether they should be returning their child nodes outside of the class at all. Generally, an AST tries to make its possible children a bit of a private matter and expects to receive a Visitor class, which can handle any given child. This is particularly nice for operations in which AST nodes should mostly be considered separately anyway. I think most of the accessors should probably go and instead a `visit` method should be used to consider the AST contents.

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
aferr: > The high level theme of most of my comments here: I was distracted by the accessors making copies, but I do wonder whether they should be returning their child nodes outside of the class at all. Generally, an AST tries to make its possible children a bit of a private matter and expects to receive a Visitor class, which can handle any given child. This is particularly nice for operations in which AST nodes should mostly be considered separately anyway. I think most of the accessors should probably go and instead a `visit` method should be used to consider the AST contents.

I was thinking about making a visitor as well, but a visitor in C++ didn't seem all that useful because since it lacks generics, the only way to write a visitor with a return is to have the return type `void *`. The other option is to never return anything with a visitor and merely fake returns by adding field to the visitor and side-effecting in it, but I find the side-effect style confusing. So I didn't like any of the visitors in C++ that I saw. Also, the functions that need to be implemented on the AST (translation into a lower level datalog IR, and just printing) need to do something in between visits to the children, so at least the style of visitor that keeps the order of traversal fixed for the AST does not work here.

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
markww: If you use a `std::unique_ptr`, you want to use it in both the constructor and in the field. So both the `value` parameter and the `value_` field should have the type `std::variant<BaseFact, std::unique_ptr<CanSay>>`.

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
aferr: Maybe the style of visitor that: 1) can return `void *`, 2) and does not fix the order of traversal in children would be fine.

src/ir/auth_logic/ast_test.cc
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#include "src/ir/auth_logic/ast.h"
+#include "src/common/testing/gtest.h"
+#include <iostream>
+#include <memory>
```
markww, line 20: This is a nit, but the inclusion order is wrong here. According to the Google style guide, the inclusion order should have the associated  header first (ast.h), then system headers (iostream and memory), then project headers (gtest.h).

aferr:
@bgogul @markww I think I will make a visitor in a followup PR, and remove all the accessors (or at least all the accessors to variants). LMK If you'd rather I remove all the accessors or just the variant ones.

src/ir/auth_logic/ast.h
aferr: done

src/ir/auth_logic/ast.h
aferr: fixed.

src/ir/auth_logic/ast.h
aferr: just removed the variant accessors for now.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,215 @@
...
+};
+
+// CanSay and Fact are forward-declared because they are mutually recursive.
+class CanSay;
+class Fact;
+
+// CanSay corresponds to an expression of the form <principal> canSay Fact
+class CanSay {
+  public:
+    explicit CanSay(Principal principal, std::unique_ptr<Fact>& fact)
```
markww, line 116: This should be a `std::unique_ptr<Fact>` rather than a reference to the `std::unique_ptr<Fact>`. The `std::move` call will force the old name to be erased when being placed in `fact_`, thus allowing only one live unique pointer to exist at a time.

src/ir/auth_logic/ast.h
```diff
@@ -0,0 +1,215 @@
...
+};
+
+// Assertions can have two forms:
+//  - Facts
+//  - Conditional assertions
+class Assertion {
+ public:
+  explicit Assertion(Fact fact) : value_(std::move(fact)) {}
+  explicit Assertion(ConditionalAssertion conditional_assertion)
+    : value_(std::move(conditional_assertion)) {}
```
markww, line 158: I think this code is fine, and you don't need to change anything.

That said, it may save you some typing in the future to know that if you have a parameter of type `std::variant<A, B>`, that parameter can receive a value of type `A`, type `B`, or `std::variant<A, B>` into that parameter. It can be quite nice for constructors as well, as you basically get two for the price of one.

markww (approved):
LGTM

PR #241, opened 2021-11-30
Return error status from bazel_run functions.

bgogul:
Otherwise, gcb incorrectly thinks that the tests and builds pass. 

See the new behavior with an intentional failure: https://pantheon.corp.google.com/cloud-build/builds/357b1fa2-6414-4087-b2ff-39e8fb5ccffa;step=3?project=google.com:raksha-ci


markww (approved):
Good catch, LGTM

PR #243, opened 2021-12-02
Prototype propagating tags using phi operations. 

markww:
This commit contains a radical rethinking of not only the propagation of
integrity tags, but our model of the dataflow graph in general. We
introduce the idea of operations, where an operation is some simple
calculation upon values in input `AccessPath`s that yields some result
into some other `AccessPath`. An `edge` is just an alias for the assign
(`=`) operation.

We have changed the dataflow model for `AccessPath`s to be SSA-style,
meaning that each AccessPath can receive the result of at most one
operation. To handle circumstances where we would normally target an
AccessPath with multiple `edge`s, we introduce a binary `phi` operation.
A `phi` operation just merges the abstract state of two incoming
`AccessPath`s. A `phi` results in a confidentiality tag if either of its
inputs had that confidentiality tag. A `phi` results in an integrity tag
if both of its inputs had that integrity tag. `phi`s can be chained to
handle many incoming edges.

This new way of organizaing the dataflow graph means that abstract state
shall never merge at `AccessPath`s; they shall always merge within
operations. We shall impose the restriction that no `operation` shall be
variadic; all must have a set number of inputs, and may be chained to
handle more inputs. This allows us to no longer use negation for
`forall` in datalog; we can exhaustively test all inputs for any
operation. This allows us to propagate integrity tags without negation
and without stratifiability issues.

Note that this commit only introduces this new regime for datalog fact
tests to prove that this can be done; it turns off the new SSA
requirement for code generated from the C++ side of Raksha to reduce the
amount of work taken to show a first prototype of this approach. This
also turns off the test of not-globally-trusted `derives from`, as that
will need to be reworked to fit into this new model.


markww:
We're not going with this solution, as #249 seems better. Some of this code may reappear in other PRs, as we are going to use the `operation` relation. But this proposed solution for `IntegrityTag` propagation, and thus this PR, is done.

PR #244, opened 2021-12-02
Replace `$` functor with new syntax `autoinc()`.

### LINKED ISSUE #237: Rename the Souffle `$` functor to `autoinc` wherever it appears.
In Souffle 2.1 (which we switched to in #235), the `$` functor has been renamed to `autoinc`. Although they still accept the `$` name, they now print a warning discouraging its use. We should rename `$` to `autoinc` throughout Raksha.

markww:
Between our previous and current version of Souffle, Souffle renamed the
`$` functor to `autoinc()`. While the `$` synatx still works, it
generates annoying warnings. This commit replaces the one usage of `$`
with `autoinc()`, thus getting rid of those warnings. This fixes #237.


bgogul (approved):
Thank you!

PR #245, opened 2021-12-06
Replace manual args with StructOpt

### LINKED ISSUE #8: [Authorization Logic] Look Into StructOpt for command line args

bMacSwigg:
Adapts the existing four command-line arguments into proper [clap::Arg](https://docs.rs/clap/2.33.1/clap/struct.Arg.html)s. Updates README documentation accordingly. This resolves #8.

This is a re-do of #181, because I messed up Git branches (the curse of mostly working with internal non-Git tools, alas).

In addition to the built-in parsing and typing, this also automatically transforms documentation into help text. For example:
```
$> cargo run -- -h
[...]
auth-logic 0.1.0
An auth logic compiler.

USAGE:
    auth-logic-prototype [OPTIONS] <filename> --in-dir <in-dir> --out-dir <out-dir>

FLAGS:
    -h, --help       Prints help information
        --skip-souffle    Whether to skip running Souffle on the compiled datalog
    -V, --version    Prints version information

OPTIONS:
    -s, --skip <decl-skip>...    List of declarations to skip when generating Souffle code
    -i, --in-dir <in-dir>        Input directory where the auth logic program lives. If unspecified, it will be the
                                 current working directory [env: PWD=/home/mcswiggen/raksha/rust/tools/authorization-
                                 logic]
    -o, --out-dir <out-dir>      Output directory, where the compiled datalog program will be saved. If unspecified, it
                                 will be the current working directory [env:
                                 PWD=/home/mcswiggen/raksha/rust/tools/authorization-logic]

ARGS:
    <filename>    The name of the auth logic program to compile
```

Note that in the normal way of running the commands (with `cargo run`), you now have to append an extra `--` before the arguments, so that they get passed on to the auth-logic program rather than interpreted by `cargo`.

Now that `decl_skip` is always a `Vec<String>`, I think we could simplify the parameter types of the functions that use it. In particular, the `Option<Vec<String>>` (in `ast_to_souffle_file` and `emit_program`) doesn't seem to be used for anything, since it's always Some() when it's passed in. (And if no values are passed for `decl_skip` we get an empty Vec, which is what None is transformed into anyway.) But, for now I've opted to limit the scope of this change.


bMacSwigg:
@bgogul This is still failing in the Bazel build with the [error](https://gist.github.com/aferr/3555ef4299a57a65f3e38d7139c60574) @aferr posted before, about `proc_macro2`. If you have a chance to take a look or have suggestions of things to try, I'd welcome ideas. This isn't blocking me on anything else though (now that I've moved the change out of my `main` branch, oops), so there's no urgency.

bgogul:
> @bgogul This is still failing in the Bazel build with the [error](https://gist.github.com/aferr/3555ef4299a57a65f3e38d7139c60574) @aferr posted before, about `proc_macro2`. If you have a chance to take a look or have suggestions of things to try, I'd welcome ideas. This isn't blocking me on anything else though (now that I've moved the change out of my `main` branch, oops), so there's no urgency.

Thanks, Brendan. I will patch it in figure out what is happening. I may not get to it right away, but will try to resolve it in the next couple of days.

bMacSwigg:
> > @bgogul This is still failing in the Bazel build with the [error](https://gist.github.com/aferr/3555ef4299a57a65f3e38d7139c60574) @aferr posted before, about `proc_macro2`. If you have a chance to take a look or have suggestions of things to try, I'd welcome ideas. This isn't blocking me on anything else though (now that I've moved the change out of my `main` branch, oops), so there's no urgency.
> 
> Thanks, Brendan. I will patch it in figure out what is happening. I may not get to it right away, but will try to resolve it in the next couple of days.

Thanks! I cherry-picked #255 on top of this, and it now seems to build successfully :)

aferr:
@bMacSwigg is this ready for review once the conflicts are resolved?

bMacSwigg:
> @bMacSwigg is this ready for review once the conflicts are resolved?

Yup! It looks like the conflicts are just in slightly different versions of the structopt dependencies. I'll clean those up and re-run raze when I'm back on my other computer.

bMacSwigg:
Resolved the merge conflicts and built locally, so it should be good now. If I re-run `cargo raze` locally it wants to change things again (e.g. go back to `raze__structopt__0_3_23` from `raze__structopt__0_3_25`, etc.), I'm guessing just from some slight differences in our local environments? In any case, `bazel build` and `cargo test` both work with it as it is now, so I'm inclined to ignore my local `cargo raze`.

aferr:
FYI we have a convention of naming all of our branches `some_name@username` to avoid collisions. So in the future, could you name your branches `something@bMacSwigg`? I might not have mentioned this before. Keeping the branch name as-is is fine for this PR.

rust/tools/authorization-logic/README.md
```diff
@@ -77,11 +77,19 @@ claim, and the recipient uses the public key to check the signature.
 To compile a program called FILENAME stored in INPUT_DIR, run:
 
 ```
-cargo run FILENAME INPUT_DIR OUTPUT_DIR
+cargo run -- --in-dir INPUT_DIR --out-dir OUTPUT_DIR FILENAME
```
aferr, line 80: Is the first `--` just because this is being run through cargo? If we build the binary, and run it outside of cargo, would we drop the first `--`? Just curious.

rust/tools/authorization-logic/src/compilation_top_level.rs
```diff
@@ -38,15 +38,14 @@ fn source_file_to_ast(filename: &str, in_dir: &str) -> AstProgram {
...
-                             .split(',')
-                             .map(|s| s.to_string())
-                             .collect());
-    souffle_interface::ast_to_souffle_file(&prog_with_imports, filename,
-                                           out_dir, &decl_skip_vec);
+    souffle_interface::ast_to_souffle_file(
+        &prog_with_imports,
+        filename,
+        out_dir,
+        &Some(decl_skip.to_vec()),
```
aferr, line 48: Nice, the `decl_skip: &Vec<string>` makes the implementation easier to read.

rust/tools/authorization-logic/src/main.rs
```diff
@@ -25,14 +25,46 @@ mod souffle;
 mod test;
 
 use std::env;
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+#[structopt(name = "auth-logic", about = "An auth logic compiler.")]
```
aferr, line 31: I think it's better to say `"An authorization logic compiler."` spelled out.

aferr (approved):

third_party/cargo/remote/BUILD.autocfg-1.0.1.bazel
```diff
@@ -10,9 +10,10 @@ load("@bazel_skylib//lib:selects.bzl", "selects")
 
 # buildifier: disable=load
 load(
-    "@rules_rust//rust:rust.bzl",
+    "@rules_rust//rust:defs.bzl",
```
bgogul, line 13: It looks like the newer version of `cargo raze` uses `defs.bzl` instead of `rust.bzl`. In any case, if these changes are needed, can we split it off to a different PR, please?

bgogul (commented):
It looks like the behavior of the `cargo raze` changed. Can you split off the package updates to different PR, please?

third_party/cargo/remote/BUILD.autocfg-1.0.1.bazel
aferr: @bgogul 

How is it possible to split the changes caused by `cargo raze` into a separate PR given that the dependency is needed for these code changes to run? Do you mean there should be a PR that adds the needed dependencies first and then a second PR comes after that one to use the new dependencies?

bMacSwigg:
Closing since this is now merged in #461 and #462

Issue #196, opened 2021-10-27
[Authorization Logic] Support semicolon (or) in authorization logic


aferr:
CC: (so that you know this does not work right now) @markww @bgogul

aferr:
To give a bit more detail, [semicolon is the syntax souffle uses to support disjunction (or) over predicates in rules](https://souffle-lang.github.io/rules#disjunction). It would be useful if our authorization logic also supported this. To support this, you'll need to:
     - Extend the [parser-generator] syntax for `#hornClausAssertion`s.
     - Extend the [AST node](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/ast.rs) for `AstAssertion` so that it can include disjunction
     - Extend the [Datalog IR node](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/datalog_ir.rs) for `DLIRAssertions` so that the predicates on the RHS can include disjunction.
     - Extend the [souffle emitter](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/souffle_emitter.rs) so that disjunctions are emitted using the souffle syntax.
     - You may also need to add support for wrapping disjunctions of predicates in parenthesis.

PR #246, opened 2021-12-06
[WIP] Disjunction

bMacSwigg:
This is a draft for #196. So far:
- Extends the ANTLR definition to add disjunctions
- Extends the AST parser to parse disjunctions
- Adds a new unit test file for the parser

Creating a quick PR to validate the structure & AST parsing make sense, before I continue plumbing this through the rest of the code. (Note that this doesn't build right now, because it isn't fully plumbed through yet.)

### DIFF
--- a/rust/tools/authorization-logic/src/ast.rs
+++ b/rust/tools/authorization-logic/src/ast.rs
@@ -54,58 +54,58 @@ impl PartialEq for AstPredicate {
 // See https://doc.rust-lang.org/std/cmp/trait.Eq.html
 impl Eq for AstPredicate {}
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub enum AstVerbPhrase {
     AstPredPhrase { p: AstPredicate },
     AstCanActPhrase { p: AstPrincipal },
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub enum AstFlatFact {
     AstPrinFact { p: AstPrincipal, v: AstVerbPhrase },
     AstPredFact { p: AstPredicate },
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub enum AstFact {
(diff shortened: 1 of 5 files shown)


aferr:
You went over this via GVC, but thought I'd leave a note:
 - PartialEq and Eq are probably fine, but if you don't mind taking a moment to check the docs to ensure that there aren't some unexpected implications, that would be helpful
 - You pointed out that the order of flat_facts/disjunctions in construct_conjunction will not be preserved. This is fine semantically because of commutativity, but a way that might let you preserve the order anyway is to define another parse term that is just `flatFact  | disjunction` (just like AstConjunctiveTerm) and then write the rule for conjunctions in terms of that.

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -54,58 +54,74 @@ impl PartialEq for AstPredicate {
 // See https://doc.rust-lang.org/std/cmp/trait.Eq.html
 impl Eq for AstPredicate {}
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
```
bMacSwigg, line 57: Discussed with @aferr about whether there was any risk (e.g. performance hit) to deriving these. I believe the answer is no, other than a small (likely negligible) increase to compile time and binary size. From [Rust docs](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html#derivable):

> When `derived` on structs, two instances are equal if all fields are equal, and not equal if any fields are not equal. When `derived` on enums, each variant is equal to itself and not equal to the other variants.

This implies that the implementation for `PartialEq` is a simple recursive equality check (for structs) and case match (for enums), which should be a fairly simple function. The [`Eq` trait](https://doc.rust-lang.org/std/cmp/trait.Eq.html#derivable) doesn't add any additional methods, it just tells Rust that the derived methods in `PartialEq` are actually full equivalence relations.

Informally, this [thread on Reddit](https://www.reddit.com/r/rust/comments/p5vsjl/is_there_any_reason_not_to_derive_every_possible/) seems to indicate the same, e.g. one user says:

> If you're exposing a struct that mostly represents a value-based bag of data, then you should be implementing/deriving Debug, Clone, PartialEq, Hash at a minimum. If you can derive Eq, you should do that, too.

aferr: OK it sounds like it should be good to use this, then. Thanks for looking into it and for a thorough answer.

PR #247, opened 2021-12-06
Draft PR showing issues with the "may have integrity tag" approach.

markww:
This draft PR shows the issues with the proposed integrity tag
propagation approach using `mayHaveIntegrityTag`, in which an
`AccessPath` must have an `IntegrityTag` if it `mayHaveIntegrityTag`
(where `mayHaveIntegrityTag` facts propagate like confidentiality tags)
and does not have a `lacksIntegrityTag` fact. Policy rules propagate
a `may` fact. In attempting to write a simple rule of this form, `lacks`
facts propagate over the newly-introduced `may` facts, making them
ineffective.

### DIFF
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -122,7 +122,7 @@ mayHaveIntegrityTag(ap, owner, integTag) :- hasAppliedIntegrityTag(ap, owner, in
 
 // Any AccessPath that has an incoming resolvedEdge where the source of that
 // edge may have an IntegrityTag may also have that IntegrityTag.
-mayHaveIntegrityTag(src, owner, integTag) :-
+mayHaveIntegrityTag(dst, owner, integTag) :-
   resolvedEdge(owner, src, dst), mayHaveIntegrityTag(src, owner, integTag),
   !removeIntegrityTag(dst, owner, integTag).
 
--- a/src/analysis/souffle/taint.dl
+++ b/src/analysis/souffle/taint.dl
@@ -64,9 +64,12 @@
 // Removes an integrity tag from some path.
 .decl removeIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
 
-// Indicates that a particular tag may lack an integrity tag, which
+// Indicates that a particular AccessPath may lack an integrity tag, which
 // indicates that it should not have an integrity tag.
 .decl lacksIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
+
+// Indicates that a particular AccessPath may have an integrity tag.
+.decl mayHaveIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
 
 // Used to test whether a particular path has a particular integrity tag.
 // Should be used as close to top-level checks as possible, and not for
@@ -113,10 +116,21 @@ lacksIntegrityTag(dst, owner, integTag) :-
   edge(src, dst), lacksIntegrityTag(src, owner, integTag),
   !hasAppliedIntegrityTag(dst, owner, integTag).
 
-// An access path must have an integrity tag if it does not lack the integrity tag.
+// Any AccessPath that has an IntegrityTag directly applied to it may (really,
+// stronger than may, must) have that IntegrityTag.
+mayHaveIntegrityTag(ap, owner, integTag) :- hasAppliedIntegrityTag(ap, owner, integTag).
+
+// Any AccessPath that has an incoming resolvedEdge where the source of that
+// edge may have an IntegrityTag may also have that IntegrityTag.
+mayHaveIntegrityTag(src, owner, integTag) :-
+  resolvedEdge(owner, src, dst), mayHaveIntegrityTag(src, owner, integTag),
+  !removeIntegrityTag(dst, owner, integTag).
+
+// An access path must have an integrity tag if it may have the IntegrityTag
+// and does not lack the IntegrityTag
 mustHaveIntegrityTag(ap, owner, integTag) :-
   isAccessPath(ap), isIntegrityTag(integTag), isPrincipal(owner),
-  !lacksIntegrityTag(ap, owner, integTag).
+  mayHaveIntegrityTag(ap, owner, integTag), !lacksIntegrityTag(ap, owner, integTag).
 
 //-----------------------------------------------------------------------------
 // Universe Populating Rules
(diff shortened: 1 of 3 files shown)


src/analysis/souffle/taint.dl
```diff
@@ -113,10 +116,21 @@ lacksIntegrityTag(dst, owner, integTag) :-
...
   !hasAppliedIntegrityTag(dst, owner, integTag).
 
-// An access path must have an integrity tag if it does not lack the integrity tag.
+// Any AccessPath that has an IntegrityTag directly applied to it may (really,
+// stronger than may, must) have that IntegrityTag.
+mayHaveIntegrityTag(ap, owner, integTag) :- hasAppliedIntegrityTag(ap, owner, integTag).
+
+// Any AccessPath that has an incoming resolvedEdge where the source of that
+// edge may have an IntegrityTag may also have that IntegrityTag.
+mayHaveIntegrityTag(src, owner, integTag) :-
```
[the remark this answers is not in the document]
markww: That was very silly of me. I have fixed this. It caused some other test failures in this PR to go away, but the results are the same on the test introduced in this PR.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -0,0 +1,71 @@
...
+edge(dividend, result), edge(divisor, result) :- div_op(result, dividend, divisor).
+
+isAccessPath("a").
+isAccessPath("b").
+isAccessPath("c").
+isAccessPath("d").
+isAccessPath("e").
+
+mult_op("c", "a", "b").
+div_op("e", "c", "d").
```
bgogul, line 32: Should this be the other way around?
```suggestion
div_op("c", "a", "b").
mult_op("e", "c", "d").
```

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -0,0 +1,71 @@
...
+// lacksIntegrityTag(ap, owner, "TimestampSeconds") :-
+//  div_op(ap, operand1, operand2),
+//  (lacksIntegrityTag(operand1, owner, "TimestampMillis");
+//   lacksIntegrityTag(operand2, owner, "Constant_1000")).
+
+// This is the second policy rule.
+mayHaveIntegrityTag(ap, owner, "TimestampMillisRedacted") :-
+  mult_op(ap, operand1, operand2),
+  mayHaveIntegrityTag(operand1, owner, "TimestampSeconds"),
+  mayHaveIntegrityTag(operand1, owner, "Constant_1000").
```
bgogul, line 55: In correct operand?
```suggestion
  mayHaveIntegrityTag(operand2, owner, "Constant_1000").
```

src/analysis/souffle/taint.dl
```diff
@@ -113,10 +116,21 @@ lacksIntegrityTag(dst, owner, integTag) :-
...
+mayHaveIntegrityTag(dst, owner, integTag) :-
+  resolvedEdge(owner, src, dst), mayHaveIntegrityTag(src, owner, integTag),
+  !removeIntegrityTag(dst, owner, integTag).
+
+// An access path must have an integrity tag if it may have the IntegrityTag
+// and does not lack the IntegrityTag
 mustHaveIntegrityTag(ap, owner, integTag) :-
   isAccessPath(ap), isIntegrityTag(integTag), isPrincipal(owner),
-  !lacksIntegrityTag(ap, owner, integTag).
+  mayHaveIntegrityTag(ap, owner, integTag), !lacksIntegrityTag(ap, owner, integTag).
```
bgogul, line 133: In this rule `isIntegrityTag` is only true for the seed tags: `TimestampMillis` and `Constant_1000`.  Specifically, `isIntegrityTag` is not true for `TimestampSeconds`, and `TimestampRedacted`.

bgogul (commented):
Found a couple of minor bugs. Fixing this required some non-trivial changes. Commit https://github.com/google-research/raksha/pull/248/commits/d759d53f29aa56b796f185e5983fece622420500 in #248 has the proposed changes. PTAL.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
markww: Fixed.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
markww: Fixed.

src/analysis/souffle/taint.dl
markww: Good point. I've added explicit `isIntegrityTag` facts for the `IntegrityTag`s added by policy rules.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -0,0 +1,71 @@
...
+// While we might use something like operators in the future, we are currently
+// doing something a bit more ad-hoc for this test by indicating that certain
+// access paths are the result of operation expressions with the isOperatorExpr
+// relation.
+
+.decl mult_op(result: AccessPath, operand1: AccessPath, operand2: AccessPath)
+.decl div_op(result: AccessPath, dividend: AccessPath, divisor: AccessPath)
+
+edge(op1, result), edge(op2, result) :- mult_op(result, op1, op2).
+edge(dividend, result), edge(divisor, result) :- div_op(result, dividend, divisor).
```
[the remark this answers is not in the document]
markww: Those rules handle the propagation via policy rules. However, without a deeper refactoring, automatic `IntegrityTag` and `ConfidentialityTag` propagation occurs only via `edge`s. Unless we wished to change the semantics of `edge`s to be not just "data flows between these two `AccessPath`s" to "an assignment occurs between these two `AccessPath`s (as we did in the `phi` node PR), we need this to preserve default propagation.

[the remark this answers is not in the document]
bgogul: Isn't this similar to the handling of `derives from` claims? Recall that we remove some intra-particle edges between inputs and outputs, when there are `derives from` claims because we should not do default propagation along some edges. Similarly, if we understand the semantics of some operation, we should use more specific propagation rules that are aligned with the semantics. no?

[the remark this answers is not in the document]
markww: Sure, I think that makes sense. In that case, getting rid of these `edge` rules would be the right thing.

markww:
We are abandoning this approach in favor of #249. Closing.

PR #248, opened 2021-12-07
Possible fixes for the  with integrity tags propagated with lacks.

bgogul:


src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -16,48 +16,100 @@
...
+// Default propagation rules.
+lacksIntegrityTag(ap, owner, tag) :-
+  div_op(ap, operand1, _),
+  lacksIntegrityTag(operand1, owner, tag),
+  tag != "TimestampMillis".
+
+lacksIntegrityTag(ap, owner, tag) :-
+  div_op(ap, _, operand2),
+  lacksIntegrityTag(operand2, owner, tag),
+  tag != "Constant_1000".
```
markww, line 78: I'm not sure this policy rule makes sense. The result of `/` is not `1000` in most circumstances, and so the lack of `Constant_1000` should propagate most of the time.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -16,48 +16,100 @@
...
+lacksIntegrityTag(ap, owner, "TimestampSeconds") :-
+   div_op(ap, operand1, operand2),
+   (lacksIntegrityTag(operand1, owner, "TimestampMillis");
+    lacksIntegrityTag(operand2, owner, "Constant_1000")).
+
+// Default propagation rules.
+lacksIntegrityTag(ap, owner, tag) :-
+  div_op(ap, operand1, _),
+  lacksIntegrityTag(operand1, owner, tag),
+  tag != "TimestampMillis".
```
markww, line 72: This policy rule is suspect. We do expect that dividing an operand with the tag `TimestampMillis` will produce something that isn't a valid `TimestampMillis` most of the time.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -16,48 +16,100 @@
...
+// Default propagation rules.
+lacksIntegrityTag(ap, owner, tag) :-
+  mult_op(ap, operand1, _),
+  lacksIntegrityTag(operand1, owner, tag),
+  tag != "TimestampSeconds".
+
+lacksIntegrityTag(ap, owner, tag) :-
+  mult_op(ap, _, operand2),
+  lacksIntegrityTag(operand2, owner, tag),
+  tag != "Constant_1000".
```
markww, line 112: Also suspicious of these default propagation rules excluding `Constant_1000` as above.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
```diff
@@ -16,48 +16,100 @@
...
 // These rules propagating lacks of integrity tags across rules can only remove
 // integrity tags. Remove them for now to illustrate issue.
-// lacksIntegrityTag(ap, owner, "TimestampSeconds") :-
-//  div_op(ap, operand1, operand2),
-//  (lacksIntegrityTag(operand1, owner, "TimestampMillis");
-//   lacksIntegrityTag(operand2, owner, "Constant_1000")).
+lacksIntegrityTag(ap, owner, "TimestampSeconds") :-
+   div_op(ap, operand1, operand2),
+   (lacksIntegrityTag(operand1, owner, "TimestampMillis");
+    lacksIntegrityTag(operand2, owner, "Constant_1000")).
```
markww, line 66: I wonder about this propagation of `lacks` across the `PolicyRule` in the case where it does not fire. Imagine that we are creating an analysis that tracks, as in `IntegrityTag`, which integers have `DefinitelyOverflowed`. By default, we believe that any arithmetic operation where both inputs have `DefinitelyOverflowed` has, itself, `DefinitelyOverflowed`. Further, we have a policy rule that applies `DefinitelyOverflowed` to the result of any addition where both operands had the `IntegrityTag` `VeryLargeInteger`. Imagine that that policy rule was automatically converted to the following datalog:

```
mayHaveIntegrityTag(ap, owner, "DefinitelyOverflowed") :-
  mayHaveIntegrityTag(op1, owner, "VeryLargeInteger"),
  mayHaveIntegrityTag(op2, owner, "VeryLargeInteger"),
  add_op(ap, op1, op2).
  
lacksIntegrityTag(ap, owner, "DefinitelyOverflowed") :-
  lacksHaveIntegrityTag(op1, owner, "VeryLargeInteger"),
  lacksHaveIntegrityTag(op2, owner, "VeryLargeInteger"),
  add_op(ap, op1, op2).
```

Now imagine that a `+` operator receives two inputs, each of which is tagged with `DefinitelyOverflowed`. By the default propagation rule, the result should also have `DefinitelyOverflowed`. However, the policy rule indicates that in the case where either of the operands does not have `VeryLargeInteger` on it, the result shall lack the integrity tag `DefinitelyOverflowed`. Therefore, the policy rule will end up *clearing* the `IntegrityTag`!

If the above seems a little hackneyed because the overflow analysis is weird (and I admit, it is weird), imagine that two different policy rules apply the same `IntegrityTag` to the same target `AccessPath` with different `IntegrityTag` preconditions. With the current `lacks` propagation, the preconditions of *both* policy rules must be met for the `IntegrityTag` to be applied rather than the intended behavior of *either* policy rule being able to apply the `IntegrityTag`.

src/analysis/souffle/taint.dl
```diff
@@ -99,12 +99,18 @@ mayHaveTag(tgt, owner, tag) :-
...
+
 // Any access point that has no incoming edges and is not directly given an
 // integrity tag lacks that integrity tag, as there is no way for it to
 // propagate to that AccessPath.
 lacksIntegrityTag(ap, owner, integTag) :-
   isAccessPath(ap), isPrincipal(owner), isIntegrityTag(integTag),
-  !edge(_, ap), !hasAppliedIntegrityTag(ap, owner, integTag).
+  // We will need to capture the fact that some propagation happens because of
+  // operators.
+  !edge(_, ap), !mult_op(ap, _, _), !div_op(ap, _, _),
```
markww, line 112: In a real system, we'd probably make all of these ops put entries in an `operator` relation and check that one relation rather than each of these specifically. But I get that this is an example PR for the sake of discussion, just flagging this issue in case it becomes the basis for a real PR.

markww:
Just left a review. @bgogul , you have sidestepped the original issue that I brought up with how you have structured default propagation. But I think that there's another issue, which is that I think that propagating `lacks` across a policy rule in the way currently described will cause policy rules to interfere with each other and possibly unintentionally clear `IntegrityTag`s in some cases.

src/analysis/souffle/tests/arcs_fact_tests/dynamic_rule_test.dl
bgogul: The rules might be buggy, but let me try to explain what I was trying. To set context, the approach I am trying is to *group* all the rules for `mayHaveIntegrityTag` and `lacksIntegrity` of a specific  tag + operator combination into a single rule. (This would be analogous to implementing the transfer functions that capture the semantics of an operator during abstract interpretation.)

Given that context, what you want to capture is `ap` is `DefinitelyOverflowed` if _any_ of following conditions hold in a _single_ rule:
   - Both `op1` and `op2` are `DefinitelyOverflowed`, or
   - `op1` is `VeryLargeInteger`, or
   - `op2` is `VeryLargeInteger`

In other words, `ap`  has `DefinitelyOverflowed` if the following holds (bear with `!` for a moment):
```
   (!lacksIntegrity(op1, "DefinitelyOverflowed"), !lacksIntegrity(op2, "DefinitelyOverflowed));
   !lacksIntegrity(op1, "VeryLargeInteger");
   !lacksIntegrity(op2, "VeryLargeInteger") 
```
You would capture `lacksIntegrity` with the negation of above, which would translate to the following:
```
lacksIntegrityTag(ap, owner, "DefinitelyOverflowed") :- 
  (lacksHaveIntegrityTag(op1, owner, "DefinitelyOverflowed");
   lacksHaveIntegrityTag(op2, owner, "DefinitelyOverflowed")), 
   lacksIntegrity(op1, "VeryLargeInteger"), 
   lacksIntegrity(op2, "VeryLargeInteger").
```

That said, I agree that it is not quite straightforward to generate these rules. It can easily lead to loads of bugs.  It is not quite readable also. I am beginning to wonder if it is worth all the trouble.

bgogul (commented):
> Just left a review. @bgogul , you have sidestepped the original issue that I brought up with how you have structured default propagation.

@markww which issue are you referring to?

>  But I think that there's another issue, which is that I think that propagating lacks across a policy rule in the

I left comments/reply to your question.

markww:
> > Just left a review. @bgogul , you have sidestepped the original issue that I brought up with how you have structured default propagation.
> 
> @markww which issue are you referring to?

I may have worded that confusingly. I meant to say "you fixed it such that the original test that I wrote works".

PR #249, opened 2021-12-08
Use predecessor ordering to propagate IntegrityTags without negation.

markww:
This commit propagates `IntegrityTag`s without altering the dataflow
graph, without restricting the tags that may be introduced
dynamically, and without using negation. It does so by using a different
formulation of `forall` for the `IntegrityTag` propagation rule. Rather
than using `!`, it assigns each predecessor `AccessPath` a predecessor
order number. Then, it calculates a relation
`predecessorsUpToOrderNumHaveIntegrityTag` inductively, using the 0th
predecessor as the base case. Then, an `AccessPath` must have an
integrity tag if `predecessorsUpToOrderNumHaveIntegrityTag` for an
`orderNum` that is one less than the number of total predecessors.

### DIFF
--- a/src/analysis/souffle/dataflow_graph.dl
+++ b/src/analysis/souffle/dataflow_graph.dl
@@ -52,7 +52,7 @@
 // to the tgt. That allows us to effectively eliminate edges that are the
 // subject of a claimNotEdge from the dataflow graph by eliminating all tags
 // from the midway point. Paths are defined in terms of resolvedEdges.
-.decl resolvedEdge(for: Principal, src: AccessPath, tgt: AccessPath)
+.decl resolvedEdge(owner: Principal, src: AccessPath, tgt: AccessPath)
 
 // A direct or transitive data flow path.
 .decl path(src: AccessPath, tgt: AccessPath)
@@ -69,8 +69,8 @@
 // predecessor have the IntegrityTag and all predecessors up to N-1 have the
 // IntegrityTag". We can then ask if all predecessors up to numPredecessors
 // - 1 have the integrity tag to accomplish forall.
-.decl orderedPredecessors(ap: AccessPath, for: Principal, src: AccessPath, orderNum: number)
-.decl numPredecessors(ap: AccessPath, for: Principal, num: number)
+.decl orderedPredecessors(ap: AccessPath, owner: Principal, src: AccessPath, orderNum: number)
+.decl numPredecessors(ap: AccessPath, owner: Principal, num: number)
 
 //-----------------------------------------------------------------------------
 // Rules
@@ -91,19 +91,19 @@ isAccessPath(y) :- resolvedEdge(_, _, y).
 
 // The 0th predecessor shall be the predecessor with the smallest src ordinal
 // number.
-orderedPredecessors(ap, for, src, 0) :-
-  resolvedEdge(for, src, ap),
-  ord(src) = min ord(candidate) : { resolvedEdge(for, candidate, ap) }.
+orderedPredecessors(ap, owner, src, 0) :-
+  resolvedEdge(owner, src, ap),
+  ord(src) = min ord(candidate) : { resolvedEdge(owner, candidate, ap) }.
 
 // The Nth precessor shall be the precessor with the smallest src ordinal
 // number larger than the N-1th ordinal number.
-orderedPredecessors(ap, for, src, prevOrd + 1) :-
-  orderedPredecessors(ap, for, prevSrc, prevOrd),
-  resolvedEdge(for, src, ap),
-  ord(src) = min ord(candidate) : { resolvedEdge(for, candidate, ap), ord(candidate) > ord(prevSrc) }.
+orderedPredecessors(ap, owner, src, prevOrd + 1) :-
+  orderedPredecessors(ap, owner, prevSrc, prevOrd),
+  resolvedEdge(owner, src, ap),
+  ord(src) = min ord(candidate) : { resolvedEdge(owner, candidate, ap), ord(candidate) > ord(prevSrc) }.
 
-numPredecessors(ap, for, maxOrd + 1) :-
-  orderedPredecessors(ap, for, _, maxOrd),
-  maxOrd = max ordNum : { orderedPredecessors(ap, for, _, ordNum) }.
+numPredecessors(ap, owner, maxOrd + 1) :-
+  orderedPredecessors(ap, owner, _, maxOrd),
+  maxOrd = max ordNum : { orderedPredecessors(ap, owner, _, ordNum) }.
 
 #endif // SRC_ANALYSIS_SOUFFLE_DATAFLOW_GRAPH_DL_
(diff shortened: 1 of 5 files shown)


src/analysis/souffle/dataflow_graph.dl
```diff
@@ -73,4 +89,21 @@ path(from, to) :- resolvedEdge(_, from, intermediate), path(intermediate, to).
 isAccessPath(x) :- resolvedEdge(_, x, _).
 isAccessPath(y) :- resolvedEdge(_, _, y).
 
+// The 0th predecessor shall be the predecessor with the smallest src ordinal
+// number.
+orderedPredecessors(ap, for, src, 0) :-
+  resolvedEdge(for, src, ap),
+  ord(src) = min ord(candidate) : { resolvedEdge(for, candidate, ap) }.
```
bgogul, line 96: Is `ord` a souffle primitive?

src/analysis/souffle/dataflow_graph.dl
```diff
@@ -73,4 +89,21 @@ path(from, to) :- resolvedEdge(_, from, intermediate), path(intermediate, to).
 isAccessPath(x) :- resolvedEdge(_, x, _).
 isAccessPath(y) :- resolvedEdge(_, _, y).
 
+// The 0th predecessor shall be the predecessor with the smallest src ordinal
+// number.
+orderedPredecessors(ap, for, src, 0) :-
```
bgogul, line 94: `for` is a bit confusing. Can we `owner` or some other name?

src/analysis/souffle/taint.dl
```diff
@@ -64,9 +64,11 @@
 // Removes an integrity tag from some path.
 .decl removeIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
 
-// Indicates that a particular tag may lack an integrity tag, which
-// indicates that it should not have an integrity tag.
-.decl lacksIntegrityTag(accessPath: AccessPath, owner: Principal, integTag: IntegrityTag)
+// Given how predecessors are ordered in orderedPredecessors, indicate whether
+// a predcessor with a given order num and all predecessors with lower order
+// nums have a particular integrity tag.
+.decl predecessorsUpToOrderNumHaveIntegrityTag(
```
bgogul, line 70: nit: `orderNum` as the first argument?

src/analysis/souffle/dataflow_graph.dl
```diff
@@ -73,4 +89,21 @@ path(from, to) :- resolvedEdge(_, from, intermediate), path(intermediate, to).
...
+  resolvedEdge(for, src, ap),
+  ord(src) = min ord(candidate) : { resolvedEdge(for, candidate, ap) }.
+
+// The Nth precessor shall be the precessor with the smallest src ordinal
+// number larger than the N-1th ordinal number.
+orderedPredecessors(ap, for, src, prevOrd + 1) :-
```
bgogul, line 100: Will cycles in the graph cause any issue? Do we have test to verify that this is OK?

bgogul (approved):
This certainly seems much simpler.

src/analysis/souffle/dataflow_graph.dl
markww: Yes. `ord` is the [ordinal number for the symbol](https://souffle-lang.github.io/types#symbol-type) that Souffle uses internally. The documentation used to have language that implied greater degrees about stability (ie, `ord` would be assigned to symbols in the order that Souffle saw them) but now basically says that `ord` will be an arbitrary unique mapping between symbols and numbers.

It's less stable than I'd like and very Souffle-specific. In the future, if we find that it matters, we can have a more stable mapping from `AccessPath`s to unique IDs by just having the IR translator output a unique ID for every `AccessPath` produced. If we switch to representing `AccessPath`s with ADTs, we will need to do that, as Souffle only produces `ord` numbers for symbols. But I think `ord` works for now.

src/analysis/souffle/dataflow_graph.dl
markww: Yep, will do.

src/analysis/souffle/taint.dl
markww: Sure, will do.

src/analysis/souffle/dataflow_graph.dl
markww: We do not have a test to verify that, I'll add one to make sure.

src/analysis/souffle/taint.dl
markww: Done.

src/analysis/souffle/dataflow_graph.dl
markww: Done.

src/analysis/souffle/dataflow_graph.dl
markww: Done.

src/analysis/souffle/dataflow_graph.dl
```diff
@@ -73,4 +89,21 @@ path(from, to) :- resolvedEdge(_, from, intermediate), path(intermediate, to).
 isAccessPath(x) :- resolvedEdge(_, x, _).
 isAccessPath(y) :- resolvedEdge(_, _, y).
 
+// The 0th predecessor shall be the predecessor with the smallest src ordinal
+// number.
```
aferr, line 93: It took me a moment to figure out that the `ord` is not a user defined thing and that it's the built-in described [here](https://souffle-lang.github.io/types) (I saw this in the souffle docs forever ago but forgot it existed). Maybe it's useful to mention that this is a built-in functor and give the link.

src/analysis/souffle/dataflow_graph.dl
```diff
@@ -52,14 +52,30 @@
...
+// each predecessor an arbitrary predecessor number between 0 and N, where N is
+// the number of predecessors a particular AccessPath has. Track the number of
+// predecessors an AccessPath has in numPredecssors. This allows us to
+// calculate whether all predecessors up to a particular order number have an
+// integrity tag, with the base case of 0 being just "does that particular
+// AccessPath have the IntegrityTag" and the inductive case being "does the nth
+// predecessor have the IntegrityTag and all predecessors up to N-1 have the
+// IntegrityTag". We can then ask if all predecessors up to numPredecessors
+// - 1 have the integrity tag to accomplish forall.
+.decl orderedPredecessors(ap: AccessPath, owner: Principal, src: AccessPath, orderNum: number)
```
aferr, line 72: This is just a nit/preference so it's not strictly necessary. I'd find "dst" more intuitive than "ap" since it conveys the direction of the edge. You already have src as the other argument, so its possible to figure it out still. Similar thought for other declarations and definitions here.

aferr (commented):
This answer looks really good! It seems extremely intuitive, I think it pretty clearly implements just the expected functionality for integrity, and it also give us another way to do quantification that might work for things other than just integrity.

The only potential "drawback" is that it is relatively souffle specific because it relies on the ordinal thing, so it might be hard to redo this in a custom solver. But that's a pretty speculative and unimportant-for-now "drawback", so this looks great.

PR #250, opened 2021-12-08
Fixes to get C++ Raksha working locally after a toolchain change.

markww:
Some piece(s) of my toolchain got updated in the night and broke my
build. These changes seem get the build fixed, at least for the C++ side
which is in my inner loop. The Rust side is still broken for me, but
I don't usually touch that code, so that's less critical for me to fix.


.bazelrc
```diff
@@ -6,5 +6,4 @@ build --cxxopt='-std=c++17'
 build --cxxopt='-Werror' --cxxopt='-Wall'
 build --cxxopt='-Wno-deprecated-declarations'
 build --host_cxxopt='-std=c++17'
-build --host_cxxopt='-Werror' --host_cxxopt='-Wall'
```
bgogul, line 9: Generally, we have been keeping `cxxopt` and `host_cxxopt` consistent. 

Can you add a comment or explanation as to why this is removed? I believe this is causing issues with rust?

markww: Yes. Just to explain here as well: it appears that applying `--host_cxxopt='-Werror'` causes dependencies that we compile from source (like protobuf) to be built with `-Werror`. This breaks our build if those dependencies are not warning-clean. It appears that getting rid of `-Werror` in `--host_cxxopt` but keeping it in `--cxxopt` addresses this issue in the dependencies, but keeps the compiler erroring on warnings in our own source.

bgogul (approved):

.bazelrc
markww: Added comment.

PR #252, opened 2021-12-09
Using cargo-chef to cache rust dependencies

aferr:
This PR caches the sources of the rust crate dependencies in the docker image using [cargo-chef](https://www.lpalmieri.com/posts/fast-rust-docker-builds/). I was hoping to cache not just the sources of the dependencies but also the built binaries, which I thought should be possible. However, it looks like this already has a pretty substantial performance improvement -- (e.g., [this build](https://pantheon.corp.google.com/cloud-build/builds;region=global/51618f64-cd29-41f6-9406-c2477da5e5b0;step=4?project=google.com:raksha-ci) without cacheing sources took ~8 minutes compared to ~3.5 minutes with cacheing in [this build](https://pantheon.corp.google.com/cloud-build/builds;region=global/2e7412c3-ea3e-4a60-80de-dab0cadc6c7b;step=4?project=google.com:raksha-ci). So I thought it would be useful to clean this up and get it in anyway.

You can also see that it really is cacheing fetching of the sources by looking at [this build](https://pantheon.corp.google.com/cloud-build/builds;region=global/a5729baf-488e-46be-a72d-f0cdde89ef0a;step=1?project=google.com:raksha-ci). In step 1 cargo-chef is used to fetch the dependencies, and to do so it uses the cache. In step 4 when authorization logic tests are done, it no longer says that it is downloading the dependencies, just compiling them.

I am not sure why recompiling the binaries is not skipped as of this PR. It looks like the crate binaries are properly copied from the directory where they are stored in the docker image into `rust/tools/authorization-logic/target` (and the `debug` and `debug/deps` subdirs also look right).


aferr:
In the build that was triggered automatically for this PR it doesn't use cached sources for the dependencies because they weren't cached yet. I re-ran the job manually from the gcloud CLI and in this one it does use the cached sources: https://pantheon.corp.google.com/cloud-build/builds;region=global/c1cab45b-3c58-409c-b0b3-d76ff5327a0c;step=3?project=google.com:raksha-ci

Dockerfile
```diff
@@ -49,3 +49,25 @@ RUN rustup toolchain install ${RUST_VERSION} \
...
+
+# Install chef.
+# It might also be possible to leverage a docker image
+# that already has chef with the commented out ocmmand
+# FROM lukemathwalker/cargo-chef:latest-rust-1.53.0 as chef
+RUN cargo install cargo-chef
+
+# Copy just the auth logic files (so we can build the deps)
+ARG AUTH_LOGIC_DIR=rust/tools/authorization-logic
+COPY ./${AUTH_LOGIC_DIR} /chef_tmp/${AUTH_LOGIC_DIR}
```
bgogul, line 66: We still seem to recompile the files. I wonder if you should do `COPY ./${AUTH_LOGIC_DIR} .` and run `cargo chef` from `${AUTH_LOGIC_DIR}`?

bgogul (approved):
This still seems to recompile the files. I left a suggestion to try. 

In any case, it should be OK to get this in and iterate.

Dockerfile
aferr: Didn't try this yet, but I kept a note to try this later to see if I can get it to go through. Thank you :)

PR #253, opened 2021-12-09
Implement operations and default tag propagation across operations.

markww:
To allow in-Datalog AST matching, better propagation rules, and possibly
some future shared general abstract interpretation on abstract state,
implement operations. Operations are simple calculation expressions that
take in some number of inputs and return some outputs. `resolvedEdge`s
are considered to be a synonym for a unary assignment operation.

All operations of any arity share the same default propagation rule for
confidentiality tags: a confidentiality tag on any input shows up on the
result. No operation other than `assign` has a default propagation rule
for `IntegrityTag`s.

This commit further shows how dynamic rules can be written by users
and/or frontends to manually propagate `IntegrityTag`s or remove
confidentiality tags in a dataflow graph containing operations.

### DIFF
--- a/src/analysis/souffle/dataflow_graph.dl
+++ b/src/analysis/souffle/dataflow_graph.dl
@@ -52,14 +52,30 @@
 // to the tgt. That allows us to effectively eliminate edges that are the
 // subject of a claimNotEdge from the dataflow graph by eliminating all tags
 // from the midway point. Paths are defined in terms of resolvedEdges.
-.decl resolvedEdge(for: Principal, src: AccessPath, tgt: AccessPath)
+.decl resolvedEdge(owner: Principal, src: AccessPath, tgt: AccessPath)
 
 // A direct or transitive data flow path.
 .decl path(src: AccessPath, tgt: AccessPath)
 
+// A common way to do "forall" in Datalog is to use !exists. Unfortunately,
+// this imposes a stratification requirement. To allow IntegrityTags to
+// propagate when all predecessors have a particular IntegrityTag, instead give
+// each predecessor an arbitrary predecessor number between 0 and N, where N is
+// the number of predecessors a particular AccessPath has. Track the number of
+// predecessors an AccessPath has in numPredecssors. This allows us to
+// calculate whether all predecessors up to a particular order number have an
+// integrity tag, with the base case of 0 being just "does that particular
+// AccessPath have the IntegrityTag" and the inductive case being "does the nth
+// predecessor have the IntegrityTag and all predecessors up to N-1 have the
+// IntegrityTag". We can then ask if all predecessors up to numPredecessors
+// - 1 have the integrity tag to accomplish forall.
+.decl orderedPredecessors(ap: AccessPath, owner: Principal, src: AccessPath, orderNum: number)
+.decl numPredecessors(ap: AccessPath, owner: Principal, num: number)
+
 //-----------------------------------------------------------------------------
 // Rules
 //-----------------------------------------------------------------------------
+
 resolvedEdge(principal, src, tgt) :-
    isPrincipal(principal),
    edge(src, tgt),
@@ -72,5 +88,22 @@ path(from, to) :- resolvedEdge(_, from, intermediate), path(intermediate, to).
 // Symbols used in resolvedEdges are access paths
 isAccessPath(x) :- resolvedEdge(_, x, _).
 isAccessPath(y) :- resolvedEdge(_, _, y).
+
+// The 0th predecessor shall be the predecessor with the smallest src ordinal
+// number.
+orderedPredecessors(ap, owner, src, 0) :-
+  resolvedEdge(owner, src, ap),
+  ord(src) = min ord(candidate) : { resolvedEdge(owner, candidate, ap) }.
+
+// The Nth precessor shall be the precessor with the smallest src ordinal
+// number larger than the N-1th ordinal number.
+orderedPredecessors(ap, owner, src, prevOrd + 1) :-
+  orderedPredecessors(ap, owner, prevSrc, prevOrd),
+  resolvedEdge(owner, src, ap),
+  ord(src) = min ord(candidate) : { resolvedEdge(owner, candidate, ap), ord(candidate) > ord(prevSrc) }.
+
+numPredecessors(ap, owner, maxOrd + 1) :-
+  orderedPredecessors(ap, owner, _, maxOrd),
+  maxOrd = max ordNum : { orderedPredecessors(ap, owner, _, ordNum) }.
 
 #endif // SRC_ANALYSIS_SOUFFLE_DATAFLOW_GRAPH_DL_
--- a/src/analysis/souffle/dataflow_graph.dl
+++ b/src/analysis/souffle/dataflow_graph.dl
@@ -77,25 +77,6 @@
 // Rules
 //-----------------------------------------------------------------------------
 
-// Cross-populate operation_1 and operation.
-operation(owner, operator, result, [input1, nil]) :- operation_1(owner, operator, result, input1).
-operation_1(owner, operator, result, input1) :- operation(owner, operator, result, [input1, nil]).
-
-// Cross-populate operation_2 and operation.
-operation(owner, operator, result, [input1, [input2, nil]]) :-
-  operation_2(owner, operator, result, input1, input2). 
-operation_2(owner, operator, result, input1, input2) :-
-  operation(owner, operator, result, [input1, [input2, nil]]).
-
-// assign is a '=' operation with a single operand.
-operation_1(owner, "=", result, input1) :- assign(owner, result, input1).
-assign(owner, result, input1) :- operation_1(owner, "=", result, input1).
-
-// resolvedEdges are the same thing as assignments. Cross populate them to each
-// other.
-assign(owner, result, input1) :- resolvedEdge(owner, input1, result).
-resolvedEdge(owner, src, dst) :- assign(owner, dst, src).
-
 resolvedEdge(principal, src, tgt) :-
    isPrincipal(principal),
    edge(src, tgt),
--- a/src/analysis/souffle/dataflow_graph.dl
+++ b/src/analysis/souffle/dataflow_graph.dl
@@ -110,17 +110,18 @@ isAccessPath(x) :- resolvedEdge(_, x, _).
 isAccessPath(y) :- resolvedEdge(_, _, y).
 
 // The 0th predecessor shall be the predecessor with the smallest src ordinal
-// number.
-orderedPredecessors(ap, owner, src, 0) :-
-  resolvedEdge(owner, src, ap),
-  ord(src) = min ord(candidate) : { resolvedEdge(owner, candidate, ap) }.
+// number. We get the ordinal number using Souffle's `ord` primitive (see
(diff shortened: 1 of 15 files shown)


src/analysis/souffle/dataflow_graph.dl
```diff
@@ -76,6 +77,25 @@
 // Rules
 //-----------------------------------------------------------------------------
 
+// Cross-populate operation_1 and operation.
+operation(owner, operator, result, [input1, nil]) :- operation_1(owner, operator, result, input1).
```
bgogul, line 81: If we can have operation as a list, why do we need `operation_x` variants?

src/analysis/souffle/taint.dl
```diff
@@ -20,11 +20,19 @@
...
 //-----------------------------------------------------------------------------
 // Relation Declarations
 //-----------------------------------------------------------------------------
 
+// This fact indicates that, in some unspecified way, accessPath is influenced
+// by influencer. Unlike `resolvedEdge`, which specifically represents an exact
+// assignment, this can represent any dataflow from any operation. Should not
+// be populated directly, is populated indirectly by edges and other
+// operations.
+.decl influencedBy(owner: Principal, accessPath: AccessPath, influencer: AccessPath)
```
bgogul, line 34: `InfluencedBy` is not a standard term and this is more like controlDependence. Can we say `controlDependentOn`, `controlDependenceEdge`,  or something like that?

Actually based on the rule below, it looks like this more like dataflow or direct data dependence. If so, why can't we use `edge` itself?

src/analysis/souffle/taint.dl
```diff
@@ -77,6 +85,10 @@
 // Rules
 //-----------------------------------------------------------------------------
 
+// A `tgt` is influenced by any `src` that is an operand to an operation where
+// the result is placed in `tgt`.
+influencedBy(owner, tgt, src) :- operationToOperands(owner, _, tgt, _, src).
```
bgogul, line 90: No, this is direct dependence, right?

bgogul (commented):
Haven't looked at the examples/test in detail. I had some questions about the changes in datalog. PTAL.

src/analysis/souffle/dataflow_graph.dl
markww: It's syntactic sugar to make writing tests and frontends easier. While `operation`s can, in general, have an arbitrary number of operations, we know that, in practice, they most often have 1 or 2 operands. Having a special syntax for that common case can provide a degree of parse-time checking for silly errors (such as using `*` with a single operand, for instance).

src/analysis/souffle/taint.dl
markww: We were discussing changing the meaning of `edge` to mean "assignment", or an exact copy of one value from one location into another. This is attempting to capture the fact that taint may flow from the `influencer` to the target `accessPath`, and is a specific concept for confidentiality tag flow.

src/analysis/souffle/dataflow_graph.dl
bgogul: I see. Call them `binary_operation` and `unary_operation` then? Where `operation` used in this PR, btw?

src/analysis/souffle/taint.dl
bgogul: If there is no special semantics for an operator, we should just generate `edge` relations. 

Didn't we think of an edge from a special access path to capture the fact that an output value may be created without depending on any input values?

src/analysis/souffle/taint.dl
bgogul: This is related to another conversation thread. Let us continue the discussion there.

src/analysis/souffle/taint.dl
markww: Yes, that does make sense. I can refactor this PR to remove `influencedBy` and add a special `NoIntegrity` `AccessPath` to block integrity propagation.

markww: Done.

src/analysis/souffle/taint.dl
markww: Removed `influencedBy`.

src/analysis/souffle/dataflow_graph.dl
markww: Renaming done, now called `binaryOperation` and `unaryOperation`.

`operation` is used when populating the `operationToOperands` relation. This allows us to uniformly handle the default propagation rules and edge-drawing for any operation of any arity. See `arbitrary_operation_tag_propagation.dl`, where this is demonstrated for an `operation` with 4 inputs.

src/analysis/souffle/operations.dl
```diff
@@ -0,0 +1,80 @@
...
+  owner: Principal, operator: Operator, result: AccessPath, operandList: OperandList,
+  partialList: OperandList)
+
+operationToPartialOperandLists(owner, operator, result, operandList, operandList) :-
+  operation(owner, operator, result, operandList).
+operationToPartialOperandLists(owner, operator, result, operandList, tail) :-
+  operationToPartialOperandLists(owner, operator, result, operandList, [_, tail]).
+
+// Map full operations to operands.
+.decl operationToOperands(
```
bgogul, line 73: Should this just be `operationToOperand(owner, operator, result, operand)`?

bgogul (commented):
I had a question about ownership, but looks good otherwise.

src/analysis/souffle/operations.dl
```diff
@@ -0,0 +1,80 @@
...
+  operation(owner, operator, result, [input1, [input2, nil]]).
+
+// assign is a '=' operation with a single operand.
+unaryOperation(owner, "=", result, input1) :- assign(owner, result, input1).
+assign(owner, result, input1) :- unaryOperation(owner, "=", result, input1).
+
+// resolvedEdges are the same thing as assignments. Cross populate them to each
+// other.
+assign(owner, result, input1) :- resolvedEdge(owner, input1, result).
+resolvedEdge(owner, src, dst) :- assign(owner, dst, src).
```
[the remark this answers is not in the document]
markww: It does not. In essence, it just says "put the contents of the `resolvedEdge` relation into the `assign` operation and vice-versa until a fixed point is reached". Doing so will ensure that they have equivalent contents.

src/analysis/souffle/operations.dl
```diff
@@ -0,0 +1,80 @@
...
+operation(owner, operator, result, [input1, [input2, nil]]) :-
+  binaryOperation(owner, operator, result, input1, input2).
+binaryOperation(owner, operator, result, input1, input2) :-
+  operation(owner, operator, result, [input1, [input2, nil]]).
+
+// assign is a '=' operation with a single operand.
+unaryOperation(owner, "=", result, input1) :- assign(owner, result, input1).
+assign(owner, result, input1) :- unaryOperation(owner, "=", result, input1).
+
+// resolvedEdges are the same thing as assignments. Cross populate them to each
```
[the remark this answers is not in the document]
markww: That is quite possibly true. I can remove the piece that moves `resolvedEdge` into `assign`.

src/analysis/souffle/operations.dl
```diff
@@ -0,0 +1,80 @@
...
+// a single character operator (such as +, =, /, *, etc), the operator will be
+// the tsring containing that character.
+.type Operator <: symbol
+
+// A list of operands, used for gathering all operators in a single relation
+// regardless of arity. Useful for rules that apply to any operation.
+.type OperandList = [ operand: AccessPath, next: OperandList ]
+
+// Operations of any arity go in this relation.
+.decl operation(owner: Principal, operator: Operator, result: AccessPath, operandList: OperandList)
```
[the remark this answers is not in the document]
markww: My thinking was that if an `operation` is placed into the dataflow graph, it must have been the result of someone claiming that it should be there. This `owner` is that someone, in authorization logic `Principal` form. I think this would be especially useful if we decided that we used these `operation`s to show some partial internal dataflow of the `Particle`, as they would then be claims about the `Particle`'s behavior akin to `derives from`.

src/analysis/souffle/operations.dl
markww: My thinking was that I would lose information in the case that the same `result` was targeted by the result of two different `operation`s with the same `owner` and `operator` but different `operandList`s. That could cause ambiguity when showing a backtrace when we get around to implementing backtraces. That said, I'm using a wildcard everywhere I'm referring to this relation right now, so that might be just fine. Would be easy to add the `operandList` back in later if we needed it. I'll take `operandList` out for now.

src/analysis/souffle/operations.dl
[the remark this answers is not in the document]
markww: Done.

src/analysis/souffle/operations.dl
markww: Done.

src/analysis/souffle/taint.dl
```diff
@@ -77,6 +78,16 @@
 // Rules
 //-----------------------------------------------------------------------------
 
+// Draw an edge from any operation's operand to its result.
+resolvedEdge(owner, src, tgt) :- operationToOperands(owner, _, tgt, src).
```
bgogul, line 82: Shouldn't these edges be added only for those operations for which we don't understand how edges propagate. e.g., if we have a special interpretation for  "*" and if we  add these edges, won't we propagate tags imprecisely?

src/analysis/souffle/taint.dl
```diff
@@ -77,6 +78,16 @@
...
+
+// For all operations other than `=`, draw an edge from the special
+// "ArbitraryComputation" `AccessPath` to the result. This prevents default
+// propagation of `IntegrityTag`s through the operation.
+isAccessPath("ArbitraryComputation").
+resolvedEdge(owner, "ArbitraryComputation", tgt) :-
```
bgogul, line 88: I have a similar question as above.

bgogul (commented):
I still have a couple of questions. Let us chat in person and resolve this?

src/analysis/souffle/taint.dl
markww: We won't. There are two cases here: `ConfidentialityTag`s and `IntegrityTag`s:

* For the `ConfidentialityTag` case, imprecise overpropagation anywhere where dataflow is plausible is what we want, so this is fine.

* For the `IntegrityTag` case, we draw an additional edge from the `ArbitraryComputation` `AccessPath` to the result of any `operation` other than `=`. `ArbitraryComputation` has no incoming edges and has no `IntegrityTag`s. Because `IntegrityTag`s only propagate to an `AccessPath` when all incoming edges start at `AccessPath`s that have that `IntegrityTag`, default propagation of `IntegrityTag`s is blocked across all non-`=` operations.

src/analysis/souffle/taint.dl
markww: Discussed this in response in other thread, I think these two comment-points may look strange in isolation but they make sense when considered together.

bgogul (approved):

src/analysis/souffle/operations.dl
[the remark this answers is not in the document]
markww: As discussed on the phone, going with this for now, as it'd be easier to remove it later than add it back in later.

PR #254, opened 2021-12-10
Fix rust build errors and add build command to CI presubmit script.

bgogul:
This fixes the error related to not finding modules `test_multimic_no_overrides`  and `test_multimic_overrides`.

### DIFF
--- a/gcb/presubmit.sh
+++ b/gcb/presubmit.sh
@@ -66,7 +66,8 @@ bazel_run build \
   //third_party/arcs/examples:consume \
   //third_party/arcs/proto:manifest_cc_proto \
   //src/analysis/souffle/tests/arcs_manifest_tests_todo/... \
-  //src/analysis/souffle/examples/...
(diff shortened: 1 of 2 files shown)


markww (approved):
LGTM. So it wasn't just my machine, presubmit wasn't running the tests? That's a bit spooky. Maybe we should consider making it possible to run `bazel test //...` to prevent these sorts of issues in the future. Might be possible to do with judicious use of the `.bazelignore` file.

bgogul:
> LGTM. So it wasn't just my machine, presubmit wasn't running the tests? That's a bit spooky. Maybe we should consider making it possible to run `bazel test //...` to prevent these sorts of issues in the future. Might be possible to do with judicious use of the `.bazelignore` file.

Yeah, I forgot to add the rust build rule when setting up the presubmit on GCB. How will `bazel test //...` would be too permissive, I think.

PR #257, opened 2021-12-17
Create a header-only Souffle library, use instead of `souffle_lib`.

markww:
On chat, we had a discussion as to whether Souffle needs to be linked
into the C++ programs generated from compiled Souffle or whether they
are standalone programs. This commit is an experiment to clarify that:
in all cases possible, this replaces `souffle_lib` with
`souffle_include_lib`, a library consisting only of Souffle header
files. It appears that this can replace `souffle_lib` in all cases in
Raksha where we are compiling Datalog, meaning that Souffle-generated
C++ programs do not need to link with Souffle.

Note that we would need to link with a native binary if we were to use
Souffle user-defined functors. However, we have not needed to resort to
that yet.

### DIFF
--- a/build_defs/raksha.bzl
+++ b/build_defs/raksha.bzl
@@ -108,7 +108,7 @@ def policy_check(name, dataflow_graph, auth_logic, expect_failure = False, visib
         ],
         linkopts = ["-pthread"],
         deps = [
-            "@souffle//:souffle_lib",
+            "@souffle//:souffle_include_lib",
             souffle_dl_cpp_target,
         ],
     )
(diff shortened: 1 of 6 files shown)


markww:
I found that the headers included in the generated Souffle contain some very non-trivial code, including exotic, subtle, lock-free data structures. As such, getting Souffle to emit languages other than C++ would still probably be pretty complicated. On the other hand, this PR does clarify and minimize the dependency graph, so I think it's still worth getting in.

bgogul (approved):
Thanks!

PR #258, opened 2022-01-06
Remove fake_predicate_arena file.

bgogul:
This was left out from a previous cleanup.


markww (approved):
Whoops, thanks for cleaning this up!

PR #261, opened 2022-01-07
Datalog IR; Translate AuthLogic to DLIR

aferr:
This is a port of the datalog ir referred to as
[DLIR](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/datalog_ir.rs)
and the compiler pass that translates auth logic AST nodes into DLIR
[here](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs)
from the rust implementations into CPP.


src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+  // (Internally, this is a stateful class because it uses a counter to
+  // generate fresh names for arguments. The public interface hides this 
+  // implementation detail.)
+  class LoweringToDatalogPass {
+   public:
+
+    static DLIRProgram* Lower(const Program& auth_logic_program) {
+      LoweringToDatalogPass* pass = new LoweringToDatalogPass();
+      DLIRProgram* dlir_program = pass->ProgToDLIR(auth_logic_program);
+      delete pass;
```
markww, line 41: Rather than allocating the `LoweringToDatalogPass` with `new` and `delete`ing it, just place it in automatic storage:

```C++
LoweringToDatalogPass pass;
```

In general, strive to write `new` and `delete` as little as you can get away with.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+      // `speaker says Y PredX :-
+      //    speaker says Y canActAs X, speaker says X PredX`
+      return DLIRCondAssertion(generated_lhs,
+          {speaker_says_y_can_act_as_x, speaker_says_x_pred});
+    }
+
+    DLIRProgram* ProgToDLIR(const Program& program) {
+      // TODO / WIP
+      return new DLIRProgram({}, {});
+    }
```
markww, line 122: Rather than returning a bare `DLIRProgram *`, I'd suggest returning a `std::unique_ptr<DLIRProgram>`. If your ownership model permits it, it will prevent the whole class of lifetime confusion problems (eg, use after free, double free, etc)

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+    // few places during the translation.
+    std::string FreshVar() {
+      return "x__" + std::to_string(fresh_var_count_++);
+    }
+
+    // This function makes a copy of a predicate that prepends the name with
+    // another string and prepends the list of arguments with additional 
+    // arguments. This is used in a few places in the translation, for
+    // example, to translate "X says blah(args)" into "says_blah(X, args)".
+    Predicate PushOntoPred(std::string modifier,
```
markww, line 61: It looks like this modifier is just used to pass in strings that are, eventually, string literals. If that's the case, using `absl::string_view` here will allow passing in the string literal without wrapping it in a `std::string` which will. This is nice because the `std::string` constructor will, semantically, copy the data in that string literal. `absl::string_view`, on the other hand, will continue allowing use of the pointer into the `.rodata` section. Even if you pass in a `std::string` in some circumstances, `absl::string_view` should play nicely with that (unless you are doing something crazy like running a multithreaded program and deleting the original string partway through).

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+        std::vector<std::string> new_args, const Predicate& predicate) {
```
markww, line 62: It looks like `predicate` here has its contents entirely copied into the new `Predicate`. If that is the case, it is likely better to have `Predicate` be passed as a value directly rather than as a reference (ie, `Predicate predicate` rather than `const Predicate &predicate) so that it plays nicely with copy elision when the original value can be discarded.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+      std::string new_name = modifier + predicate.name();
+      new_args.insert(new_args.end(), predicate.args().begin(),
+          predicate.args().end());
```
markww, line 65: If you follow my suggestion to make `predicate` be of type `Predicate`, you will want to use [std::make_move_iterator](https://en.cppreference.com/w/cpp/iterator/make_move_iterator) to efficiently move the values out of `predicate.args_`

src/ir/auth_logic/lowering_ast_datalog.h
markww, line 61: It looks like this whole method copies all of `Predicate` just to push a modifier onto the `Predicate`'s name. I think there are a few ways to do this that are a bit more concise:

1. Abandon the immutability of `Predicate` and modify the name field. Very concise, but could cause bad behavior with how you've laid out building these things up.
2. Make `Predicate` be an interface with two variants: `LeafPredicate` and `PredicateModifier`. `LeafPredicate` can have the same fields as `Predicate` right now, while `PredicateModifier` shall look something like this:

```C++
class PredicateModifier {
  /* ... */
  private:
  absl::string_view modifier_;
  Predicate wrapped_predicate_;
};
```

This, effectively, produces a linked list of modifiers to lazily apply to the wrapped name.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! This file contains the datalog IR (DLIR) which makes the translation from
+//! this authorization logic into datalog simpler.
+#ifndef SRC_IR_AUTH_LOGIC_LOWERING_AST_DATALOG_H_
+#define SRC_IR_AUTH_LOGIC_LOWERING_AST_DATALOG_H_
+
+#include "ast.h"
```
markww, line 22: This `#include` should contain the full path from the root of the project to this header.

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,67 @@
...
+
+  private:
+    std::variant<Predicate, DLIRCondAssertion> value_;
+};
+
+class DLIRProgram {
+  public:
+    DLIRProgram(std::vector<DLIRAssertion> assertions,
+        std::vector<std::string> outputs)
+      : assertions_(assertions), outputs_(outputs) {}
```
markww, line 58: These `vector`s should be `std::move`'d into the fields.

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+//! This file contains the datalog IR (DLIR) which makes the translation from
+//! this authorization logic into datalog simpler.
+#ifndef SRC_IR_AUTH_LOGIC_DATALOG_IR_H_
+#define SRC_IR_AUTH_LOGIC_DATALOG_IR_H_
+
+#include <vector>
+#include "ast.h"
```
markww, line 23: This needs to be reformatted according to https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! This file contains the datalog IR (DLIR) which makes the translation from
+//! this authorization logic into datalog simpler.
```
markww, line 18: What does this `//!` syntax mean? Does it differ from just `//`?

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,67 @@
...
+      : assertions_(assertions), outputs_(outputs) {}
+
+  private:
+    std::vector<DLIRAssertion> assertions_;
+    std::vector<std::string> outputs_;
+};
+
+} // namespace raksha::ir::auth_logic
+
+#endif  // SRC_IR_AUTH_LOGIC_AST_H_
```
markww, line 67: This comment does not match the include guard name.

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright 2021 The Raksha Authors
```
markww, line 2: It is now 2022.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+   private:
+    LoweringToDatalogPass()
+      : fresh_var_count_(0) {}
+
+    int fresh_var_count_;
+
+    // This returns a fresh variable name. Fresh variables are introduced in a
+    // few places during the translation.
+    std::string FreshVar() {
+      return "x__" + std::to_string(fresh_var_count_++);
```
markww, line 54: In general, avoid using the postfix `++` operator, preferring the prefix version instead (https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement). postfix `++` has some very subtle semantics that are easy to get confused on, while prefix `++` is easy to reason about.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,128 @@
...
+
+      // This is `speaker says X PredX`
+      Predicate speaker_says_x_pred = PushPrin(std::string("says_"),
+          speaker, main_predicate);
+
+      // This is the full generated rule:
+      // `speaker says Y PredX :-
+      //    speaker says Y canActAs X, speaker says X PredX`
+      return DLIRCondAssertion(generated_lhs,
+          {speaker_says_y_can_act_as_x, speaker_says_x_pred});
```
markww, line 116: As the final uses of `generated_lhs`, `speaker_says_y_can_act_as_x`, and `speaker_says_x_pred`, these should use `std::move`.

aferr: fixed

src/ir/auth_logic/lowering_ast_datalog.h
aferr: fixed.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,212 @@
...
+    return std::make_pair(dummy_pred, dummy_assertions);
+  }
+
+  // WIP need to figure out error message about the visitor being
+  // non-exhaustive even though it looks exhaustive AFAICT
+  std::vector<DLIRAssertion>
+  SingleSaysAssertionToDLIR(Principal speaker,
+      // BaseFact base_fact,
+      Assertion assertion) {
+
```
aferr, line 174: @markww @bgogul Any idea why this doesn't work with AssertionVariantType? I get an error that this overloaded/visit is non-exhaustive even though `auto arg` should cover everything (also explicitly exhausting the 2 possibilities does not work either). This same overloaded/visit thing works just fine on BaseFact.GetValue() even if an implementation for just `auto arg` is given.

If this takes too long to look at, don't worry about it.

aferr: Never mind, I figured this out (locally).

aferr:
@markww now it might be a good point to do a re-review just at the `is this good C++` level and not so much at the `does this algorithm make sense` level (I'll add better comments to justify the second one later). See what you think of the way variants are visited (in particular that an accessor to the variant is added back in). Also I think I needed to add more const references to iterate over vectors... maybe some of these can actually be eliminated somehow, though?

src/ir/auth_logic/ast.h
```diff
@@ -63,9 +63,10 @@ class Predicate {
 class Attribute {
  public:
   explicit Attribute(Principal principal, Predicate predicate)
-    : principal_(principal), predicate_(predicate) {}
+      : principal_(principal), predicate_(predicate) {}
```
markww, line 66: `std::move` from parameters to field initializers.

src/ir/auth_logic/ast.h
```diff
@@ -76,34 +77,34 @@ class Attribute {
...
+// other facts through "canSay". "Fact" includes the recursive case. It is
+// necessary that BaseFact is a separate class from Fact because the RHS of
+// conditional assertions can only contain BaseFacts.
 // Base facts have three forms:
 //  - A predicate
 //  - An attribute
 //  - A canActAs expression
-// The particular choice of form is represented with a variant which is 
-// encapsulated.
+using BaseFactVariantType = std::variant<Predicate, Attribute, CanActAs>;
```
markww, line 98: Putting public aliases in the header in this fashion is not allowed by the style guide, unfortunately. See https://google.github.io/styleguide/cppguide.html#Aliases . You could put this `using` alias in the `private` section of the class definition, but that wouldn't help you with not repeating the type.

src/ir/auth_logic/ast.h
```diff
@@ -112,61 +113,65 @@ class Fact;
...
+ public:
+  explicit CanSay(Principal principal, std::unique_ptr<Fact>& fact)
       : principal_(principal), fact_(std::move(fact)) {}
-    const Principal& principal() const { return principal_; }
-    const Fact* fact() const { return fact_.get(); }
-  private:
-    Principal principal_;
-    std::unique_ptr<Fact> fact_;
+  const Principal& principal() const { return principal_; }
+  const Fact* fact() const { return fact_.get(); }
```
markww, line 120: You can make this slightly more null safe by dereferencing before returning, like so:

```C++
const Fact& fact() const { return *fact_.get(); }
```

That advice assumes, of course, that `fact_.get()` can never be `nullptr`.

src/ir/auth_logic/ast.h
```diff
@@ -112,61 +113,65 @@ class Fact;
...
+ private:
+  Principal principal_;
+  std::unique_ptr<Fact> fact_;
 };
 
-// Fact corresponds to either a base fact or a an expression of the form 
-// <principal> canSay 
+// Fact corresponds to either a base fact or a an expression of the form
+// <principal> canSay <Fact>
+using FactVariantType = std::variant<BaseFact, std::unique_ptr<CanSay>>;
```
markww, line 129: Similar comment to the other `using` alias, unfortunately.

src/ir/auth_logic/ast.h
```diff
@@ -112,61 +113,65 @@ class Fact;
...
+
+ private:
+  Fact lhs_;
+  std::vector<BaseFact> rhs_;
 };
 
 // Assertions can have two forms:
 //  - Facts
 //  - Conditional assertions
+using AssertionVariantType = std::variant<Fact, ConditionalAssertion>;
```
markww, line 158: Similar to the other public `using` alias.

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2022 The Raksha Authors
```
markww, line 2: Shouldn't this be `Copyright 2022 Google LLC.`?

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,315 @@
+/*
+ * Copyright 2022 The Raksha Authors
```
markww, line 2: `Copyright 2022 Google LLC`?

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,315 @@
...
+class LoweringToDatalogPass {
+ public:
+  static DLIRProgram Lower(Program auth_logic_program) {
+    return LoweringToDatalogPass().ProgToDLIR(auth_logic_program);
+  }
+
+ private:
+  LoweringToDatalogPass() : fresh_var_count_(0) {}
+
+  int fresh_var_count_;
```
markww, line 47: Unlikely to matter, but if we're just assuming that a counter never rolls over, I like to make it explicitly unsigned 64 bit. That gives you enough space to argue that with confidence. I'd say `uint64_t fresh_var_count_;`.

markww: Also, a nit: the declarations are not in the preferred order (fields, usually, should be towards the end of the section) https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,315 @@
...
+            [this, &speaker](Attribute attribute) {
+              std::vector<DLIRAssertion> pred_list = 
+                {SpokenAttributeToDLIR(speaker, attribute)};
+              return std::make_pair(AttributeToDLIR(attribute), pred_list);
+            },
+            [this, &speaker](CanActAs canActAs) {
+              std::vector<DLIRAssertion> pred_list =
+                {SpokenCanActAsToDLIR(speaker, canActAs)};
+              return std::make_pair(CanActAsToDLIR(canActAs), pred_list);
+            }},
```
markww, line 191: Rather than explicitly listing the types and doing almost the same thing with each type, would a generic lambda be better? I would view using `overloaded` in this way as a bit of a later resort; it would be nicer if the same code could uniformly handle these different variants.

src/ir/auth_logic/ast.h
aferr: done.

src/ir/auth_logic/ast.h
aferr: Assumption should hold. Done.

src/ir/auth_logic/datalog_ir.h
aferr: Oh I must have missed a `sed` on the codebase. done.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: done.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: OK done. also added an assert to cover rollover.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: Hmmmm.... but it's not doing almost the same thing for each type. It calls a different function for each type. There's almost no code reuse opportunity between the variants AFAICT. (Could be I don't really understand your proposal though).

src/ir/auth_logic/ast.h
aferr: From what I can tell, the decision in the style guide is that aliases can't be public in a header unless they are meant to be a part of the public API. In this case, I think it's fair to say that these types _are_ a part of the public API because here variants are traversed by `std::visit`s in compiler passes that operate on the AST. This might seem strange because it might seem like the variant should be encapsulated, but really I think there is no need to encapsulate this since the variant just represents the alternative forms for this piece of syntax. I added a comment to document how these aliases are meant to be used, like the style guide suggests should be done in these cases.

aferr: (Leaving this one open to see if you agree :)

src/ir/auth_logic/ast.h
aferr: see other thread.

src/ir/auth_logic/ast.h
aferr: See other thread.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: OK I think this could be done later without disrupting how this pass or the emitter work. I'm still inclined to get this working and tested and then do this as another PR after this is farther along. I opened an issue to track this here: [[Authorization Logic // C++] Performance Improvement in Translation to Datalog #267
](https://github.com/google-research/raksha/issues/267)

src/ir/auth_logic/ast.h
markww: I guess I could buy that, especially with the errors that it prevents in restating the type. I still do feel like this alias should be encapsulated though; perhaps you could make it a public member of the class in which it appears?

src/ir/auth_logic/lowering_ast_datalog.h
markww: Although each case is slightly different, you can view them as "the same" in an abstract sense. I think each of the branches fits the following pattern:

```C++
std::vector<DLIRAssertion> pred_list = SpokenItemToDLIRPredList(speaker, item);
return std::make_pair(ItemToDLIR(item), pred_list);
```

Now you can imagine `SpokenItemToDLIRPredList` as being overloaded in the following fashion:

```C++
std::vector<DLIRAssertion> SpokenItemToDLIRPredList(Principal, Predicate) { return { } }

std::vector<DLIRAssertion> SpokenItemToDLIRPredList(Principal speaker, Attribute attr) {
  return SpokenAttributeToDLIR(speaker, attr);
}

std::vector<DLIRAssertion> SpokenItemToDLIRPredList(Principal speaker, CanActAs caa) {
  return SpokenAttributeToDLIR(speaker, canActAs);
}
```

Other patterns might be cleaner. For instance, you could add a `SpokenCanActAsToDLIRPredList` to each of the classes and just call a single member function on `item`. The exact pattern to be used is up in the air. But the point is, in an abstract sense, these overloads are all the same function, and you might as well use C++'s overload resolution and generic lambda functionality to use less code to do this.

src/ir/auth_logic/ast.h
```diff
@@ -112,61 +117,73 @@ class Fact;
...
+// This type alias represents the different forms of Facts
+// and should be used by client code that traverses the AST. This type may be
+// extended to include more forms in the future.
+using FactVariantType = std::variant<BaseFact, std::unique_ptr<CanSay>>;
 class Fact {
  public:
   explicit Fact(BaseFact base_fact) : value_(std::move(base_fact)) {}
-  explicit Fact(std::unique_ptr<CanSay> can_say)
-    : value_(std::move(can_say)) {}
+  explicit Fact(std::unique_ptr<CanSay> can_say) : value_(std::move(can_say)) {}
```
markww, line 141: Rather than creating a separate constructor for each variant, you can just have a single constructor with argument type `FactVariantType`, and the implicit conversions from variant options to the variant type will do the work for you.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+    // an additional rule to implement this behavior. If the attribute
+    // under translation is `X PredX`, the additional rule is:
+    // `speaker says Y PredX :-
+    //    speaker says Y canActAs X, speaker says X PredX`
+    // (Where Y is a fresh variable)
+    Principal prin_y = Principal(FreshVar());
+
+    // This is `speaker says Y PredX`
+    Predicate generated_lhs =
+        PushPrin(std::string("says_"), prin_y, main_predicate);
```
markww, line 192: In general, if you're wrapping a string literal with `std::string`, that's a sign that something is suboptimal. `std::string` is basically an owned buffer of characters, and introduces a copy (which the compiler may or may not be able to magically eliminate through optimizations). This is bad because string literals are highly optimized; each string literal within a translation unit with the same contents points to the same place in the `.rodata` section. Usually, this means that the parameter that this argument feeds into should be an `absl::string_view`, which handles this case without copying.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+      ret.insert(ret.end(), std::make_move_iterator(single_translation.begin()),
+                 std::make_move_iterator(single_translation.end()));
+    }
+    return ret;
+  }
+
+// Queries are like predicates with arity 0. Souffle does not have predicates
+// with arity 0, so we represent them as having one argument which is a
+// constant.
+#define DUMMY_PREDICATE Predicate("grounded_dummy", {"dummy_var"}, kPositive)
```
markww, line 406: In general, [macros are not encouraged in Google C++]( https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros). Try making this a `static const` global instead, like so:

```C++
static const Predicate kDummyPredicate("grounded_dummy", {"dummy_var"}, kPositive);
```

The compiler will "do the right thing" and inline construction of this global if optimizations are on and it thinks it will help.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+
+  DLIRAssertion QueryToDLIR(const Query& query) {
+    // The assertions that are normally generated during the translation
+    // from facts to dlir facts are not used for queries.
+    auto [main_pred, not_used] = FactToDLIR(query.principal(), query.fact());
+    main_pred = PushPrin(std::string("says_"), query.principal(), main_pred);
+    auto lhs = Predicate(query.name(), {"dummy_var"}, kPositive);
```
markww, line 413: Don't do this `auto var = Type(...);` syntax. Instead just do `Type var(...);`.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+    return DLIRAssertion(DLIRCondAssertion(lhs, {main_pred, DUMMY_PREDICATE}));
+  }
+
+  std::vector<DLIRAssertion> QueriesToDLIR(const std::vector<Query>& queries) {
+    std::vector<DLIRAssertion> ret = {};
+    for (const Query& query : queries) {
+      ret.push_back(QueryToDLIR(query));
+    }
+    return ret;
+  }
```
markww, line 423: Might be nicer to just do the following:

1. Have `queries` be of value type `std::vector<Query>`, introducing an elidable copy at calls to the function.
2. Have `ret` be populated via an `insert` at the end, like so:
```C++
insert(ret.end(), std::make_move_iterator(queries.begin()), std::make_move_iterator(queries.end()):
```

(I did not look up the `insert` function, syntax might be slightly wrong).

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+
+  DLIRProgram ProgToDLIR(const Program& program) {
+    auto dlir_assertions = SaysAssertionsToDLIR(program.says_assertions());
+    auto dlir_queries = QueriesToDLIR(program.queries());
+    // We need to add a fact that says the dummy variable used in queries is
+    // grounded.
+    auto dummy_assertion = DLIRAssertion(DUMMY_PREDICATE);
```
markww, line 430: Just make the locals have the explicit type and follow the name with constructor args.

src/ir/auth_logic/ast.h
aferr: ok done.

src/ir/auth_logic/ast.h
```diff
@@ -76,34 +77,38 @@ class Attribute {
...
+// extended to include more forms in the future.
+using BaseFactVariantType = std::variant<Predicate, Attribute, CanActAs>;
 class BaseFact {
  public:
-  explicit BaseFact(Predicate predicate) : value_(std::move(predicate)) {};
-  explicit BaseFact(Attribute attribute) : value_(std::move(attribute)) {};
-  explicit BaseFact(CanActAs canActAs) : value_(std::move(canActAs)) {};
+  explicit BaseFact(Predicate predicate) : value_(std::move(predicate)){};
+  explicit BaseFact(Attribute attribute) : value_(std::move(attribute)){};
+  explicit BaseFact(CanActAs canActAs) : value_(std::move(canActAs)){};
```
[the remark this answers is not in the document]
aferr: ok done.

src/ir/auth_logic/ast.h
aferr: done.

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
+// A Datalog IR assertion is either:
+//  - an unconditional fact which is just a predicate
+//  - a conditional assertion
+class DLIRAssertion {
+ public:
+  explicit DLIRAssertion(Predicate predicate) : value_(std::move(predicate)) {}
+  explicit DLIRAssertion(DLIRCondAssertion condAssertion)
+      : value_(condAssertion) {}
```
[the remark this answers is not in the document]
aferr: effectively done

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+// ```Datalog
+//      Q_NAME("dummy_var") :- grounded("dummy_var"), ASSERTION(<ARGS>).
+//      grounded("dummy_var").
+//      .output Q_NAME
+// ```
+
+class LoweringToDatalogPass {
+ public:
+  static DLIRProgram Lower(Program auth_logic_program) {
+    return LoweringToDatalogPass().ProgToDLIR(auth_logic_program);
```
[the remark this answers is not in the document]
aferr: done.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+                    std::make_move_iterator(predicate.args().begin()),
+                    std::make_move_iterator(predicate.args().end()));
+    Sign sign_copy = predicate.sign();
+    return Predicate(new_name, new_args, sign_copy);
+  }
+
+  Predicate PushPrin(std::string modifier, Principal principal,
+                     Predicate predicate) {
+    return PushOntoPred(modifier, {principal.name()}, predicate);
+  }
```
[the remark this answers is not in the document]
aferr: modifier yes, predicate no. fixed.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: is `absl::string_view` better than just leaving it as a literal? I just left it as a literal.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: done

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,449 @@
...
+    return LoweringToDatalogPass().ProgToDLIR(auth_logic_program);
+  }
+
+ private:
+  LoweringToDatalogPass() : fresh_var_count_(0) {}
+
+  // This returns a fresh variable name. Fresh variables are introduced in a
+  // few places during the translation.
+  std::string FreshVar() {
+    assert(fresh_var_count_ < std::numeric_limits<uint64_t>::max());
```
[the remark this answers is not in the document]
aferr: done (locally).

src/ir/auth_logic/lowering_ast_datalog.h
aferr: That doesn't quite compile. I did a similar thing that's probably also fine?

src/ir/auth_logic/lowering_ast_datalog.h
aferr: Can't make queries be of value type because program is a const reference since it should not be moved.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: done

aferr:
@bgogul ready for review when you have a moment. Although there are no tests. Would you rather I add the emitter and write tests in the same PR since this took a while, or is it better to get this in without tests?

aferr:
@bgogul Please do take a look despite what the failed build says. For some reason my command for linting things causes breaks, it seems. I'll revert the lint commit, and it should work.

aferr:
OK FYI the includes in the file `ast_test.cc` are not in quite the Google style order because putting them in that order causes clang to stop believing that `std::unique_ptr` is real. Hopefully this is either an acceptable exception or you have thoughts on some other way to fix this.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: OK did this for the one you are commenting on. For the other ones in which, for reasons I can't fathom, the compiler insists the args need to be const refs, I can't get the overloading to work. So the other `visit`s, are still using the `overloaded` thing.

markww:
The Google-standard include order was chosen to, to the extent possible, expose issues where headers relied upon headers `#include`d by headers that happened to be placed previously. I'd say reorder the headers into the Google-standard order and see whether you need to include the header in which `std::unique_ptr` is introduced (I think `<memory>`?) in any other places .

aferr:
@markww BTW adding `<memory>` in more places did work.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,450 @@
...
+  // - A predicate, this is the main result of the translation
+  // - An additional rule that allows attributes to be passed with canActAs
+  // This function is for generating the additional rule.
+  // This takes both a speaker and the actual attribute as argument because
+  // the speaker is needed to generate the additional rule. Note that all base
+  // facts have a speaker either explicitly or implicitly:
+  // - If it appears on the LHS of an assertion, it explicitly has a speaker
+  // - If it appears on the RHS of an assertion, it behaves semantically
+  // like it has the same speaker as the head of the assertion.
+  DLIRAssertion SpokenAttributeToDLIR(Principal speaker, Attribute attribute) {
```
bgogul, line 180: Functions with large implementation should go to C++ file.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,450 @@
...
+        SpokenCanActAsToDLIR(speaker, canActAs)};
+    return std::make_pair(CanActAsToDLIR(canActAs), pred_list);
+  }
+
+  // The second return value represents 0 or 1 newly generated rules, so an
+  // option might seem more intuitive. However, the interface that consumes
+  // this needs to construct a vector anyway, so a vector is used in the
+  // return type.
+  std::pair<Predicate, std::vector<DLIRAssertion>> BaseFactToDLIR(
+      Principal speaker, BaseFact base_fact) {
```
bgogul, line 284: Can you change the arguments of all functions (except constructors)  to pass by const reference?  i.e., `const Principal&` instead of `Principal`, `const BaseFact&` instead of `BaseFact`. 

Without `std::move`, we seem to be making too many copies.

bgogul (commented):
@aferr a couple of style suggestions. 
 - Let us arguments of functions (other than constructors) arguments by const reference. i.e., let us use `const T&` in the place of `T`.
 - Move the large functions to a C++ file.

aferr:
> @aferr a couple of style suggestions.
> 
> * Let us arguments of functions (other than constructors) arguments by const reference. i.e., let us use `const T&` in the place of `T`.
> * Move the large functions to a C++ file.

@bgogul , I think you and @markww gave opposite feedback about whether or not `const T&` should be used :). IIRC, Mark says sounds like modern C++ compilers have copy elision, and so it is actually better to not use `const T&` anymore.

bgogul:
> > @aferr a couple of style suggestions.
> > 
> > * Let us arguments of functions (other than constructors) arguments by const reference. i.e., let us use `const T&` in the place of `T`.
> > * Move the large functions to a C++ file.
> 
> @bgogul , I think you and @markww gave opposite feedback about whether or not `const T&` should be used :). IIRC, Mark says modern C++ compilers have copy elision, and so it is actually better to not use `const T&` anymore.

I know. :) It is a bit subtle.  Copy elision also requires that use `std::move` when you are passing values to other functions, but we don't use `std::move` at the required places in this PR.  However, using `std::move` can introduce use-after-free like errors if we are not careful. Given how you pass values around a bunch of functions in this PR, `const T&` seems better here.

As a rule of thumb, let us default to `const T&` unless we are absolutely sure about it. It will give us less head ache in the long run.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: done. (also for a few others).

src/ir/auth_logic/lowering_ast_datalog.h
aferr: Yes. This makes sense to me.

src/ir/auth_logic/datalog_ir.h
```diff
@@ -0,0 +1,69 @@
...
+
+// A Datalog IR assertion is either:
+//  - an unconditional fact which is just a predicate
+//  - a conditional assertion
+class DLIRAssertion {
+ public:
+  // DLIRAssertionVariantType represents the alternative forms for
+  // DLIRAssertions. Client code should use this type for traversing
+  // DLIRAssertions. This type may be changed in the future.
+  using DLIRAssertionVariantType = std::variant<Predicate, DLIRCondAssertion>;
```
bgogul, line 48: Same concerns as other uses of typedef-ed variants. OK for now as long as we resolve it soon.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  // "X says blah(args)" into "says_blah(X, args)".
+  Predicate PushOntoPredicate(absl::string_view modifier,
+                              std::vector<std::string> new_args,
+                              const Predicate& predicate) {
+    std::string new_name = absl::StrCat(std::move(modifier), predicate.name());
+    new_args.insert(new_args.end(),
+                    std::make_move_iterator(predicate.args().begin()),
+                    std::make_move_iterator(predicate.args().end()));
+    Sign sign_copy = predicate.sign();
+    return Predicate(new_name, new_args, sign_copy);
```
bgogul, line 160: You will need `std::move(new_args)` here.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  // This function is an abbreviation for `PushPrin` where:
```
bgogul, line 163: `PushPrin` in the language?

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  // - a modifier is added
+  // - just one new principal is added as an argument.
+  // This is a common case in this translation because it is used for
+  // `x says blah(args)` and `x canActAs y` and other constructions involving a
+  // principal name.
+  Predicate PushPrincipal(std::string modifier, const Principal& principal,
```
bgogul, line 169: `absl::string_view modifier` might be better. Also, remove `std::move(modifier)` if you use `absl::string_view`.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  // - A predicate, this is the main result of the translation
+  // - An additional rule that allows attributes to be passed with canActAs
+  // This function is for generating the additional rule.
+  // This takes both a speaker and the actual attribute as argument because
+  // the speaker is needed to generate the additional rule. Note that all base
+  // facts have a speaker either explicitly or implicitly:
+  // - If it appears on the LHS of an assertion, it explicitly has a speaker
+  // - If it appears on the RHS of an assertion, it behaves semantically
+  // like it has the same speaker as the head of the assertion.
+  DLIRAssertion SpokenAttributeToDLIR(Principal speaker, Attribute attribute);
```
bgogul, line 192: `(const Principal& speaker, const Attribute& attribute)`

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  // CanActAs also results in both a predicate and an extra rule that passes
+  // these facts around. This function is for generating the extra rule and
+  // it works similarly to "SpokenAttributeToDLIR".
+  DLIRAssertion SpokenCanActAsToDLIR(const Principal& speaker,
+                                     const CanActAs& can_act_as);
+
+  std::pair<Predicate, std::vector<DLIRAssertion>> BaseFactToDLIRInner(
+      const Principal& speaker, const Predicate& predicate) {
+    std::vector<DLIRAssertion> pred_list = {};
+    return std::make_pair(predicate, pred_list);
```
bgogul, line 212: `return std::make_pair(predicate, {})`  might work as well. Otherwise, use `std::move(pred_list)`.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  std::pair<Predicate, std::vector<DLIRAssertion>> BaseFactToDLIRInner(
+      const Principal& speaker, const Attribute& attribute) {
+    std::vector<DLIRAssertion> pred_list = {
+        SpokenAttributeToDLIR(speaker, attribute)};
+    return std::make_pair(AttributeToDLIR(attribute), pred_list);
```
bgogul, line 219: Use `std::move(pred_list)` to avoid a copy. 

Alternatively, the following should avoid a copy too:
```
    return std::make_pair(AttributeToDLIR(attribute), {SpokenAttributeToDLIR(...)}); 
```
However, it might make the code less readable.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  std::pair<Predicate, std::vector<DLIRAssertion>> BaseFactToDLIRInner(
+      const Principal& speaker, const CanActAs& canActAs) {
+    std::vector<DLIRAssertion> pred_list = {
+        SpokenCanActAsToDLIR(speaker, canActAs)};
+    return std::make_pair(CanActAsToDLIR(canActAs), pred_list);
```
bgogul, line 226: `std::move(pred_list)`

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+
+  // In the same way that attributes are passed around with "CanActAs", so
+  // are other "CanActAs" facts. To implement this, the translation of
+  // CanActAs also results in both a predicate and an extra rule that passes
+  // these facts around. This function is for generating the extra rule and
+  // it works similarly to "SpokenAttributeToDLIR".
+  DLIRAssertion SpokenCanActAsToDLIR(const Principal& speaker,
+                                     const CanActAs& can_act_as);
+
+  std::pair<Predicate, std::vector<DLIRAssertion>> BaseFactToDLIRInner(
```
bgogul, line 209: For a follow up, the versions of  `BaseFactTo...` functions could have been implemented with a visitor of `BaseFact`.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  // The second return value represents 0 or 1 newly generated rules, so an
+  // option might seem more intuitive. However, the interface that consumes
+  // this needs to construct a vector anyway, so a vector is used in the
+  // return type.
+  std::pair<Predicate, std::vector<DLIRAssertion>> BaseFactToDLIR(
+      const Principal& speaker, const BaseFact& base_fact) {
+    return std::visit(
+        [this, &speaker](auto value) {
```
bgogul, line 236: Do you need `&speaker`? Or, `speaker` sufficient?

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+
+  std::vector<DLIRAssertion> SaysAssertionToDLIR(
+      const SaysAssertion& says_assertion) {
+    std::vector<DLIRAssertion> ret = {};
+    for (const Assertion& assertion : says_assertion.assertions()) {
+      std::vector<DLIRAssertion> single_translation =
+          SingleSaysAssertionToDLIR(says_assertion.principal(), assertion);
+      ret.insert(ret.end(), std::make_move_iterator(single_translation.begin()),
+                 std::make_move_iterator(single_translation.end()));
+    }
```
bgogul, line 258: Just venting. No action needed. Native support for functional constructs like `map` or `flatmap` would make this much more readable. Oh well!

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+    return ret;
+  }
+
+  std::vector<DLIRAssertion> SaysAssertionsToDLIR(
+      const std::vector<SaysAssertion>& says_assertions) {
+    std::vector<DLIRAssertion> ret = {};
+    for (const SaysAssertion& says_assertion : says_assertions) {
+      auto single_translation = SaysAssertionToDLIR(says_assertion);
+      ret.insert(ret.end(), std::make_move_iterator(single_translation.begin()),
+                 std::make_move_iterator(single_translation.end()));
```
bgogul, line 268: Can you make this into an utility function and use it here and other places in this PR? 

(See the `&&` for `src` in the `MoveAppend, it is not a type. :) )
```C++
template <typename T> 
void MoveAppend(std::vector<T>& dst, std::vector<T>&& src) {
  dst.insert(
    dst.end(), 
    std::make_move_iterator(src.begin()), 
    std::make_move_iterator(src.end()));
}
```

Then, we can use as follows here:
```C++
  auto single_translation = SaysAssertionToDLIR(says_assertion)
  MoveAppend(ret, std::move(single_translation));
```

Once this PR is in, we can refactor `MoveAppend` to `src/utils`.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+    // The assertions that are normally generated during the translation
+    // from facts to dlir facts are not used for queries.
+    auto [main_pred, not_used] = FactToDLIR(query.principal(), query.fact());
+    main_pred = PushPrincipal("says_", query.principal(), main_pred);
+    Predicate lhs(query.name(), {"dummy_var"}, kPositive);
+    return DLIRAssertion(DLIRCondAssertion(lhs, {main_pred, kDummyPredicate}));
+  }
+
+  std::vector<DLIRAssertion> QueriesToDLIR(const std::vector<Query>& queries) {
+    std::vector<DLIRAssertion> ret = {};
```
bgogul, line 289: `ret.reserve(queries.size());` will avoid reallocations within the loop.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+    return ret;
+  }
+
+  DLIRProgram ProgToDLIR(const Program& program) {
+    auto dlir_assertions = SaysAssertionsToDLIR(program.says_assertions());
+    auto dlir_queries = QueriesToDLIR(program.queries());
+    // We need to add a fact that says the dummy variable used in queries is
+    // grounded.
+    DLIRAssertion dummy_assertion(kDummyPredicate);
+    dlir_assertions.insert(dlir_assertions.end(),
```
bgogul, line 302: Use `MoveAppend` if possible.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  // This function takes a `predicate` and creates a new predicate as follows:
+  //   - Prepend the `modifier` to the name of the given `predicate`.
+  //   - Prepend the given `args` to the original list of arguments of
+  //   `predicate`.
+  //
+  // This is used in a few places in the translation, for example, to translate
+  // "X says blah(args)" into "says_blah(X, args)".
+  Predicate PushOntoPredicate(absl::string_view modifier,
```
bgogul, line 152: Actually, all functions from here on could move to the cc file.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,319 @@
...
+    main_pred = PushPrincipal("says_", query.principal(), main_pred);
+    Predicate lhs(query.name(), {"dummy_var"}, kPositive);
+    return DLIRAssertion(DLIRCondAssertion(lhs, {main_pred, kDummyPredicate}));
+  }
+
+  std::vector<DLIRAssertion> QueriesToDLIR(const std::vector<Query>& queries) {
+    std::vector<DLIRAssertion> ret = {};
+    for (const Query& query : queries) {
+      ret.push_back(QueryToDLIR(query));
+    }
```
bgogul, line 292: We could also create helper called  for this. Leave it for now. I will try to add one to src/utils.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,177 @@
...
+ */
+
+#include "src/ir/auth_logic/lowering_ast_datalog.h"
+
+#include "src/ir/auth_logic/ast.h"
+
+namespace raksha::ir::auth_logic {
+
+DLIRAssertion LoweringToDatalogPass::SpokenAttributeToDLIR(
+    Principal speaker, Attribute attribute) {
```
bgogul, line 24: `const &` for both.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,177 @@
...
+  Predicate main_predicate = AttributeToDLIR(attribute);
+
+  // Attributes interact with "canActAs" because if "Y canActAs X"
+  // then Y also picks up X's attributes. We need to generate
+  // an additional rule to implement this behavior. If the attribute
+  // under translation is `X PredX`, the additional rule is:
+  // `speaker says Y PredX :-
+  //    speaker says Y canActAs X, speaker says X PredX`
+  // (Where Y is a fresh variable)
+  Principal prin_y = Principal(FreshVar());
```
bgogul, line 34: `Principal prin_y(FreshVar())` is more idiomatic C++.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,177 @@
...
+      generated_lhs, {std::move(speaker_says_y_can_act_as_x),
+                      std::move(speaker_says_x_can_act_as_z)}));
+}
+
+std::pair<Predicate, std::vector<DLIRAssertion>>
+LoweringToDatalogPass::FactToDLIR(const Principal& speaker, const Fact& fact) {
+  return std::visit(
+      raksha::utils::overloaded{
+          [this, &speaker](BaseFact base_fact) {
+            return BaseFactToDLIR(speaker, base_fact);
```
bgogul, line 98: Ideally this should be a function in `BaseFact`. OK to defer to later as it is in the same spirit of exposing `std::variant`.

bgogul (commented):
This is a lot of code without tests, but I understand that you want to iterate. :) Address the comments when you get a chance. We will do iterate once more and get this in.

src/ir/auth_logic/ast.h
```diff
@@ -76,34 +78,35 @@ class Attribute {
...
-// encapsulated.
 class BaseFact {
  public:
-  explicit BaseFact(Predicate predicate) : value_(std::move(predicate)) {};
-  explicit BaseFact(Attribute attribute) : value_(std::move(attribute)) {};
-  explicit BaseFact(CanActAs canActAs) : value_(std::move(canActAs)) {};
+  // BaseFactVariantType gives the different forms of BaseFacts. Client code
+  // should use this type to traverse these forms. This type may be changed in
+  // the future.
+  using BaseFactVariantType = std::variant<Predicate, Attribute, CanActAs>;
```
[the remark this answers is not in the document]
aferr: > How much do you rely on this in the public interface though?

The `BaseFactVariantType` would be used in all the passes that traverse the AST (right now this is `lowering_ast_datalog`).

This comment about how the public type alias should be used exists because the style guide says there should be such a comment: https://google.github.io/styleguide/cppguide.html#Aliases

> In any case, OK for now. However, it is better to get this resolved sooner rather than later before create too much dependency on this type.

I think making the variant accessible through a public accessor is OK (and it is not poor encapsulation). I think if the variant is publicly accessible it means the different variations of `BaseFactVariantType` are exposed in the API. This means the structure of the AST is visible in the API, but I think there is no way to hide the structure of the AST from the API. Even with a visitor, I think some of the passes that work with the visitor would need to change if the structure of the AST changes.

Using a visitor is definitely the more traditional way to implement this, but I think using `variant` and `std::visit` makes good use of the modern C++ features, and is in some ways improved from a conventional visitor (for example the visitors can have return types more easily).

src/ir/auth_logic/datalog_ir.h
aferr: I'll resolve this one since we have another thread open to discuss this.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: No this is just an internal "helper function" for just this pass because this is a common pattern in the translation (some predicate is extended with a modifier and new arguments are added). I fixed the comment so that it has the full name "PushPrincipal" like you suggested for the function name.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: done.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: Oh nice. OK done.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: thanks. I learned something new. fixed.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: ok will resolve for now then.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: done.

bgogul (commented):
Some more code refactoring suggestions. Also, some of the changes (like copyright change) don't seem to be in the PR.

src/ir/auth_logic/datalog_ir.h
aferr: Yikes. I don't know how this happened... I hope it wasn't some bad use of git that dropped other changes. Fixed now anyway.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: Fixed.

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
aferr: OK here is an issue [Visitor For Authorization Logic AST #345](https://github.com/google-research/raksha/issues/345)

src/ir/auth_logic/lowering_ast_datalog.h
aferr: For some reason I thought you just meant all the functions after these internal helpers. I moved this one and the one below it over now.

src/ir/auth_logic/lowering_ast_datalog.cc
aferr: I'm assuming [this issue](https://github.com/google-research/raksha/issues/345) has resolved this for now, so I am resolving this. (?)

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,281 @@
+/*
+ * Copyright 2022 The Raksha Authors
```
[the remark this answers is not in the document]
aferr: changed.

src/ir/auth_logic/move_append.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2022 The Raksha Authors
```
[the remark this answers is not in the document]
aferr: changed.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,281 @@
...
+  return DLIRAssertion(DLIRCondAssertion(lhs, {main_pred, kDummyPredicate}));
+}
+
+std::vector<DLIRAssertion> LoweringToDatalogPass::QueriesToDLIR(
+    const std::vector<Query>& queries) {
+  std::vector<DLIRAssertion> ret = {};
+  ret.reserve(queries.size());
+  for (const Query& query : queries) {
+    ret.push_back(QueryToDLIR(query));
+  }
```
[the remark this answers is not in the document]
aferr: ok done. (A few bug fixes in the MapIter code).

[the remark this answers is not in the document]
aferr: Template inference is a bit disappointing... looks like we have to give the types for U and T on the call side sometimes.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,281 @@
...
+  // We need to add a fact that says the dummy variable used in queries is
+  // grounded.
+  DLIRAssertion dummy_assertion(kDummyPredicate);
+  MoveAppend(dlir_assertions, std::move(dlir_queries));
+  dlir_assertions.push_back(dummy_assertion);
+
+  std::vector<std::string> outputs = {};
+  for (const Query& query : program.queries()) {
+    outputs.push_back(query.name());
+  }
```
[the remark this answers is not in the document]
aferr: done.

src/ir/auth_logic/lowering_ast_datalog.h
aferr: I think we resolved this.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,281 @@
...
+    const std::vector<Query>& queries) {
+  std::vector<DLIRAssertion> ret = {};
+  ret.reserve(queries.size());
+  for (const Query& query : queries) {
+    ret.push_back(QueryToDLIR(query));
+  }
+  return ret;
+}
+
+DLIRProgram LoweringToDatalogPass::ProgToDLIR(const Program& program) {
```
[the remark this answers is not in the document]
aferr: See other thread.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -0,0 +1,249 @@
...
+      const Principal& speaker, const Fact& fact);
+
+  std::vector<DLIRAssertion> SingleSaysAssertionToDLIR(
+      const Principal& speaker, const Assertion& assertion);
+
+  std::vector<DLIRAssertion> SaysAssertionToDLIR(
+      const SaysAssertion& says_assertion);
+
+  std::vector<DLIRAssertion> SaysAssertionsToDLIR(
+      const std::vector<SaysAssertion>& says_assertions);
```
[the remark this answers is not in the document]
aferr: A lot of these do actually need to be member functions because they either directly call `FreshVar()` which is stateful or they call some other member that calls `FreshVar()`. I did remove the ones that are not stateful (which includes the `PushPredicate` helpers and the two base fact variants.)

aferr:
@bgogul opened issue https://github.com/google-research/raksha/issues/346

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+ */
+
+#include "src/ir/auth_logic/lowering_ast_datalog.h"
+
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/auth_logic/map_iter.h"
+#include "src/ir/auth_logic/move_append.h"
+
+namespace raksha::ir::auth_logic {
+
```
bgogul, line 24: Add an anonymous namespace around the file level local functions: 
```
namespace raksha::ir:auth_logic {
namespace {
   PushOntoPredicate(...) { 
   }
}
}
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+// This is used in a few places in the translation, for example, to translate
+// "X says blah(args)" into "says_blah(X, args)".
+Predicate PushOntoPredicate(absl::string_view modifier,
+                            std::vector<std::string> new_args,
+                            const Predicate& predicate) {
+  std::string new_name = absl::StrCat(std::move(modifier), predicate.name());
+  // This seemingly pointless line is needed to help C++ figure out the type
+  // of predicate.args() when it is passed to MoveAppend:
+  std::vector<std::string> pred_args = std::move(predicate.args());
+  MoveAppend(new_args, std::move(pred_args));
```
bgogul, line 39: MoveAppend(new_args, std::vector<std::string>>(predicate.args()));

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+// principal name.
+Predicate PushPrincipal(absl::string_view modifier, const Principal& principal,
+                        const Predicate& predicate) {
+  return PushOntoPredicate(modifier, {principal.name()}, predicate);
+}
+
+Predicate AttributeToDLIR(const Attribute& attribute) {
+  // If attribute is `X pred(args...)` the following predicate is
+  // `pred(X, args...)`
+  return PushPrincipal(std::string(""), attribute.principal(),
```
bgogul, line 58: `PushPrincipal("", ...);`

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+
+  // This is the full generated rule:
+  // `speaker says Y PredX :-
+  //    speaker says Y canActAs X, speaker says X PredX`
+  return DLIRAssertion(
+      DLIRCondAssertion(generated_lhs, {std::move(speaker_says_y_can_act_as_x),
+                                        std::move(speaker_says_x_pred)}));
+}
+
+Predicate CanActAsToDLIR(const CanActAs& can_act_as) {
```
bgogul, line 97: Move this along with other anonymous functions in this file..

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+  return DLIRAssertion(DLIRCondAssertion(
+      generated_lhs, {std::move(speaker_says_y_can_act_as_x),
+                      std::move(speaker_says_x_can_act_as_z)}));
+}
+
+std::pair<Predicate, std::vector<DLIRAssertion>>
+LoweringToDatalogPass::BaseFactToDLIRInner(const Principal& speaker,
+                                           const Predicate& predicate) {
+  std::vector<DLIRAssertion> pred_list = {};
+  return std::make_pair(predicate, std::move(pred_list));
```
bgogul, line 143:
```suggestion
  return std::make_pair(predicate, {});

// if above does not work:  return std::make_pair(predicate, std::vector<DLIRAssertion>({}));
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+}
+
+std::pair<Predicate, std::vector<DLIRAssertion>>
+LoweringToDatalogPass::BaseFactToDLIRInner(const Principal& speaker,
+                                           const Attribute& attribute) {
+  std::vector<DLIRAssertion> pred_list = {
```
bgogul, line 149: not a big deal, but you can inline the arguments in the `std::make_pair`

```C++
std::make_pair(AttributeToDLIR(attribute), {SpokenAttributeToDLIR(speaker, attribute)});
```

Similarly below if above pattern works.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -0,0 +1,304 @@
...
+        return BaseFactToDLIRInner(speaker, value);
+      },
+      base_fact.GetValue());
+}
+
+std::pair<Predicate, std::vector<DLIRAssertion>>
+LoweringToDatalogPass::FactToDLIR(const Principal& speaker, const Fact& fact) {
+  return std::visit(
+      raksha::utils::overloaded{
+          [this, speaker](BaseFact base_fact) {
```
bgogul, line 176: `const BaseFact& `?

bgogul (approved):
Looks good modulo some more comments.

src/ir/auth_logic/lowering_ast_datalog.cc
aferr: done

src/ir/auth_logic/lowering_ast_datalog.cc
aferr: done

src/ir/auth_logic/lowering_ast_datalog.cc
aferr: It really did need to be the commented out one.

src/ir/auth_logic/lowering_ast_datalog.cc
aferr: Right. That doesn't work. I need to put a `std::vector<DLIRAssertion>()` around it. So I just did that.

src/ir/auth_logic/lowering_ast_datalog.cc
aferr: yes that works. done.

Issue #262, opened 2022-01-12
Stop compiling custom, complex `CHECK`s into the Datalog program.


markww:
Currently, in order to handle the complex `check` tag predicates that may be expressed in the Arcs dataflow model, we are translating those `check` predicates into custom Datalog which we then compile into our Datalog program which performs taint analysis. This is necessary if we wish to have our Datalog analysis return whether or not the `check`s have succeeded or failed because these checks are too complex to express as simple Datalog `input` relations. While convenient, this approach has some serious downsides. It prevents us from precompiling our core analysis implementation for pushing tags around on the dataflow graph and we must compile a new version for each set of `check`s that may be provided by an incoming program. Compilation is a necessary step of running our analysis.

We could make it so that we communicate with our Datalog-based analysis only through `input` and `output` relations, allowing us to precompile the analysis, if we refactored the responsibilities of interpreting whether or not a check on tags has passed or failed. Rather than having the Datalog analysis interpret whether a check has passed or failed, we can consider its responsibility to be pushing tags around on the Dataflow graph only. Then, the C++ wrapper code that calls the Datalog analysis can use the `output` relations describing which tags are present at what locations to decide whether Arcs checks pass or fail.

Note that this would not free us entirely from compiling custom intelligence into the analysis. In particular, an analysis that we are intending to perform on SQL queries in Raksha supplies policies that we expect to dynamically match to nodes in a dataflow graph by dynamically generating Datalog to perform this matching. However, having to compile a different version of the analysis per policy that we wish to enforce is still better than having to compile a different version of the analysis per system that we wish to check against a policy.

markww:
Actually, I think the policies might be expressable as input relations as well with a bit of work. So I think it might be possible to make a truly precompiled, reusable Datalog analysis.

markww:
The approach I outlined in the original comment could plausibly work, but it was too awkward to split the responsibility across tools as described. Instead, I'm essentially transcribing the `Predicate` AST nodes from the IR into Datalog ADTs, which are then evaluated on the Datalog side.

markww:
This relates to the frontend for classic Arcs manifests, which is deprecated. Closing.

PR #263, opened 2022-01-14
Create CheckPredicate for performing checks via facts in DL.

markww:
Until now, we've been translating Arcs checks into Datalog rules. While
this gives us a great deal of expressive power, it is tricky to reason
about and requires recompiling before analyzing any Arcs system.

This PR takes the first step in a different direction: instead,
represent checks as regular old facts, with the predicates being
expressed as an ADT. These predicates can then be evaluated inside
Datalog through rules that interpret the predicates encoded in the ADT
rather than encoding the predicates directly as rule bodies.

This PR leaves the infrastructure for the previous way of handling
checks and predicates intact. However, this solution could (and quite
probably should) be used to switch off of that method entirely, allowing
us to precompile the Datalog analysis.

This issue makes progress towards, but does not fix, issue #262 .


build_defs/test_helpers.bzl
```diff
@@ -82,3 +89,68 @@ def extracted_datalog_string_test(
...
+        cmd = "sort $< > $@",
+        visibility = visibility
+    )
+
+    native.genrule(
+      name = name + "_check_names",
+      outs = ["checkNames"],
+      srcs = [check_file],
+      testonly = True,
+      cmd = "cut --delimiter=';' --fields=1 $< > $@",
```
bgogul, line 138: Document what is happening here?

build_defs/test_helpers.bzl
```diff
@@ -82,3 +89,68 @@ def extracted_datalog_string_test(
...
+      input_files: List; List of labels indicatinginput files to taint_exec.
+    """
+
+    input_files = other_input_files + [check_file]
+
+    facts_dir_opts = [
+        "--facts=`dirname $(location {})`".format(input_file) for input_file in input_files
+    ]
+
+    native.genrule(
```
bgogul, line 112: A single line comment for each genrule would be useful as documentation.

src/analysis/souffle/check_predicate.dl
```diff
@@ -0,0 +1,82 @@
...
+#ifndef SRC_ANALYSIS_SOUFFLE_CHECK_PREDICATE_DL_
+#define SRC_ANALYSIS_SOUFFLE_CHECK_PREDICATE_DL_
+
+#include "authorization_logic.dl"
+#include "dataflow_graph.dl"
+#include "tags.dl"
+
+.type Bool <: number
+
+.type CheckPredicate =
```
bgogul, line 25: Is my understanding correct that this is more expressive than what we have in C++ representation?

For instance, the C++ representation can only express something like `ap is tag1 or (tag2 and !tag3)`. However, this can represent `ap1 is tag1 or ap2 is tag2 and ap2 is tag3`. That is, we can combine predicates of different access paths. Is that correct?

src/analysis/souffle/check_predicate.dl
```diff
@@ -0,0 +1,82 @@
...
+  CP_TagPresence { ap: AccessPath, principal: Principal, tag: Tag }
+  | CP_And { lhs: CheckPredicate, rhs: CheckPredicate }
+  | CP_Or { lhs: CheckPredicate, rhs: CheckPredicate }
+  | CP_Implies { antecedent: CheckPredicate, consequent: CheckPredicate }
+  | CP_Not { to_negate: CheckPredicate }
```
bgogul, line 30: Consider `CP_Not { some_predicate }`. If `some_predicate` is `true`, will  `CP_Not { some_predicate }` be true or false?

src/analysis/souffle/check_predicate.dl
```diff
@@ -0,0 +1,82 @@
...
+  checkAndExpectation(name, pred, _), predicateEval(pred, result), result != 0.
+
+checkAndResult(name, "FAIL") :-
+  checkAndExpectation(name, pred, _), predicateEval(pred, result), result = 0.
+
+// A check which matches the given expectation.
+.decl checkMatchesExpectation(name: symbol)
+.output checkMatchesExpectation(delimiter=";")
+
+checkMatchesExpectation(name) :- checkAndResult(name, result), checkAndExpectation(name, _, result).
```
bgogul, line 80: Should these be part of a test util rather than a part of the core analysis?

src/analysis/souffle/check_predicate.dl
```diff
@@ -0,0 +1,82 @@
...
+//-----------------------------------------------------------------------------
+#ifndef SRC_ANALYSIS_SOUFFLE_CHECK_PREDICATE_DL_
+#define SRC_ANALYSIS_SOUFFLE_CHECK_PREDICATE_DL_
+
+#include "authorization_logic.dl"
+#include "dataflow_graph.dl"
+#include "tags.dl"
+
+.type Bool <: number
+
```
bgogul, line 24: Do we have a good sense of what the performance of ADT in datalog is?

bgogul (commented):
I like the overall direction, but left some questions and comments. PTAL.

src/analysis/souffle/check_predicate.dl
```diff
@@ -0,0 +1,82 @@
...
+.type CheckPredicate =
+  CP_TagPresence { ap: AccessPath, principal: Principal, tag: Tag }
```
aferr, line 26: Just FYI:
- the error messages when using ADTs can be a bit cryptic at times (you might have already run into this :)
- ADTs are not a usual Datalog feature, they're Souffle-specific. So if we pivot to rolling our own solver, the ADTs might be less portable to our own solver.

You shouldn't necessarily let these points scare you off from using ADTs, but I just thought I'd let you know.

build_defs/test_helpers.bzl
markww: Sounds good, will do.

build_defs/test_helpers.bzl
markww: Will do.

src/analysis/souffle/check_predicate.dl
markww: No, this is exactly as expressive as what we have in the C++ representation. Each of the ADT branches correspond directly to a `Predicate` we use within checks on the C++ side.

You might be thinking of `claim`s, where we have retreated from the full expressiveness of Arcs because we're not sure what some of that expressiveness means in the overapproximated world of our analysis (for instance, if we claim that some node has `tag1 or tag2`, what does that mean for downstream nodes.

src/analysis/souffle/check_predicate.dl
markww: The result of `CP_Not` will be `false` if its input is `true`.

src/analysis/souffle/check_predicate.dl
markww: Perhaps. Let me see if I can refactor to remove test mechanism from the implementation.

src/analysis/souffle/check_predicate.dl
markww: That is an interesting question for which I cannot find a full answer online. But I can find some clues.

See this commit to Souffle: https://github.com/souffle-lang/souffle/commit/ddca2fdaeef03a3f6de77abff3a86403fe649b58 .

This commit, in its own words, optimizes ADTs by:

```
encode branch with a single argument as [branch_id, argument]
instead of [branch_id, [argument]]
```

What this tells me is that the various fields of a branch of an ADT are represented as a linked list internally. This means that exploding them and placing each field in a relation, as we do to evaluate the ADT, is probably very cheap. What might not be cheap is getting the last field on a branch of an ADT with many fields, as traversing to each successive field incurs an additional pointer dereference. None of our branches, however, have many fields, so this is probably ok.

That's all I know about ADTs in general. However, I think I have a pretty good understanding of how expensive using ADTs for this particular purpose will be. The evaluation of `check` ADTs will occur at the very end of the analysis; it must, because the leaf `TagPresence` branch consults `!mayHaveTag`, introducing a stratifiability requirement. This means that evaluation of the `CheckPredicate` ADTs will not be interleaved with pushing around tags for the analysis. I believe that this means that this will not occur as part of the inner loop of the analysis, and thus will be of least concern from a performance perspective.

src/analysis/souffle/check_predicate.dl
markww: Yes, that's true. If we switched to another Datalog dialect with lists and not ADTs, we could definitely just represent the ADTs as lists with a little less compiler support. If we switched to one without even that, we might be forced to do something ugly by using IDs to point across relations. But I think it is possible with plain ol' Datalog, so I'm not too worried about it for now.

src/analysis/souffle/check_predicate.dl
bgogul: The root `CP_TagPresence` also has an `AccessPath` as an argument. So, "(ap1 is tag1) or (ap2 is tag2)` can be represented as follows: 
```
CP_Or { 
   lhs: CP_TagPresence("ap1", "someone", "tag1"),
   rhs: CP_TagPresence("ap2", "someone", "tag2")
}
```

no? On the other hand, the C++ `TagPresence` only has a tag: 
 https://github.com/google-research/raksha/blob/ba14ca7d5a4aebbe31405bc39164dbd6079221e1/src/ir/predicate.h#L232

markww: Oh, I guess you are right. It is more expressive than the original in being able to speak about different `AccessPath`s in the same predicate. I hadn't noticed that.

Another interesting note: because it explicitly lists the `owner` predicate, it can talk about different `owner`s within the same expression tree as well. `owner`s cannot be explicitly listed in the current C++ representation. Currently, the C++ representation translates to a Datalog rule which accepts any owner of the given `tag`; I was thinking of adding a `TagPresenceDynamicOwner` to capture this behavior. I think it is valuable to have a way to require a particular owner as well, hence why I'm thinking of adding an alternative version of `TagPresence`.

bgogul: Sounds good. This is a good change in terms of expressiveness as long as it doesn't blow up the performance. :) Based on your other comment, it seems like we would be OK.

src/analysis/souffle/check_predicate.dl
bgogul: Can you document it? It is not clear from the `to_negate` name.

bgogul (approved):
Looks good modulo refactoring, documentation and clean up suggestions.

build_defs/test_helpers.bzl
markww: Done.

build_defs/test_helpers.bzl
markww: Done.

src/analysis/souffle/check_predicate.dl
markww: I've refactored the code. It's not quite what you asked for, as the separation between test utils and regular analysis that we have previously accomplished is sort of an artifact of the approach of generating custom datalog which is compiled per-test.   However, I have made the core analysis code not depend upon test relations and have separated the test relations into a conditionally-compiled, test only portion of the Datalog file.

src/analysis/souffle/check_predicate.dl
markww: Done.

markww:
@bgogul , I think I addressed all of your comments. If you have a moment, though, it could be good for you to take a look at how I separated out the test relations for `CheckPredicate`s from the core relations. Separating them out into a separate test header included only by the test doesn't make sense in a world where the tests don't contain any custom Datalog, so I separated them out into a conditionally-compiled, test-only region of the `check_predicate.dl` Datalog file.

src/analysis/souffle/check_predicate.dl
```diff
@@ -0,0 +1,90 @@
...
+.decl checkAndResult(name: symbol, result: symbol)
+.output checkAndResult(delimiter=";")
+
+checkAndResult(name, "PASS") :- checkP(name, pred), predicateEval(pred, result), result != 0.
+
+checkAndResult(name, "FAIL") :- checkP(name, pred), predicateEval(pred, result), result = 0.
+
+// Relations used to support tests involving check facts. Not present in
+// production builds.
+#ifdef TEST
```
bgogul, line 69: I actually meant, move this to a separate file. OK to leave it as is for now.

PR #265, opened 2022-01-19
Rename Type to TypeBase and introduce a new Type wrapper class.

bgogul:
This refactoring is a staging step to make type a copyable value type. Specifically, this PR  does the following:
  - Renames `Type` to `TypeBase`
  - Introduces a wrapper `Type` class that holds onto `std::unique_ptr<TypeBase>`
  - replaces all _original_ uses of `std::unique_ptr<Type>` with `Type`.

### DIFF
--- a/src/ir/handle_connection_spec.h
+++ b/src/ir/handle_connection_spec.h
@@ -29,24 +29,24 @@ namespace raksha::ir {
 // equivalent ArcsManifestProtos.
 class HandleConnectionSpec {
  public:
-  explicit HandleConnectionSpec(
-    std::string name, bool reads, bool writes,
-    std::unique_ptr<types::Type> type)
-    : name_(std::move(name)), reads_(reads), writes_(writes),
-      type_(std::move(type)) {}
+  explicit HandleConnectionSpec(std::string name, bool reads, bool writes,
+                                types::Type type)
+      : name_(std::move(name)),
+        reads_(reads),
+        writes_(writes),
+        type_(std::move(type)) {}
 
   const std::string &name() const { return name_; }
   bool reads() const { return reads_; }
   bool writes() const { return writes_; }
-  const types::Type &type() const { return *type_; }
+  const types::Type &type() const { return type_; }
 
   // Get all of the AccessPaths rooted at the associated ParticleSpec
   // (indicated through the provided name) through leaf fields of type_.
   std::vector<AccessPath> GetAccessPaths(
       std::string particle_spec_name) const {
     std::vector<AccessPath> result_paths;
-    AccessPathSelectorsSet selectors_set =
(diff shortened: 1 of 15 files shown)


src/ir/types/type.h
```diff
@@ -19,6 +19,24 @@ class Type {
...
+    return type_->GetAccessPathSelectorsSet();
+  }
+
+  // TODO: This should be removed to hide the TypeBase.  method is added as a
+  // stop-gap solution to work with the rest of the code base.
+  const TypeBase& type_base() const { return *type_.get(); }
+
+ private:
+  std::unique_ptr<TypeBase> type_;
+};
```
markww, line 37: I wonder about this approach. I was thinking that `TypeBase` would be paired with something like `TypeRef`, where `TypeRef` would contain within it a small, efficiently-copyable unique ID that could be used to look up a `TypeBase` in a map in the factory. This `TypeRef` could then be used in multiple places without running into uniqueness problems, while the `std::unique_ptr` would be held by the factory.

I'm also concerned about whether or not we're helping by wrapping the `std::unique_ptr` in this fashion. Certainly `Type` is more concise than `std::unique_ptr<TypeBase>`, but it does hide the fact that the type cannot be copied and that it implies ownership.

src/ir/handle_connection_spec.h
```diff
@@ -29,24 +29,24 @@ namespace raksha::ir {
...
+      : name_(std::move(name)),
+        reads_(reads),
+        writes_(writes),
+        type_(std::move(type)) {}
 
   const std::string &name() const { return name_; }
   bool reads() const { return reads_; }
   bool writes() const { return writes_; }
-  const types::Type &type() const { return *type_; }
+  const types::Type &type() const { return type_; }
```
markww, line 42: If `Type` is a wrapper for `std::unique_ptr<TypeBase>`, this causes us to return a pointer to a pointer. I'm not worried about the performance per-se, I think though that this is perhaps a code smell of sorts.

src/ir/types/type.h
bgogul: This is just an intermediate step to unblock the rest of the IR work. Having something like a copyable `TypeRef` is the plan. In fact, `Type` would be the `TypeRef` you are referring to. However, this required fleshing out the canonicalization of types. So, I just started by creating the wrapper interface so that we can use it in the `operator`, `operation`, etc.

src/ir/handle_connection_spec.h
bgogul: This is just a temporary state and will be fixed by having `Type` cheaply copyable.

bgogul:
Thanks for the review, @markww. I replied to your concerns. PTAL.

markww (approved):
I think this is fine as an intermediate state. LGTM.

PR #266, opened 2022-01-19
Add type factory for creating types::Type instances.

bgogul:
This PR introduces a `TypeFactory` class that is used to create and register types.  The `TypeFactory` will eventually be used to canonicalize types. For now, the `TypeFactory` simply owns all the Schemas that are created. The newly introduced `TypeFactory` instance is plumbed through to all the methods that need to create Schemas/Types.

### DIFF
--- a/src/ir/types/type_factory.cc
+++ b/src/ir/types/type_factory.cc
@@ -1,5 +1,5 @@
 //-----------------------------------------------------------------------------
-// Copyright 2021 Google LLC
+// Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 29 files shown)


src/ir/types/type_factory.cc
```diff
@@ -0,0 +1,44 @@
...
+
+namespace raksha::ir::types {
+
+Type TypeFactory::MakePrimitiveType() {
+  return Type(std::make_unique<PrimitiveType>());
+}
+
+Type TypeFactory::MakeEntityType(const Schema& schema) {
+  return Type(std::make_unique<EntityType>(schema));
+}
```
markww, line 32: This pattern means that every use of a `PrimitiveType` must use a different instance of that `PrimitiveType`, as each usage point owns its `PrimitiveType`. As mentioned in the other PR, perhaps the factory should return a non-owning ID for the type which can be used to look up the full type, which would be owned by the factory.

src/ir/types/type_factory.h
```diff
@@ -0,0 +1,49 @@
...
+  // Create an entity type and wrap it in `types::Type`.
+  Type MakeEntityType(const Schema& schema);
+
+  // Create and register the given schema.
+  const Schema& RegisterSchema(std::optional<std::string> name,
+                               absl::flat_hash_map<std::string, Type> fields);
+
+ private:
+  // Known schemas.
+  std::vector<std::unique_ptr<types::Schema>> schemas_;
```
markww, line 44: I take it that this is temporary and we'll uniquify equivalent schemas in a later PR?

src/ir/types/type_factory.cc
bgogul: Yes, the plan is to canonicalize all types include `PrimitiveType`. As I mentioned in the other PR, I wanted to introduce the `Type` wrapper and `TypeFactory` so that we can start building out the `Operator`, `Operation`, etc. Do you see any concerns? e.g., does the `Type` wrapper as defined prevent us from achieving the canonicalization goals?

src/ir/types/type_factory.h
bgogul: Yes.

markww (approved):
I don't see any concerns from transitioning from this state to a state where the type factory owns and hands out IDs to the various types. Clearly it has all of the information needed to construct and own these types, as it is doing exactly that but handing the type ownership back out. It should be easy to switch to using type refs instead of unique pointers, as type refs are strictly easier to handle (being copyable and not owners). I think this makes sense as an intermediate state. LGTM.

PR #269, opened 2022-01-20
Make all datalog generated by IR frontend consist of simple facts.

markww:
This change causes all Datalog generated by the IR frontend consist of
simple facts without a rule body. While the IR frontend still generates
custom datalog, this change is a significant step to make it stop
generating custom datalog and instead just produce fact data that can be
consumed by the analysis.

This makes progress on issue #262.


src/analysis/souffle/may_will.dl
```diff
@@ -37,8 +37,23 @@ disallowedUsage(dataConsumer, usage, owner, tag) :-
 .decl allUsagesAllowed(dummy_var: symbol)
 
 .decl isUsage(usage: Usage)
+// DO NOT SUBMIT
+.output isUsage(IO=stdout)
```
bgogul: Left over from debugging?

src/analysis/souffle/may_will.dl
```diff
@@ -37,8 +37,23 @@ disallowedUsage(dataConsumer, usage, owner, tag) :-
 .decl allUsagesAllowed(dummy_var: symbol)
 
 .decl isUsage(usage: Usage)
+// DO NOT SUBMIT
```
aferr, line 40: It seems like there's some chance that this comment was meant to remind you not to submit this commit? Maybe it means something else, though. I thought I'd bring it up either way.

src/analysis/souffle/may_will.dl
```diff
@@ -37,8 +37,23 @@ disallowedUsage(dataConsumer, usage, owner, tag) :-
...
+allUsagesAllowed("dummy_var") :- count : disallowedUsage(_, _, _, _) = 0.
+
+// TODO: Find a better way to do this.
+// Catch may-will issues as if they were a check.
+checkAndResult("may_will", "FAIL") :- !allUsagesAllowed("dummy_var").
+
+// We phrase this as depending on saysMay so that tests that do not have "may"
+// or "will" involved will not be considered as having had a check of this form
+// at all. Again, we should find a better way to do this.
+checkAndResult("may_will", "PASS") :- allUsagesAllowed("dummy_var").
```
bgogul: Why was the earlier version not sufficient?

src/analysis/souffle/authorization_logic.dl
```diff
@@ -86,4 +100,11 @@ ownsTag(owner, tag) :- says_ownsTag(_, owner, tag).
 isPrincipal(owner) :- ownsTag(owner, _).
 isPrincipal(owner) :- ownsAccessPath(owner, _).
 
+isPrincipal(speaker), isAccessPath(ap), isPrincipal(owner), isTag(tag) :-
```
aferr, line 103: I didn't know you could write rules with multiple heads (or I forgot). Nice, that's easier to read.

src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver_with_check_facts.cc
```diff
@@ -0,0 +1,91 @@
+//-----------------------------------------------------------------------------
```
bgogul, line 1: We have one too many test drivers. :) Why could we not reuse the `fact_test_driver` itself?

src/analysis/souffle/may_will.dl
```diff
@@ -37,8 +37,23 @@ disallowedUsage(dataConsumer, usage, owner, tag) :-
...
-allUsagesAllowed("dummy_var") :- !disallowedUsage(dataConsumer, usage, owner, tag),
-    isPrincipal(dataConsumer), isUsage(usage), isTag(tag), ownsTag(owner, tag).
+isPrincipal(speaker), isPrincipal(actor), isUsage(usage), isTag(tag) :-
+  saysMay(speaker, actor, usage, tag).
+
+isPrincipal(speaker), isUsage(usage), isAccessPath(ap) :- saysWill(speaker, usage, ap).
+
+allUsagesAllowed("dummy_var") :- count : disallowedUsage(_, _, _, _) = 0.
+
+// TODO: Find a better way to do this.
```
aferr, line 50: Not sure if this TODO was meant to be resolved before or after this PR. Opening a comment just to catch the accidental case.

aferr:
This looks like an improvement. Just curious, what's the motivation to avoid generating rules? Is it to move more of the TCB into the part of the datalog that stays fixed? (This does seem worthwhile, to me).

src/analysis/souffle/authorization_logic.dl
markww: Isn't it nice? Especially when population should be correlated, as in this case.

src/analysis/souffle/may_will.dl
markww: It's to be resolved after this PR. But that confusion makes me realize that I should probably open an issue about it.

src/analysis/souffle/may_will.dl
markww: Good catch, I thought I got all of these. I put, effectively, debug prints by making some relations `.output`. I marked those with `// DO NOT SUBMIT`. I'll remove this.

markww:
The main win by generating only simple facts is to, eventually, make it so that we can take the compiler out of the loop when performing analysis. Currently, we have to recompile the datalog for each system that we wish to analyze. That's slow and can have unpredictable effects on the analysis if someone managed to get us to generate a rule or fact that we didn't anticipate.

After this PR, we still have to do that. It is, however, a stepping stone to having the frontend only generate facts to input to the analysis. This would allow us to address #262 (although the method I proposed there is wrong).

aferr:
Ah nice, I would not have thought of the performance implication. That does sound good too.

src/analysis/souffle/taint.dl
```diff
@@ -171,4 +171,16 @@ predicateEval($CP_TagPresence(ap, owner, tag), 1) :-
 predicateEval($CP_TagPresence(ap, owner, tag), 0) :-
   checkSubExprs($CP_TagPresence(ap, owner, tag)), !mayHaveTag(ap, owner, tag).
 
+// This helper relation is used to make it very clear that predicateEval can be
+// only 1 xor 0 on a $CP_TagPresenceDynamicOwner.
+.decl tagPresentForSomeOwner(ap: AccessPath, tag: Tag)
+tagPresentForSomeOwner(ap, tag) :-
+  checkSubExprs($CP_TagPresenceDynamicOwner(ap, tag)),
+  ownsAccessPath(owner, ap), mayHaveTag(ap, owner, tag).
```
bgogul, line 179: This seems to completely disconnect the principal from the tag. We should actually verify that for every owner `owner` of an access path `ap`, `mayHaveTag(ap, owner, tag)` is true?  However, `tagPresentForSomeOwner` will be true if any owner has a tag?

Doesn't this weaken the existing checks?

src/analysis/souffle/may_will.dl
markww: Done.

src/analysis/souffle/may_will.dl
markww: Updated this to point at issue #270 .

src/analysis/souffle/check_predicate.dl
```diff
@@ -24,6 +24,7 @@
 
 .type CheckPredicate =
   CP_TagPresence { ap: AccessPath, principal: Principal, tag: Tag }
+  | CP_TagPresenceDynamicOwner { ap: AccessPath, tag: Tag }
```
bgogul, line 27: I see why you introduced this, but it actually weakens the check. See the comment around `tagPresentForSomeOwner`.

bgogul (commented):
@markww the changes seem to weaken the checks. I am not sure if I am missing something. I left comments. PTAL.

src/ir/tag_check.h
```diff
@@ -33,21 +33,13 @@ class TagCheck {
...
-  // each check using the DatalogPrintContext. We unconditionally add that
-  // label to isCheck and add that label to check only if the check's
-  // predicate passes. This allows us, in a driver, to assert that the
-  // elements of isCheck and check are equal. If that is the case, all checks
-  // passed. If it is not, at least one check failed.
+  // Print out the tag check as a datalog fact.
   std::string ToDatalog(DatalogPrintContext &ctxt) const {
     constexpr absl::string_view kCheckHasTagFormat =
-        R"(isCheck("$0", "$1"). check("$0", owner, "$1") :-
-  ownsAccessPath(owner, "$1"), $2.)";
```
markww, line 46: Leaving this comment only so that I can refer to it elsewhere.

src/xform_to_datalog/datalog_facts.h
```diff
@@ -68,15 +68,8 @@ class DatalogFacts {
 .output duplicateTestCaseNames(IO=stdout)
 .output disallowedUsage(IO=stdout)
 
-.decl isCheck(check_index: symbol, path: AccessPath)
-.decl check(check_index: symbol, owner: Principal, path: AccessPath)
-
-allTests(check_index) :- isCheck(check_index, _).
-testFails(cat(check_index, "-", owner, "-", path)) :-
-  isCheck(check_index, path), ownsAccessPath(owner, path),
-  !check(check_index, owner, path).
```
markww, line 77: Another comment just so I can refer to this place.

src/analysis/souffle/taint.dl
markww: I don't think it does. See what we previously did at https://github.com/google-research/raksha/pull/269/files#r789881510 .

```C++
 constexpr absl::string_view kCheckHasTagFormat =
        R"(isCheck("$0", "$1"). check("$0", owner, "$1") :-
  ownsAccessPath(owner, "$1"), $2.)";
```

What that says is that we introduce a row in the `check` relation for any owner of the given `AccessPath` that also owns that tag. If the `check` row was present, we used that to populate `testFails` at this location https://github.com/google-research/raksha/pull/269/files#r789884147 as follows:

```C++
testFails(cat(check_index, "-", owner, "-", path)) :-
  isCheck(check_index, path), ownsAccessPath(owner, path),
  !check(check_index, owner, path).
```

That just says that the test fails if there does not exist an `owner` of the `path` for which `check` is populated.

Looking at my implementation, I populate `tagPresentForSomeOwner` if there exists an `owner` of the `AccessPath` indicated in the `CP_TagPresenceDynamicOwner` that also owns the `tag` we expect to be present. This is then used to populate `predicateEval` for that `CP_TagPresenceDynamicOwner` as true or false.

I think the two notations are actually semantically identical. If I'm missing an aspect that I've weakened, though, happy to change this to strengthen the enforcement of `CP_TagPresenceDynamicOwner`s.

src/analysis/souffle/examples/simple_may_will/simple_may_will_fail.authlogic
```diff
@@ -19,6 +19,7 @@
     ownsTag("EndUser", "rawVideo").
     ownsAccessPath("EndUser", "R.P1#0.foo").
     "P1" canSay hasTag(PathX, "EndUser", "rawVideo") :- isAccessPath(PathX).
+    ownsAccessPath("EndUser", "R.P1#0.foo").
```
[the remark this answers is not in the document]
markww: Whoops. I think I did this by mistake when I was trying to get tests to pass. Let me remove this.

src/analysis/souffle/taint.dl
bgogul:
```
testFails(cat(check_index, "-", owner, "-", path)) :- 
  isCheck(check_index, path), ownsAccessPath(owner, path), 
  !check(check_index, owner, path).
```
> That just says that the test fails if there does not exist an owner of the path for which check is populated.

Just to make sure we are on the same page, the check is not simply populated. It is a conditional relation as follows: 
``` 
check(index, owner, path) :- 
  ownsAccessPath(owner, path), mayHaveTag(path, owner, "some-tag-from-particle-spec").
```

Therefore, the test will fail even if _any_ owner of the path disagrees that it has the corresponding tag. However, in the new version, the test will succeed even if some owner of the path disagrees that it has the corresponding tag. no?

markww: Good point. I have updated the code to make it insist that all owners agree that the tag is present and the number of owners is greater than 1.

src/analysis/souffle/check_predicate.dl
markww: See new version of this, I think I've addressed the issue.

src/analysis/souffle/may_will.dl
markww: There are a few aspects to why I made this change.

1. Having the rule populating `checkAndResult("may_will"...)` occur in the boilerplate wrapper code produced by the IR translator is at odds with the goal of this PR, to make the IR translator only produce base facts. The changes of this PR make the IR translator's boilerplate contain only `.decl`s, base facts, and a few rules that are obviously aliasing (eg, `saysMay(...) :- says_may(...).`).

2. Initially, the rule only would add an entry into `testFails`; that's great if you are looking for the absence of failures, but it precludes performing other checks, like making sure that the number of `PASS`s is equal to the number of tests. This can cause confusing brittleness in checking test results. The new recipe always populates the `"may_will"` check as passing or failing.

3. The `testFails` rule must be populated by a custom rule; it is not set up to be populated by `check` facts. The new formulation can work with `check` facts.

src/analysis/souffle/examples/simple_may_will/simple_may_will_fail.authlogic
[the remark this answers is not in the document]
markww: Removed.

src/analysis/souffle/may_will.dl
markww: Yep. Removed.

src/analysis/souffle/tests/arcs_fact_tests/fact_test_driver_with_check_facts.cc
markww: OK, good point, I'll get rid of the new one and use one test driver at a time.

markww: Fixed.

markww:
I think this is ready for another look when you have a moment. I know it's not exactly critical path, but I'd like to get it in so it doesn't decay.

bgogul:
> Is this PR still happening?

It is waiting on me to resolve the discussion about the behavior of new checks.

markww:
This PR has been long abandoned.

Issue #226, opened 2021-11-18
Add a test to GCB to detect changes to docker image used in RBE


bgogul:
The docker image required for Remote Build Execution is specified in `raksha/gcb/rbe/configs/path/config/BUILD`:
https://github.com/google-research/raksha/blob/8fdfef31685e27f1a5101e6b10bae23e83a0f44f/gcb/rbe/configs/path/config/BUILD#L45

The GCB CI also rebuilds the docker image if the context used in [Dockerfile](https://github.com/google-research/raksha/blob/main/Dockerfile) changes. See
https://github.com/google-research/raksha/blob/8fdfef31685e27f1a5101e6b10bae23e83a0f44f/cloudbuild.yaml#L9

If a new image is build, it is also pushed to a gcp artifacts repository:
https://github.com/google-research/raksha/blob/8fdfef31685e27f1a5101e6b10bae23e83a0f44f/cloudbuild.yaml#L21

If the docker image changes due to a PR, we should detect and report that it is different from the one used for RBE.

bgogul:
The GCP image for `ubuntu:focal` seems to be update quite frequently and therefore, our image changes even if we did not make any local changes. We should freeze the ubuntu image to fixed release first:
https://github.com/google-research/raksha/blob/8fdfef31685e27f1a5101e6b10bae23e83a0f44f/Dockerfile#L1

bgogul:
Ah, on second thought it may not be a good idea to freeze it as we won't get security patches.

PR #271, opened 2022-01-21
Make presubmit enforce standard formatting on Skylark files and fix Skylark file formatting.

markww:


markww:
I've updated this PR to add a check enforcing formatting with `buildifier` in the presubmit check. You can see it actually failing the build due to the formatting being wrong in the build for `22acbb3`. This is a big enough change that I think it could use another review @bgogul , even though I know you already approved this once.

Dockerfile
```diff
@@ -31,6 +31,10 @@ ENV JAVA_HOME /usr/lib/jvm/default-java
 RUN wget https://github.com/bazelbuild/bazelisk/releases/download/v1.10.1/bazelisk-linux-amd64 \
     -O /usr/local/bin/bazelisk && chmod a+x /usr/local/bin/bazelisk
 
+# Set up buildifier
```
bgogul, line 34: I think we will need to update the image the rbe uses. You could do it in a follow up CL. https://github.com/google-research/raksha/blob/6f701da6e7ebd40aff235f3b0c6531485014187c/gcb/rbe/configs/path/config/BUILD#L45

Related issue #226

This still caught the error in presubmit because we don't use RBE for buildifer?

WORKSPACE
```diff
@@ -179,10 +179,71 @@ http_archive(
...
+http_archive(
+    name = "com_google_protobuf",
+    sha256 = "9b4ee22c250fe31b16f1a24d61467e40780a3fbb9b91c3b65be2a376ed913a1a",
+    strip_prefix = "protobuf-3.13.0",
+    urls = [
+        "https://github.com/protocolbuffers/protobuf/archive/v3.13.0.tar.gz",
+    ],
+)
+
+load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
```
bgogul, line 238: Don't we already import protobuf? See https://github.com/google-research/raksha/blob/6f701da6e7ebd40aff235f3b0c6531485014187c/WORKSPACE#L22

gcb/rbe/configs/path/cc/BUILD
```diff
@@ -38,7 +38,10 @@ filegroup(
 
 filegroup(
     name = "compiler_deps",
-    srcs = glob(["extra_tools/**"], allow_empty = True) + [":builtin_include_directory_paths"],
+    srcs = glob(
```
bgogul, line 41: Ah, I thought these were new config changes. It is just formatting chagnes from BUILDIFIER?

(This file was autogenerated from rbe helper scripts. )

src/analysis/souffle/BUILD
```diff
@@ -38,14 +38,14 @@ gen_souffle_cxx_code(
 gen_souffle_cxx_code(
     name = "taint_cxx_for_test",
     src = "taint.dl",
+    for_test = True,
```
bgogul, line 41: what is this?

bgogul (approved):
Mostly clarification questions. Update the title and description of the PR as the scope of change has increased/

src/analysis/souffle/BUILD
markww: `buildifier` just moved `for_test` from the end of the rule to the beginning. In an earlier PR, in which I split `gen_souffle_cxx_code` out from `souffle_cc_library`, I replaced `testonly` with `for_test` in the rule parameters, as this new parameter simultaneously sets `testonly` on subrules and sets a conditional compilation parameter to include or not include Datalog used only for tests.

gcb/rbe/configs/path/cc/BUILD
markww: Yes, this is just the automatic application of `buildifier`'s fixes.

WORKSPACE
markww: Good call, got rid of this.

Dockerfile
markww: Updated the build image.

markww: I think we do use RBE for buildifier. But I think it caught the error in presubmit because building the image had a cache miss upon trying to fetch the result of the buildifier line and built the image from scratch from there. While still correct, that is slow. Upon updating the hash, it does appear to finish building the image much more quickly.

PR #278, opened 2022-01-24
DLIR to Souffle Emitter

aferr:
This implements a pass that generates a Seouffle program as a string from a DLIR program. This pass also generates just the necessarily declarations for the predicates that are used.


aferr:
I might also take a look at using `std::transform` and `std::accumulate` for some of this.

aferr:
Actually, scratch that thought about trying `std::transform` and `std::accumulate`. It turns out `std::accumulate` and `std::reduce` are C++20 features and AFAICT `std::transform` has neither an aesthetic or performance improvement in this case.

aferr:
Tests are not there yet, but because of the timezone difference, it could be useful to get early review on this.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    std::string EmitAssertionInner(DLIRCondAssertion cond_assertion) {
+      std::string lhs = EmitPredicate(cond_assertion.lhs());
+      std::vector<std::string> rhs_translated;
+      for(auto arg: cond_assertion.rhs()) {
+        rhs_translated.push_back(EmitPredicate(arg));
+      }
+      std::string rhs = EmitVecSeparated(rhs_translated, ", ");
+      return std::move(lhs) + " :- " + std::move(rhs) + ".";
```
markww, line 94: When combining multiple strings together like this, just use `absl::StrCat`. https://abseil.io/docs/cpp/guides/strings#abslstrcat

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+      std::string args = "";
+      bool is_first = true;
+      for(auto elm : predicate.args()) {
+        if(!is_first) { args += ", "; } else { is_first = false; }
+        args += elm + ": symbol";
+      }
+      return ".decl " + predicate.name() + "("
+        + args + ")";
+    }
+
```
markww, line 129: This looks like another place where `absl::StrJoin` would be a better choice.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+      }
+      return EmitVecSeparated(body_translated, "\n");
+    }
+
+    std::string EmitOutputs(DLIRProgram program) {
+      std::string ret;
+      for (auto out : program.outputs()) {
+        ret += ".output " + out + "\n";
+      }
+      return ret;
```
markww, line 116: Use `StrJoin` with a custom formatter instead: https://abseil.io/docs/cpp/guides/strings#abslstrjoin-for-joining-elements-within-a-string

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+
+    std::string EmitDeclarations() {
+      std::vector<std::string> ret;
+      for(auto decl : declarations_) {
+        ret.push_back(EmitDeclaration(decl));
+      }
+      return EmitVecSeparated(ret, "\n");
+    }
+
+    std::unordered_set<Predicate, Predicate::HashFunction> declarations_;
```
markww, line 138: use an Abseil set, probably `flat_hash_set`, instead.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+    std::string EmitProgram(DLIRProgram program) {
+      SouffleEmitter emitter;
+      std::string body = emitter.EmitProgramBody(program);
+      std::string outputs = emitter.EmitOutputs(program);
+      std::string decls = emitter.EmitDeclarations();
+      return std::move(body) 
+        + "\n" 
+        + std::move(outputs) 
+        + "\n" 
+        + std::move(decls);
```
markww, line 37: Use `absl::StrCat` here instead of multiple concatenations. Also, you can avoid the need for `std::move` if you never give the values names in the first place; you could just not make the `body`, `outputs`, and `decls` locals and inline them directly into the call to `StrCat`.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+      return ret;
+    }
+
+    std::string EmitPredicate(Predicate predicate) {
+      // Whenever a new predicate is encountered, it is added to the set of 
+      // declarations (which does not include duplicates because it is a set).
+      declarations_.insert(PredToDeclaration(predicate));
+      return (predicate.sign() == kNegated ? "!" : "")
+          + predicate.name()
+          + "(" + EmitVecSeparated(predicate.args(), ", ") + ")";
```
markww, line 80: Use `absl::StrCat` here instead of multiple string concatenations.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+    // there are no duplicate delcarations (which would otherwise happen
+    // whenever a predicate is referenced more than once with different
+    // arguments).
+    // To prevent instances of negated and non-negated uses of the predicate 
+    // from generating two declarations, the sign here is always positive.
+    Predicate PredToDeclaration(Predicate predicate) {
+      std::vector<std::string> decl_args;
+      for( int i=0; i < predicate.args().size(); i++ ) {
+        decl_args.push_back("x" + std::to_string(i));
+      }
```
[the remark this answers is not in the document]
aferr: fixed.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,143 @@
...
+      return Predicate(predicate.name(), predicate.args(), kPositive);
+    }
+
+    std::string EmitVecSeparated(std::vector<std::string> vec,
+        std::string sep) {
+      std::string ret = "";
+      bool is_first = true;
+      for(auto elm : vec) {
+        if( !is_first ) { ret += sep; } else { is_first = false; }
```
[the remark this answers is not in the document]
aferr: fixed.

src/ir/auth_logic/souffle_emitter.h
aferr: fixed

src/ir/auth_logic/souffle_emitter.h
aferr: fixed

src/ir/auth_logic/souffle_emitter.h
aferr: fixed

src/ir/auth_logic/souffle_emitter.h
aferr: fixed

src/ir/auth_logic/souffle_emitter.h
aferr: fixed

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,126 @@
...
+    }
+    return absl::StrJoin(body_translated, "\n");
+  }
+
+  std::string EmitOutputs(DLIRProgram program) {
+    std::string ret;
+    for (auto out : program.outputs()) {
+      absl::StrAppend(&ret, absl::StrCat(".output ", out, "\n"));
+    }
+    return ret;
```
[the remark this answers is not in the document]
aferr: OK

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,126 @@
...
+                        absl::StrJoin(args_with_types, ", "), ")");
+  }
+
+  std::string EmitDeclarations() {
+    std::vector<std::string> ret;
+    for (auto decl : declarations_) {
+      ret.push_back(EmitDeclaration(decl));
+    }
+    return absl::StrJoin(ret, "\n");
+  }
```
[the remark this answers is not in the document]
aferr: done.

src/ir/auth_logic/souffle_emitter_build_stub.cc
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2022 The Raksha Authors
```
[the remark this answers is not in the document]
aferr: done/

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,126 @@
...
+
+  std::string EmitDeclaration(Predicate predicate) {
+    std::vector<std::string> args_with_types;
+    for (auto elm : predicate.args()) {
+      args_with_types.push_back(absl::StrCat(elm, ": symbol"));
+    }
+    return absl::StrCat(".decl ", predicate.name(), "(",
+                        absl::StrJoin(args_with_types, ", "), ")");
+  }
```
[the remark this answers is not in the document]
aferr: ok

src/ir/auth_logic/souffle_emitter_test.cc
```diff
@@ -0,0 +1,255 @@
...
+              Predicate("foo", {"bar", "baz"}, kPositive))))));
+    }
+
+    TEST(EmitterTestSuite, SimpleTest) {
+      std::string expected =
+R"(says_foo(TestPrincipal, bar, baz).
+grounded_dummy(dummy_var).
+
+.decl says_foo(x0: symbol, x1: symbol, x2: symbol)
+.decl grounded_dummy(x0: symbol))";
```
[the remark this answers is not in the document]
aferr: Why does this matter?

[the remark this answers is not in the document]
markww: My email seems to indicate that you might have asked why this matters, although I don't see that now. Historically, Unix systems have expected text files to end with a trailing newline and have been unhappy in various ways if they are not there.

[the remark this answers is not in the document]
aferr: Right. Edited just because I looked up why it might matter. That does sound valid! Thanks for pointing it out.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,122 @@
...
+    for (int i = 0; i < predicate.args().size(); i++) {
+      decl_args.push_back(absl::StrCat("x", std::to_string(i)));
+    }
+    return Predicate(predicate.name(), decl_args, kPositive);
+  }
+
+  std::string EmitPredicate(const Predicate& predicate) {
+    // Whenever a new predicate is encountered, it is added to the set of
+    // declarations (which does not include duplicates because it is a set).
+    declarations_.insert(PredToDeclaration(predicate));
```
markww, line 61: Definitely `CHECK` that you don't have a collision here on `Predicate` name.

src/ir/auth_logic/ast.h
```diff
@@ -53,6 +53,14 @@ class Predicate {
   const std::vector<std::string>& args() const { return args_; }
   Sign sign() const { return sign_; }
 
+  // < operator is needed for btree_set, which is only used for declarations.
+  // Since declarations are uniquely defined by the name of the predicate,
+  // this implementation that just uses < on the predicate names should be
+  // sufficent in the context where it is used.
+  bool operator<(const Predicate& otherPredicate) const {
+    return this->name() < otherPredicate.name();
+  }
```
markww, line 62: This is a perfectly fine implementation of `operator<`. but why do you need order on predicates? Why not just use an unordered `flat_hash_set`?

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  std::string EmitDeclarations() {
+    return absl::StrJoin(declarations_, "\n",
+                         [this](std::string* out, auto decl) {
+                           return absl::StrAppend(out, EmitDeclaration(decl));
+                         });
+  }
+
+  absl::btree_set<Predicate> declarations_;
```
markww, line 117: Why a `btree_set` instead of a `flat_hash_set`?

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,122 @@
...
+    SouffleEmitter emitter;
+    std::string body = emitter.EmitProgramBody(program);
+    std::string outputs = emitter.EmitOutputs(program);
+    std::string decls = emitter.EmitDeclarations();
+    return absl::StrCat(std::move(body), "\n", std::move(outputs), "\n",
+                        std::move(decls), "\n");
+  }
+
+ private:
+  SouffleEmitter() : declarations_(absl::btree_set<Predicate>()){};
```
markww, line 39: I'm pretty sure `declarations_`'s default ctor creates an empty set as you would expect. Therefore, you can just write this as `declarations_` or omit it altogether and just have `SouffleEmitter() { }` or (I think) even `SouffleEmitter() = default;`.

markww (commented):
Looks good mostly, but my big question is why you're using a `btree_set` instead of `flat_hash_set`.

src/ir/auth_logic/souffle_emitter.h
aferr: For testing purposes, I need to make sure that the order of declarations is consistent and does not change from one run to the next. With a `flat_hash_set` (which is what it was before) a consistent order is not guaranteed (or at least it wasn't once I changed the hash to the abseil one instead of the the role-my-own hash). (I guess another solution is to keep it as a `flat_hash_set` and generate a vector from the set and then order that vector at the end, but that seems just worse than this).

src/ir/auth_logic/ast.h
aferr: see other thread about `btree_set` and order of things.

src/ir/auth_logic/souffle_emitter.h
markww: It's better on some dimensions, worse on others. Using a `btree_set`, as you have here, is very concise. Asymptotically, it is the same as using a `flat_hash_set` and doing a big sort at the end; `btree_set` does `O(log(N))` insertion for each of the `N` elements you insert and can naively print the contents in order, while a `flat_hash_map` would do `O(1)` insertion an then have to do an `O(N * log(N))` sort at the end.

The difference between the two is that `btree_set` has to do an online sort as the elements are inserted, while the `sort` at the end with the `flat_hash_set` is offline, after all elements are known. In general, offline algorithms do much better than online algorithms, even if their complexity looks the same.

The code for performing the sort would be pretty easy too; it would look something like this:

```C++
std::vector elements_vec(declarations_.size());
elements_vec.insert(
  elements_vec.end(), std::make_move_iterator(declarations.begin()), std::make_move_iterator(declarations.end()));
std::sort(elements_vec);
```

In the end, I think that the conciseness of your current solution and the fact that this isn't likely to be performance critical means that the `absl::btree_set` is good enough to stay. But it's not "just worse", it's actually a pretty close consideration. I'd say leave it and just add a comment above `declarations_` saying that online ordering is only needed for tests. That way, if we're looking for ways to speed this up in the future, we can more easily remember that we have this knob to turn.

markww (approved):
LGTM other than commenting the `btree_set`.

src/ir/auth_logic/souffle_emitter.h
aferr: OK you are right. That's better. I did it that way.

src/ir/auth_logic/ast.h
aferr: resolving because it's back to `flat_hash_set`

src/ir/auth_logic/souffle_emitter.h
markww: Looks good :+1:

src/ir/auth_logic/ast.h
```diff
@@ -53,6 +55,35 @@ class Predicate {
...
+    return true;
+  }
+
+  // < operator is needed for btree_set, which is only used for declarations.
+  // Since declarations are uniquely defined by the name of the predicate,
+  // this implementation that just uses < on the predicate names should be 
+  // sufficent in the context where it is used.
+  bool operator< (const Predicate& otherPredicate) const {
+    return this->name() < otherPredicate.name();
+  }
```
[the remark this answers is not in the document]
aferr: opened https://github.com/google-research/raksha/issues/364

src/ir/auth_logic/ast.h
```diff
@@ -53,6 +55,35 @@ class Predicate {
   const std::vector<std::string>& args() const { return args_; }
   Sign sign() const { return sign_; }
 
+  template <typename H>
+  friend H AbslHashValue(H h, const Predicate& p) {
+    return H::combine(std::move(h), p.name(), p.args(), p.sign());
+  }
+    // Equality is also needed to use a Predicate in a flat_hash_set
```
bgogul, line 62: nit: formatting

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -78,16 +77,17 @@ DLIRAssertion LoweringToDatalogPass::SpokenAttributeToDLIR(
   Principal prin_y(FreshVar());
 
   // This is `speaker says Y PredX`
-  Predicate generated_lhs = PushPrincipal("says_", prin_y, main_predicate);
-
-  Predicate y_can_act_as_x("canActAs",
-                           {prin_y.name(), attribute.principal().name()},
-                           Sign::kPositive);
+  Predicate y_predX = AttributeToDLIR(Attribute(prin_y, attribute.predicate()));
```
bgogul, line 80: Why is the logic changing here? Or, is it just some refactoring?

src/ir/auth_logic/ast.h
```diff
@@ -53,6 +55,35 @@ class Predicate {
...
+  bool operator==(const Predicate& otherPredicate) const {
```
bgogul, line 63: do we have a test for this?

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -0,0 +1,126 @@
...
+  // be used when generating declarations of the predicates. It replaces
+  // argument names with generic numbered ones. It is applied
+  // to predicates before they are added to the set of declarations so that
+  // there are no duplicate delcarations (which would otherwise happen
+  // whenever a predicate is referenced more than once with different
+  // arguments).
+  // To prevent instances of negated and non-negated uses of the predicate
+  // from generating two declarations, the sign here is always positive.
+  Predicate PredToDeclaration(const Predicate& predicate) {
+    std::vector<std::string> decl_args;
```
bgogul, line 51: It looks like you can use `MapIter`?

src/ir/auth_logic/datalog_ir.h
```diff
@@ -48,6 +50,7 @@ class DLIRAssertion {
   using DLIRAssertionVariantType = std::variant<Predicate, DLIRCondAssertion>;
   explicit DLIRAssertion(DLIRAssertionVariantType value)
       : value_(std::move(value)) {}
+  const DLIRAssertionVariantType& GetValue() const { return value_; }
```
bgogul, line 53: nit: I still don't like this. We should definitely create a visitor. :)

bgogul (commented):
Some minor questions and comments.

src/ir/auth_logic/datalog_ir.h
aferr: we have https://github.com/google-research/raksha/issues/345

src/ir/auth_logic/souffle_emitter.h
aferr: ok I did that and also used it in one other place.

src/ir/auth_logic/ast_equality_test.cc
```diff
@@ -0,0 +1,64 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#include "src/common/testing/gtest.h"
+#include "src/ir/auth_logic/ast.h"
+
+namespace raksha::ir::auth_logic {
+
+TEST(AstEqualityTestSuite, Test1) {
```
bgogul, line 22: Change test name to `PredicateTest` and also indicate what is testing. e.g., 
`TEST(PredicateTest, EqualReturnsForIdenticalPredicates)`

Similarly for other TESTs.

src/ir/auth_logic/ast_equality_test.cc
```diff
@@ -0,0 +1,64 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
bgogul, line 2: Copyright year.

This should also go into a file predicate_test.cc

src/ir/auth_logic/ast_equality_test.cc
aferr: fixed

src/ir/auth_logic/ast_equality_test.cc
aferr: fixed

PR #280, opened 2022-01-25
Include gtest-death-test in common/testing/gtest.h

bgogul:


markww:
I think death tests are a good idea, but is this import necessary? It looks like I'm already using death tests in this repo:

https://github.com/google-research/raksha/blob/942a2f8b99e01826609ba45048cac1d901f98209/src/ir/system_spec_test.cc#L41

bgogul:
> I think death tests are a good idea, but is this include necessary? It looks like I'm already using death tests in this repo:
> 
> https://github.com/google-research/raksha/blob/942a2f8b99e01826609ba45048cac1d901f98209/src/ir/system_spec_test.cc#L41

Ah. For some reason, I got an error that `EXPECT_DEATH` was not defined when I added a new death test. Let me see why that was the case and circle back.

bgogul:
@markww You were right. This is not needed. Not sure why I saw that error in the first place. :)

PR #281, opened 2022-01-25
Add a reference counter class for intrusive ptrs.

bgogul:
Adds a counter class for reference counting in an intrusive ptr implementation.

### DIFF
--- a/src/utils/ref_counter.h
+++ b/src/utils/ref_counter.h
@@ -16,14 +16,13 @@
 #ifndef SRC_UTILS_REF_COUNTER_H_
 #define SRC_UTILS_REF_COUNTER_H_
 
-#include <atomic>
 #include <cstddef>
 
 #include "src/common/logging/logging.h"
 
 namespace raksha {
 
-template <bool thread_safe>
+// Note that this is not a thread-safe class.
 class RefCounter {
  public:
   RefCounter() : count_(0) {}
@@ -32,16 +31,14 @@ class RefCounter {
   operator size_t() const { return count_; }
 
   // Adds the given increment and returns the old value.
-  size_t FetchAdd(size_t increment,
-                  std::memory_order order = std::memory_order_seq_cst) {
+  size_t FetchAdd(size_t increment) {
     size_t old = count_;
     count_ += increment;
     return old;
   }
 
   // Subtracts the given increment and returns the old value.
-  size_t FetchSub(size_t decrement,
-                  std::memory_order order = std::memory_order_seq_cst) {
+  size_t FetchSub(size_t decrement) {
     CHECK(count_ >= decrement) << "FetchSub can decrement below zero.";
     size_t old = count_;
     count_ -= decrement;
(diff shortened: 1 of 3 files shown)


src/utils/ref_counter.h
```diff
@@ -0,0 +1,58 @@
...
+#define SRC_UTILS_REF_COUNTER_H_
+
+#include <atomic>
+#include <cstddef>
+
+#include "src/common/logging/logging.h"
+
+namespace raksha {
+
+template <bool thread_safe>
```
markww, line 26: It looks like the thread safety isn't being used right now; is it necessary?

src/utils/ref_counter.h
```diff
@@ -0,0 +1,58 @@
...
+class RefCounter {
+ public:
+  RefCounter() : count_(0) {}
+  RefCounter(size_t init) : count_(init) {}
+
+  operator size_t() const { return count_; }
+
+  // Adds the given increment and returns the old value.
+  size_t FetchAdd(size_t increment,
+                  std::memory_order order = std::memory_order_seq_cst) {
```
markww, line 36: Is sequential consistency necessary, or can we get away with acqrel? Also, do we need this when we haven't implemented the atomicity portion?

src/utils/ref_counter.h
```diff
@@ -0,0 +1,58 @@
...
+    size_t old = count_;
+    count_ += increment;
+    return old;
+  }
+
+  // Subtracts the given increment and returns the old value.
+  size_t FetchSub(size_t decrement,
+                  std::memory_order order = std::memory_order_seq_cst) {
```
markww, line 44: See other question about memory order param.

src/utils/ref_counter.h
```diff
@@ -0,0 +1,58 @@
...
+    CHECK(count_ >= decrement) << "FetchSub can decrement below zero.";
+    size_t old = count_;
+    count_ -= decrement;
+    return old;
+  }
+
+ private:
+  // TODO: Create thread-safe version.
+  mutable size_t count_;
```
markww, line 53: Why is this `mutable`? Are we expecting to manipulate it through a `const` pointer?

src/utils/ref_counter.h
bgogul: Just wanted to future proof the interface. I guess it is not hard to add this later. Will remove and simplify.

src/utils/ref_counter.h
bgogul: Not used at all. Just a place holder when we add atomicity.

src/utils/ref_counter.h
bgogul: Just a place holder. Will remove and simplify.

src/utils/ref_counter.h
bgogul: Actually, this is not needed here.

bgogul:
@markww, simplified this. PTAL.

markww (approved):
LGTM

PR #282, opened 2022-01-25
Add `licenses(["notice"])` to BUILD files; add presubmit check for this. Also presubmit-check that BUILD files have copyright headers.

markww:


gcb/presubmit.sh
```diff
@@ -62,6 +62,16 @@ bazel_run() {
...
 
+# Verify that all Bazel build files have a license rule. Exclude files in the
+# gcb directory.
+# BUILD_FILES_WITHOUT_LICENSES=$(fgrep -L 'licenses(' $(find . -name BUILD) | grep -v gcb)
+# NUM_BUILD_FILES_WITHOUT_LICENSES=$(echo "$BUILD_FILES_WITHOUT_LICENSES" | wc -w)
+# if [ "$NUM_BUILD_FILES_WITHOUT_LICENSES" -ne 0 ]; then
+#   echo "Found build files without licenses:"
+#   echo "$BUILD_FILES_WITHOUT_LICENSES"
+#   return 1
+# fi
```
bgogul, line 73: stale?

Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apt-get update && apt-get install -y \
     flex \
     g++ \
     git \
+    grep \
```
bgogul, line 14: stale?

bgogul (approved):
LGTM except for code left from debugging.

gcb/presubmit.sh
markww: No, I'm trying to figure out why the build was failing. I committed the presubmit code commented out to check to see whether it was the cause of the problem. Still waiting on the results there.

bgogul: Oh, you want to make this part of the presubmit check? Sounds good.

third_party/arcs/examples/BUILD
```diff
@@ -15,3 +15,5 @@ arcs_manifest_proto(
     src = "VisionPipeline.arcs",
     deps = [],
 )
+
+licenses(["notice"])
```
bgogul, line 19: Why are the `licenses(` line at the end? Why not add after the copyright block on the top?

bgogul (commented):
nit about `licenses` at the end vs start of the file.

third_party/arcs/examples/BUILD
markww: I used a shell script to add `licenses` to each file to reduce toil. I was hoping that `buildifier` would fix the placement, but it did not. I can take another pass at the shell script to put it at the top; even if `buildifier` doesn't fix that, it's at least a more natural place to find it.

markww: Actually, there are only 18 files. I'll just do it by hand.

gcb/presubmit.sh
markww: Thank you for your help debugging this, this is in and working (on my Docker image anyway).

Dockerfile
markww: Nope, needed for presubmit checks.

third_party/arcs/examples/BUILD
markww: Fixed this. Also fixed the copyright headers in these files and added a presubmit check for this for `BUILD` files only; we should add it for more files in the future.

markww:
Hi Gogul, I added just another little snippet of presubmit checking in this PR for copyright headers, as I had to fix a bunch of them when fixing `licenses` rules. Take a quick look if you can.

gcb/presubmit.sh
```diff
@@ -59,9 +59,40 @@ bazel_run() {
   fi
 }
 
+check_build_files_for_licenses_rule() {
```
bgogul, line 62: Do we need an explicit `return 0` at the end of these functions?

Also, how much additional time do these add to the presubmit runtime?

bgogul (approved):

gcb/presubmit.sh
markww: It appears that we do not, but I can add those in a follow up PR just to make sure. I didn't measure the runtime of `presubmit.sh`. My guess is that it's probably faster than the `buildifier` check which actually does parsing, but I do not know for sure.

bgogul: > It appears that we do not, but I can add those in a follow up PR just to make sure. I didn't measure the runtime of `presubmit.sh`. My guess is that it's probably faster than the `buildifier` check which actually does parsing, but I do not know for sure.

OK. I see the presubmit times are between 12 and 25 mins. I think the increase in time is because we keep rebuilding the docker images between the different PRs (due to Dockerfile changes). Once all the Dockerfile changes are in, I hope it comes back to ~7 mins average that we have. Otherwise, we will need to revisit this.

PR #283, opened 2022-01-26
Add Hedron's compile commands extractor to WORKSPACE.

bgogul:
This allows us to generate `compile_commands.json` which is required to integrate `clangd` into vscode.

### DIFF
--- a/external
+++ b/external
@@ -1,0 +1,1 @@
+bazel-out/../../../external
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ bazel-testlogs
 **/*.swp
 .vscode/**
 .clwb/**
+.cache/clangd/**
+compile_commands.json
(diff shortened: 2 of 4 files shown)


external
```diff
@@ -0,0 +1 @@
+bazel-out/../../../external
```
markww, line 1: What is this?

markww (approved):
This mostly looks good. I just don't understand the `external` file.

external
bgogul: Just adds a symbolic link to `external` folder of bazel.

bgogul: Just adds a symbolic link to `external` folder of bazel. See https://github.com/google-research/raksha/tree/compile_commands%40bgogul

markww: OK, as long as it works, that's fine.

PR #284, opened 2022-01-26
Add a RefCounted helper for intrusive_ptr.

bgogul:
This is a class from which other classes can inherit if they want to be reference counted.

### DIFF
--- a/src/utils/ref_counted.h
+++ b/src/utils/ref_counted.h
@@ -47,12 +47,12 @@ class RefCounted {
     CHECK(count_ != 0) << "Reference count is already zero.";
     size_t old = count_.FetchSub(1);
     if (old == 1) {
-      delete static_cast<T*>(this);
+      delete this;
     }
   }
 
  protected:
-  ~RefCounted() {}
+  virtual ~RefCounted() {}
 
  private:
   // So that it can call `Retain` and `Release`.
--- a/src/utils/ref_counted.h
+++ b/src/utils/ref_counted.h
@@ -1,0 +1,65 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


src/utils/ref_counted.h
```diff
@@ -0,0 +1,65 @@
...
+  RefCounted& operator=(const RefCounted&) = delete;
+
+ private:
+  void Retain() { count_.FetchAdd(1); }
+
+  void Release() {
+    CHECK(count_ != 0) << "Reference count is already zero.";
+    size_t old = count_.FetchSub(1);
+    if (old == 1) {
+      delete static_cast<T*>(this);
```
markww, line 50: I don't understand how this is supposed to work. In this context, `this` is the `RefCounted` class, which contains within it only a `RefCounter` field. If `T` has a non-trivial dtor, the field alignments would be all off.

The usual design I see for refcounted objects is that the ref counting header (which I assume this is) assumes that it is allocated directly before an instance of its payload (T in this context). So I assume that the fields should be more like:

```
mutable RefCounter count_;
T payload_;
```

and the `Release()` method would simply do:

```
delete this;
```

bgogul: Classes that need to reference count are supposed to inherit from this class (or define their own reference counting mechanism with `RefCountManager`). See https://github.com/google-research/raksha/pull/286/files#diff-50fe78aea3d9579425f207389cd8a115614f50882838f6fed0aa7e83f496023d

For this to work correctly, the destructor needed to be virtual.  I have fixed it now. The type `T` will be provided as part of the template. In any case, I removed the `static_cast` to avoid confusion. Given that destructor is virtual, I am not sure if `static_cast` helps at all.

src/utils/ref_counted.h
```diff
@@ -0,0 +1,65 @@
...
+
+ private:
+  void Retain() { count_.FetchAdd(1); }
+
+  void Release() {
+    CHECK(count_ != 0) << "Reference count is already zero.";
+    size_t old = count_.FetchSub(1);
+    if (old == 1) {
+      delete this;
+    }
```
markww, line 51: In this context, you check whether the returned value `old` from `FetchSub` was 1, as you would if we were doing an atomic subtract operation. However, the `CHECK` assertion looks directly at the `count_` value, which does not show the same level of respect to the possibly-atomic abstraction. This wouldn't cause any problems, because we aren't doing anything atomic, but we should probably be consistent as to whether we are upholding an abstraction as if the count could be atomic.

markww (approved):
Small abstraction quibble, looks good otherwise.

src/utils/ref_counted.h
bgogul: Good point. Actually, this check turned out to be redundant as `FetchSub` would already throw an error. I removed this CHECK and updated the test to reflect the new error message.

PR #286, opened 2022-01-26
Implementation of an intrusive_ptr class. 

bgogul:


src/utils/intrusive_ptr.h
```diff
@@ -0,0 +1,81 @@
...
+//-----------------------------------------------------------------------------
+#ifndef SRC_UTILS_INTRUSIVE_PTR_H_
+#define SRC_UTILS_INTRUSIVE_PTR_H_
+
+#include "src/utils/ref_counted.h"
+
+namespace raksha {
+
+template <typename T>
+class intrusive_ptr {
```
markww, line 24: Not wild about the name `intrusive_ptr`; it's long, and it doesn't describe *why* we are intruding into the pointed-at memory. Can we make it something like `refct_ptr` instead?

src/utils/intrusive_ptr.h
```diff
@@ -0,0 +1,81 @@
...
+    return *this;
+  }
+
+  T& operator*() const {
+    CHECK(ptr_ != nullptr) << "Dereferencing a nullptr!";
+    return *ptr_;
+  }
+
+  T* operator->() const { return ptr_; }
+  T* get() const { return ptr_; }
```
markww, line 48: I wonder whether we need `get()`. It seems like it could undermine our proof that no uncounted copies of the reference exist. It looks like this is mostly used for comparing equality in tests; can we make a `PointersEqual` method instead so as not to exflitrate the contained pointer?

src/utils/intrusive_ptr_test.cc
```diff
@@ -0,0 +1,155 @@
...
+  auto value = std::make_unique<RefCountedType>();
+  EXPECT_EQ(value->count(), 0);
+
+  RefCountedTypePtr ptr(value.get());
+  EXPECT_EQ(value->count(), 1);
+
+  RefCountedTypePtr moved = std::move(ptr);
+  EXPECT_EQ(value->count(), 1);
+  EXPECT_EQ(ptr.get(), nullptr);
+}
```
markww, line 81: This pattern you use in this file of having a `unique_ptr` and an intrusive pointer point to the same value is very easy to misunderstand. I think it could use some comments documenting its behavior.

src/utils/intrusive_ptr_test.cc
```diff
@@ -0,0 +1,155 @@
...
+  assigned = ptr;
+  EXPECT_EQ(value->count(), 2);
+  EXPECT_EQ(ptr.get(), assigned.get());
+}
+
+TEST(IntrusivePtrTest, GetReturnsSavedPointer) {
+  auto value = std::make_unique<RefCountedType>();
+  RefCountedTypePtr ptr(value.get());
+  EXPECT_EQ(ptr.get(), value.get());
+}
```
markww, line 102: Will this test double free the `RefCountedType`?

src/utils/intrusive_ptr_test.cc
```diff
@@ -0,0 +1,155 @@
...
+
+TEST(IntrusivePtrTest, DereferenceReturnsSavedObject) {
+  auto value = std::make_unique<RefCountedType>();
+  RefCountedTypePtr ptr(value.get());
+  EXPECT_EQ(&(*ptr), value.get());
+}
```
markww, line 108: Similarly, is there a double free here?

src/utils/intrusive_ptr_test.cc
```diff
@@ -0,0 +1,155 @@
...
+      "Dereferencing a nullptr!");
+}
+
+TEST(IntrusivePtrTest, ArrowOperatorProvidesAccessToSavedPtr) {
+  auto value = std::make_unique<RefCountedType>("NewValue");
+  RefCountedTypePtr ptr(value.get());
+  EXPECT_EQ(value->id(), "NewValue");
+  ptr->set_id("UpdatedValue");
+  EXPECT_EQ(value->id(), "UpdatedValue");
+}
```
markww, line 125: Double free?

src/utils/intrusive_ptr_test.cc
```diff
@@ -0,0 +1,155 @@
...
+  auto value = std::make_unique<RefCountedType>();
+  EXPECT_EQ(value->count(), 0);
+
+  {
+    RefCountedTypePtr ptr(value.get());
+    EXPECT_EQ(value->count(), 1);
+    // RefCountedTypePtr is destroyed at the end of this scope.
+  }
+  EXPECT_EQ(value->count(), 0);
+}
```
markww, line 152: Double free?

src/utils/intrusive_ptr.h
bgogul: It is somewhat standard term used for objects with embedded reference counts: https://www.boost.org/doc/libs/1_63_0/libs/smart_ptr/intrusive_ptr.html

I will add a comment that this is inspired by boost (which is true).

src/utils/intrusive_ptr.h
bgogul: Good point.  Let me address this in a follow up PR as I want to add pointer comparators anyway. 

In any case, even if we remove the `get` method, we can still get the raw pointer value using `&*p`.

src/utils/intrusive_ptr_test.cc
bgogul: Good point. It is not good idea to leave bad code examples in the test. I refactored this without `unique_ptr` and documented why we are using raw pointer for tests.

src/utils/intrusive_ptr_test.cc
bgogul: Refactored the code, but this would not have caused a double free earlier. In the previous version,  the custom implementation of `Retain` and `Release` only kept track of ref counts and did not the object. It is changed now.

src/utils/intrusive_ptr_test.cc
bgogul: See other comment.

src/utils/intrusive_ptr_test.cc
bgogul: See other comment.

src/utils/intrusive_ptr_test.cc
bgogul: No. See reply other comment.

src/utils/intrusive_ptr.h
markww: ok sounds good

PR #290, opened 2022-01-31
Update GRPC to 4.1.1.

markww:
Initially, I upgraded this in hopes that 4.1.1 removed the need for the
`protos` attribute, allowing us to just put proto files in `deps` again.
That is not what happened. Committing the update anyway because it's
good to keep our dependencies fresh.


bgogul (approved):
Wait for #293 to land and run tests.

PR #291, opened 2022-01-31
Add "taint.dl" to the export list in //src/analysis/souffle package.

markww:


bgogul (approved):
We had issues with GCB. Not sure if it is fixed yet.

PR #295, opened 2022-01-31
Made operator.h use `Block`'s full defn rather than a forward decl.

markww:
The `Operator` class contains within it a member, `implementation_`,
that uses a `std::unique_ptr` pointing to `Block`. Previously, this
header had only a forward declaration for `Block`. This is bad, because
`unique_ptr` needs `Block`'s full definition to get its dtor and to
perform `sizeof` upon that class. This change makes `operator.h` include
the full `block.h` header, giving it access to the full definition of
`Block`.

### DIFF
--- a/src/ir/data_decl.h
+++ b/src/ir/data_decl.h
@@ -19,10 +19,7 @@
 #include <vector>
 
 #include "absl/container/flat_hash_map.h"
-#include "src/ir/operation.h"
-#include "src/ir/operator.h"
 #include "src/ir/types/type.h"
-#include "src/ir/value.h"
 
 namespace raksha::ir {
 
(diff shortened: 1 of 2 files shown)


markww:
Although this PR appears to build correctly in this environment, I suspect it is misguided. Specifically, I think that there is an inclusion cycle between block.h, data_decl.h, operation.h and operator.h . This is probably why `Block` was a forward decl in the first place. Closing this for now.

markww:
I was able to get rid of some unnecessary headers in `data_decl.h` that may fix the cycle. Reopening this PR.

bgogul (approved):
Looks good. I might get rid of implementation completely in coming PRs though. :)

Issue #43, opened 2021-07-22
Includes in dl files should not use relative paths


bgogul:
Ideally, we want to follow the convention of specifying the full path starting from WORKSPACE root. e.g., `#include "src/analysis/souffle/taint.dl"` instead of `#include "../../taint.dl"`.

Supporting full paths requires adding the bazel workspace root as an additional include path in `souffle_cc_library`rule.

harsha-mandadi:
Fixed in PR #297

PR #297, opened 2022-02-01
Update datalog import path to start with raksha root

harsha-mandadi:
Issue #43


markww:
It looks like the `#include` directives were updated correctly, but this PR doesn't build. I think it's likely that the `-I` include roots provided to Souffle need to be updated as well.

I think that may just come down to changing [these lines](https://github.com/google-research/raksha/blob/main/build_defs/souffle.bzl#L51) to just include the Raksha repo root.

harsha-mandadi:
Earlier, build was failing because generate_datalog_program_test.sh was comparing the diff between newly generated datalog(with absolute include paths) and existing ones [here](https://github.com/google-research/raksha/pull/297/files#diff-96430a802b6152fc905847c42247abfed0c3b0f0dc1ebdd209453839cffccf80). Did not make any include root changes to souffle to pass tests.

bgogul (approved):
LGTM. 

@markww Hope this does not interfere with import into Google.

markww:
@bgogul It should be fine. We're already re-rooting `#include`s in C++ sources and headers, I've expanded the pattern to match `.dl` files as well.

markww (approved):
LGTM

PR #301, opened 2022-02-02
More fixing of unnecessary and transitive build dependencies.

markww:


bgogul (approved):
Thank you for doing this!

PR #322, opened 2022-02-03
Add `-fexceptions` for all C++ rules building Souffle code in Raksha.

markww:
Souffle uses exceptions in its generated code. This causes problems when
importing Raksha into Google's internal codebase, because Google's build
process assumes `-fno-exceptions` by default. This change adds
`-fexceptions` explicitly everywhere where Souffle code is built to deal
with that issue.


build_defs/souffle.bzl
```diff
@@ -100,6 +100,7 @@ def souffle_cc_library(
         srcs = [src],
         testonly = testonly,
         copts = [
+            "-fexceptions",
```
bgogul, line 103: What will happen internally? We will get rid of this flag?

bgogul (approved):

build_defs/souffle.bzl
markww: Actually, quite the opposite. We need `-fexceptions` internally, but it is the default in most C++ builds. So `-fexceptions` is usually a restatement of the default outside of Google, but will cause exceptions to be allowed inside of Google.

bgogul: SG.

PR #323, opened 2022-02-03
Turn off header modules everywhere where exceptions are turned on.

markww:
Using exceptions with header modules causes problems within Google's
internal code base. This PR turns off header modules everywhere where
`-fexceptions` is used.

### DIFF
--- a/build_defs/raksha.bzl
+++ b/build_defs/raksha.bzl
@@ -119,6 +119,7 @@ def policy_check(name, dataflow_graph, auth_logic, expect_failure = False, visib
             "-fexceptions",
             "-Iexternal/souffle/src/include/souffle",
         ],
+        features = ["-use_header_modules"],  # Incompatible with -fexceptions.
         linkopts = ["-pthread"],
         deps = [
             "@souffle//:souffle_include_lib",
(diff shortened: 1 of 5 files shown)


build_defs/raksha.bzl
```diff
@@ -119,6 +119,7 @@ def policy_check(name, dataflow_graph, auth_logic, expect_failure = False, visib
             "-fexceptions",
             "-Iexternal/souffle/src/include/souffle",
         ],
+        features = ["-use_header_modules"],  # Incompatible with -fexceptions.
```
bgogul, line 122: What are `header_modules`? Also, why does the comment say `incompatible with -fexceptions` even though we have `-fexceptions` in the copts above?

bgogul (approved):
Looks good modulo using `-fexceptions` along with `-use_header_modules`.

build_defs/raksha.bzl
markww: I think the comment and/or the `features` syntax is confusing. Combining `-` with `use_header_modules` means "turn off the `use_header_modules` feature". I'll fix the comments.

markww: `header_modules` are a form of precompiled header. Google compiles all things (including precompiled headers) with `-fno-exceptions` by default and Clang gets very upset when mixing binaries that have `-fno-exceptions` and `-fexceptions`. Setting `-use_header_modules` causes the default C++ behavior of pasting the header code into the TU to be used which fixes the incompatibility.

markww: Clarified the comment.

PR #325, opened 2022-02-03
Fix a memory leak in intrusive ptr test.

markww:
One of the clauses of the intrusive_ptr test FactoryConstructionWorks
constructed a pointer with an initial count of `10`. However, this means
that when the pointer goes out of scope, it just decreases the count by
`1`. This change manually decreases it `10` more times so it will be
deleted.

### DIFF
--- a/src/utils/intrusive_ptr_test.cc
+++ b/src/utils/intrusive_ptr_test.cc
@@ -25,10 +25,11 @@ namespace {
 
 class RefCountedType {
  public:
-  RefCountedType(std::string id = "<unknown>", unsigned int initial_count = 0)
-      : id_(id), count_(initial_count) {}
+  RefCountedType(std::string id = "<unknown>", bool some_flag = false)
+      : id_(id), some_flag_(some_flag), count_(0) {}
 
   unsigned int count() const { return count_; }
+  bool some_flag() const { return some_flag_; }
   void set_id(std::string id) { std::swap(id_, id); }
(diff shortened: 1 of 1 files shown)


src/utils/intrusive_ptr_test.cc
```diff
@@ -155,6 +155,12 @@ TEST(IntrusivePtrTest, FactoryConstructionWorks) {
   EXPECT_EQ(another_ptr->count(), 11);
   EXPECT_EQ(another_ptr->id(), "AnotherValue");
   EXPECT_NE(ptr.get(), another_ptr.get());
+  // Note that another_ptr going out of scope would only cause the count to be
```
bgogul, line 158: This test was just to check that non-trivial constructors work. I think it was a bad idea to use `count` in the constructor.  Instead of doing this, can you just move `initial_count` from the constructor and change/remove this part of the test completely? I don't think we are using `initial_count` for tests in any way.

markww: Sounds good, I think that's much better as well.

bgogul: BTW, I meant remove `initial_count` from the constructor completely. I just realized that it typed it as "move `initial_count`" from constructor. :)

markww: No prob, that is what I have done :)

bgogul (approved):

PR #328, opened 2022-02-04
Add an Attribute class and tests to hold constants in IR.

bgogul:


src/ir/attribute.h
```diff
@@ -0,0 +1,80 @@
...
+  std::string ToString() const override { return value_; }
+
+ private:
+  StringAttribute(absl::string_view value)
+      : AttributeBase(Kind::kString), value_(value) {}
+
+  std::string value_;
+};
+
+using NamedAttributeMap = absl::flat_hash_map<std::string, Attribute>;
```
markww, line 76: It certainly makes sense to look up `Attribute`s by name, but isn't the name part of the identity of the `Attribute`? Perhaps the string should be placed inside the `Attribute` as well as in the hash map.

Also, shouldn't the value be an `intrusive_ptr` pointing to an `Attribute` rather than a value `Attribute`?

src/ir/attribute.h
```diff
@@ -0,0 +1,80 @@
...
+  virtual ~AttributeBase() {}
+
+  virtual Kind kind() const { return kind_; }
+  virtual std::string ToString() const = 0;
+
+ private:
+  Kind kind_;
+};
+
+using Attribute = intrusive_ptr<const AttributeBase>;
```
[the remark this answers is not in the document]
markww: Alternatively, should `Attribute`s just be value types and, to make the `Attribute`s more palatably naively-copyable, should we intern the strings in a registry and make `StringAttribute` contain an `absl::string_view`?

src/ir/attribute.h
```diff
@@ -0,0 +1,80 @@
...
+class IntAttribute : public AttributeBase {
+ public:
+  static Attribute Create(int value) {
+    return Attribute(new IntAttribute(value));
+  }
+
+  int value() const { return value_; }
+
+  std::string ToString() const override {
+    return absl::StrFormat("%d", value_);
```
markww, line 51: Should this be just the bare integer value, or should this have some indication of the fact that this is an attribute? Perhaps something like `"int_attr: %d"`?

src/ir/attribute_test.cc
```diff
@@ -0,0 +1,51 @@
...
+//----------------------------------------------------------------------------
+#include "src/ir/attribute.h"
+
+#include "src/common/testing/gtest.h"
+
+namespace raksha::ir {
+namespace {
+
+class IntAttributeTest
+    : public testing::TestWithParam<std::pair<int, const char*>> {};
```
markww, line 24: Why `const char *` here? Isn't `absl::string_view` preferred in these cases?

src/ir/attribute_test.cc
```diff
@@ -0,0 +1,51 @@
...
+INSTANTIATE_TEST_SUITE_P(IntAttributeTest, IntAttributeTest,
+                         testing::Values(std::make_pair(10, "10"),
+                                         std::make_pair(-30, "-30"),
+                                         std::make_pair(0, "0")));
+
+class StringAttributeTest : public testing::TestWithParam<absl::string_view> {};
+
+TEST_P(StringAttributeTest, KindAndToStringWorks) {
+  const auto& value = GetParam();
+  auto int_attribute = StringAttribute::Create(value);
```
markww, line 42: I think this is a copy-paste error. Should be `str_attribute` or similar?

src/ir/attribute.h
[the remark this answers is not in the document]
bgogul: I would like to use  uniqued attributes as well. Using `intrusive_ptr` is just a temporary workaround.  That is why I did the following
 - Have `Create` methods that within each derived Attribute class to later make it return uniqued attributes.
 - Named this `Attribute` and not `AtttributePtr`. We will replace `Attribute` later with class that holds a uniqued-id (or a pointer). We have uses of `->` from an `Attribute` instance, but the compiler will catch them.

src/ir/attribute.h
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+ private:
+  IntAttribute(int value) : AttributeBase(Kind::kInt), value_(value) {}
+
+  int value_;
```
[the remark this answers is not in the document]
bgogul: Good point. Just made it align  with the `IntAttribute` name. Will change this to `int64_t` and also Int64Attribute.

src/ir/attribute.h
bgogul: This map will be used within an operation to have named attributes. e.g., `ReadFieldOp`, we will have a map like this:
```
{"schema": "MyStruct", "field": "a"}
```

src/ir/attribute_test.cc
bgogul: Will change it to absl::string_view.

src/ir/attribute_test.cc
bgogul: yes. will fix it.

src/ir/attribute.h
bgogul: I plan to use this to dump the IR. Let us leave it as is for now and can make changes later as needed?

bgogul:
@markww Addressed style comments and replied to your questions. PTAL.

src/ir/attribute.h
markww: Sounds good. Also, just remembered `Attribute` is an `intruisve_ptr<AttributeBase>`, so the other part of my comment doesn't matter.

markww (approved):
LGTM, using `intrusive_ptr` as an intermediate step makes sense.

PR #329, opened 2022-02-04
Cleanup DataDecl API and add unit tests.

bgogul:


src/ir/data_decl_test.cc
```diff
@@ -0,0 +1,93 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
markww, line 2: Is this the right year?

markww (approved):
LGTM other than copyright year nit.

src/ir/data_decl_test.cc
bgogul: Will fix it.

PR #330, opened 2022-02-04
Simplify operator API and add unit test.

bgogul:
Let us start with the simplest definition and iterate it.

### DIFF
--- a/src/ir/operator.h
+++ b/src/ir/operator.h
@@ -29,7 +29,7 @@ class Operator {
  public:
   Operator(absl::string_view name) : name_(name) {}
 
-  const std::string& name() const { return name_; }
+  absl::string_view name() const { return name_; }
 
  private:
   std::string name_;
--- a/src/ir/operator_test.cc
+++ b/src/ir/operator_test.cc
@@ -23,7 +23,7 @@ namespace {
 class OperatorTest : public testing::TestWithParam<absl::string_view> {};
 
 TEST_P(OperatorTest, NameIsCorrectlySet) {
-  const auto& name = GetParam();
+  absl::string_view name = GetParam();
   Operator op(name);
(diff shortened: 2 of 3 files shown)


src/ir/operator.h
```diff
@@ -30,22 +27,14 @@ namespace raksha::ir {
 // operations.
 class Operator {
  public:
-  Operator(absl::string_view name, std::unique_ptr<Block> implementation)
-      : name_(name), implementation_(std::move(implementation)) {
-    CHECK(implementation_ != nullptr);
-  }
+  Operator(absl::string_view name) : name_(name) {}
 
   const std::string& name() const { return name_; }
```
markww, line 32: Should this just return an `absl::string_view`?

src/ir/operator_test.cc
```diff
@@ -0,0 +1,36 @@
...
+
+#include "src/common/testing/gtest.h"
+
+namespace raksha::ir {
+namespace {
+
+class OperatorTest : public testing::TestWithParam<absl::string_view> {};
+
+TEST_P(OperatorTest, NameIsCorrectlySet) {
+  const auto& name = GetParam();
```
markww, line 26: Since the parameter is a `string_view`, which is naively copyable, shouldn't this just be `absl::string_view` instead of this `auto` ref (which I belive will become `const absl::string_view &name`)?

markww (approved):
LGTM other than some nits.

src/ir/operator.h
bgogul: Ah, I meant to change this use, but looks like I left it out. Will do now.

src/ir/operator_test.cc
bgogul: Yes, will do. Was an autopilot. :)

PR #333, opened 2022-02-05
Add Attribute class to keep constants in IR.

bgogul:


bgogul:
This was reviewed in #328, which I accidentally merged it into non-main branch.

PR #338, opened 2022-02-07
Utility for creating SSA names when printing IR.

bgogul:
e.g, this will be useful to print values like `OperationResult(op, "value")` as `%39.value`.

### DIFF
--- a/src/ir/ssa_names.h
+++ b/src/ir/ssa_names.h
@@ -1,0 +1,62 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+#ifndef SRC_IR_SSA_NAMES_H_
+#define SRC_IR_SSA_NAMES_H_
+
+#include "absl/container/flat_hash_map.h"
+
+namespace raksha::ir {
+
+class Block;
+class Operation;
+
+class SSANames {
+ public:
+  using ID = int;
+  SSANames() : next_block_id_(0), next_operation_id_(0) {}
+
+  // Disable copy (and move) semantics.
+  SSANames(const SSANames &) = delete;
+  SSANames &operator=(const SSANames &) = delete;
+
+  ID GetOrCreateID(const Operation &operation) {
+    return GetOrCreateIDHelper(operation_ids_, next_operation_id_, &operation);
+  }
+
+  ID GetOrCreateID(const Block &block) {
+    return GetOrCreateIDHelper(block_ids_, next_block_id_, &block);
+  }
+
+ private:
+  template <typename T, typename U>
+  static ID GetOrCreateIDHelper(T &ids_table, ID &next_id, U entity) {
+    auto find_result = ids_table.find(entity);
+    if (find_result == ids_table.end()) {
+      ID result = next_id++;
+      ids_table.insert({entity, result});
+      return result;
+    }
+    return find_result->second;
+  }
+  ID next_block_id_;
+  absl::flat_hash_map<const Block *, ID> block_ids_;
+  ID next_operation_id_;
+  absl::flat_hash_map<const Operation *, ID> operation_ids_;
+};
+
+}  // namespace raksha::ir
+
+#endif  // SRC_IR_CONTEXT_H_
--- a/src/ir/ssa_names.h
+++ b/src/ir/ssa_names.h
@@ -23,13 +23,13 @@ namespace raksha::ir {
 class Block;
 class Operation;
 
-class SSANames {
+class SsaNames {
  public:
   using ID = int;
-  SSANames() {}
+  SsaNames() {}
   // Disable copy (and move) semantics.
-  SSANames(const SSANames &) = delete;
-  SSANames &operator=(const SSANames &) = delete;
+  SsaNames(const SsaNames &) = delete;
+  SsaNames &operator=(const SsaNames &) = delete;
 
   ID GetOrCreateID(const Operation &operation) {
     return operation_ids_.GetOrCreateID(&operation);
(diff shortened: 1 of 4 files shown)


src/ir/ssa_names.h
```diff
@@ -0,0 +1,62 @@
...
+    return GetOrCreateIDHelper(operation_ids_, next_operation_id_, &operation);
+  }
+
+  ID GetOrCreateID(const Block &block) {
+    return GetOrCreateIDHelper(block_ids_, next_block_id_, &block);
+  }
+
+ private:
+  template <typename T, typename U>
+  static ID GetOrCreateIDHelper(T &ids_table, ID &next_id, U entity) {
```
markww, line 45: You could imagine getting rid of a template parameter by writing the type of `ids_table` as `absl::flat_hash_map<const U *, ID>`, but you can ignore that if you think you might want to use types of `id_table`s other than `flat_hash_map`s.

src/ir/ssa_names.h
```diff
@@ -0,0 +1,62 @@
...
+    auto find_result = ids_table.find(entity);
+    if (find_result == ids_table.end()) {
+      ID result = next_id++;
+      ids_table.insert({entity, result});
+      return result;
+    }
+    return find_result->second;
```
markww, line 52: I think this might be more concisely and idiomatically written like this:

```C++
auto insert_result = ids_table.insert({entity, next_id});
if (insert_result.second) ++next_id;
return insert_result.first->second;
```

`insert` does nothing if a key already exists, returning a `pair` with the iterator to the preexisting entry in `first` and `false` in second. If a key doesn't already exist, it returns the iterator to the new entry in `first` and `true` in second. That means the following:

1. Regardless of whether insertion happened, the ID you want is in the value field of the entry pointed to by the iterator, or `insert_result.first->second`.

2. Whether you need to increment the ID is exactly the same as whether an entry was inserted, or `insert_result.second`.

As an added bonus, you look up in the map once instead of twice, which is a fun optimization, but which probably doesn't matter unless this is in the inner loop somewhere.

Also, preincrement is preferred over postincrement.

src/ir/ssa_names.h
```diff
@@ -0,0 +1,62 @@
...
+  }
+  ID next_block_id_;
+  absl::flat_hash_map<const Block *, ID> block_ids_;
+  ID next_operation_id_;
+  absl::flat_hash_map<const Operation *, ID> operation_ids_;
```
markww, line 57: What if, instead of `N` pairs of associated IDs and maps, you introduced another class called `ItemIdManager` of the following form:

```C++
template<class Item>
class ItemIdManager {
 public:
   ID GetOrCreateIdHelper(T entity) {
     /*...*/
   }
 private:
  ID next_id_;
  absl::flat_hash_map<const Item *, ID> item_ids_;
};
```

Then the `SSANames` class could be written as follows:

```
class SSANames {
  private:
    ItemIdManager<Block> block_id_manager_;
    OperationIdManager<Operation> operation_id_manager_;
};
```

src/ir/ssa_names.h
```diff
@@ -0,0 +1,62 @@
...
+#define SRC_IR_SSA_NAMES_H_
+
+#include "absl/container/flat_hash_map.h"
+
+namespace raksha::ir {
+
+class Block;
+class Operation;
+
+class SSANames {
```
markww, line 26: I think the C++ standard dictates that this should be `SsaNames`. See definition of "word" in the [general naming rules](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules) and the caps-case directive in [type naming rules](https://google.github.io/styleguide/cppguide.html#Type_Names).

src/ir/ssa_names_test.cc
```diff
@@ -0,0 +1,58 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
markww, line 2: Copyright year.

src/ir/value.h
```diff
@@ -107,6 +107,8 @@ class StoredValue {
   StoredValue(const StoredValue&) = default;
   StoredValue& operator=(const StoredValue&) = default;
 
+  const Storage& storage() const { return *storage_; }
```
markww, line 110: This change seems a little off-theme for this PR.

src/ir/value.h
```diff
@@ -124,8 +126,8 @@ class Any {};
 // A class that represents a data value.
 class Value {
  public:
-  Value(value::BlockArgument arg): value_(std::move(arg)) {}
-  Value(value::BlockResult arg): value_(std::move(arg)) {}
+  Value(value::BlockArgument arg) : value_(std::move(arg)) {}
+  Value(value::BlockResult arg) : value_(std::move(arg)) {}
```
markww, line 130: This change also seems a little off-theme for this PR.

src/ir/ssa_names.h
bgogul: SSA abbreviates Static Single Assignment. So, I think this is OK.

src/ir/ssa_names.h
bgogul: Simplified with `IDManager<T>`

src/ir/ssa_names.h
bgogul: Done. Used the idiomatic code in `IDManager`.

src/ir/ssa_names.h
bgogul: Good idea. Done.

src/ir/value.h
bgogul: Yeah, this change is needed for the code to compile. Otherwise, we get an unused field error. All of these get cleaned up in the child PRs though.

src/ir/value.h
bgogul: reverted this.

bgogul:
@markww Addressed your comments. PTAL.

src/ir/ssa_names.h
markww: I know, but the general naming rules say that abbreviations count as words for the purpose of naming. You can override me if you want because I think it looks better as-is, but I think that's the standard.

src/ir/ssa_names.h
```diff
@@ -0,0 +1,63 @@
...
+   public:
+    ID GetOrCreateID(const T *entity) {
+      auto insert_result = item_ids_.insert({entity, next_id_});
+      if (insert_result.second) ++next_id_;
+      return insert_result.first->second;
+    }
+
+   private:
+    ID next_id_;
+    absl::flat_hash_map<const T *, ID> item_ids_;
```
markww, line 54: You could actually get rid of `next_id_` and have next ID be implicitly described by `item_ids_.size()`. But that's not super important.

markww: It would be nice, though, because it would get rid of that `if` deciding whether to increment on the `insert` result.

src/ir/ssa_names.h
bgogul: I went back and forth on this. In the end `SsaNames` seemed easier to parse. So, I went with it.

src/ir/ssa_names.h
bgogul: good idea.done.

PR #339, opened 2022-02-07
Add tests for Storage class

bgogul:


src/ir/storage_test.cc
```diff
@@ -0,0 +1,52 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
markww, line 2: Copyright year

src/ir/storage_test.cc
```diff
@@ -0,0 +1,52 @@
...
+  EXPECT_EQ(input_storage.type().type_base().kind(),
+            types::TypeBase::Kind::kPrimitive);
+
+  const types::Schema& schema = type_factory_.RegisterSchema("TestStruct", {});
+  Storage output_storage("output", type_factory_.MakeEntityType(schema));
+  EXPECT_EQ(output_storage.name(), "output");
+  const types::TypeBase& type_base = output_storage.type().type_base();
+  EXPECT_EQ(type_base.kind(), types::TypeBase::Kind::kEntity);
+  EXPECT_EQ(static_cast<const types::EntityType&>(type_base).schema().name(),
+            "TestStruct");
```
markww, line 43: If this were to fail due to `type_base` not being downcastable to `EntityType`, we would get a segfault due to dereferencing `nullptr`. Perhaps it would be better to split this into a statement that does the downcast, then `ASSERT_NOT_NULL` (or whether the appropriate assertion is) then the `EXPECT_EQ`.

markww (approved):
LGTM. Copyright year and a small suggestion on how to make the test fail in a more friendly way in one case.

src/ir/storage_test.cc
bgogul: I intended to capture it by the check `EXPECT_EQ(type_base.kind(), types::TypeBase::Kind::kEntity)` in the previous line? However, I think you are right that that previous check should actually be `ASSERT_EQ` so that we don't segfault on `42`. Will change it.

PR #340, opened 2022-02-07
Add an IRContext class

bgogul:
This is needed to keep the global state (like registered operators) necessary for building the IR.

### DIFF
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -1,0 +1,58 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


src/ir/ir_context_test.cc
```diff
@@ -0,0 +1,54 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
markww, line 2: Copyright year

src/ir/ir_context.h
```diff
@@ -0,0 +1,58 @@
...
+                          << ins_res.first->first << "'.";
+    return *ins_res.first->second.get();
+  }
+
+  // Returns the Operator with a particular name. If there is no
+  // operator with that name, returns nullptr.
+  const Operator *GetOperator(absl::string_view operators_name) const {
+    auto find_res = operators_.find(operators_name);
+    return (find_res != operators_.end()) ? find_res->second.get() : nullptr;
+  }
```
markww, line 47: It might be nicer, null-safety wise, to have a `HasOperator` call and have `GetOperator` assert if called when `HasOperator` is false. Then `GetOperator` can always return `const Operator &` and we don't need to worry about the nullity aspect.

markww (approved):
Looks good, but I had a suggestion about factoring that might make the null-safety of the interface a little nicer.

src/ir/ir_context.h
bgogul: Done. Added an `IsRegisteredOperator` and changed `GetOperator` to return `const&`.

bgogul:
@markww addressed your comments. PTAL.

markww (approved):
LGTM

PR #342, opened 2022-02-07
Add test for Operation class.

bgogul:
This PR adds some test to make sure the `Operation` class is initialized correctly and `ToString` generates the correct output.


src/ir/operation_test.cc
```diff
@@ -0,0 +1,107 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
markww, line 2: Copyright year

markww (approved):
LGTM other than copyright year.

PR #344, opened 2022-02-08
Update `manifest.proto` to latest version.

markww:
The version of `manifest.proto` that we had in the Raksha repo was not
the latest version. This change updates it, and eliminates the
`optional` field from types that it eliminates. All other changes were
append-only, and we can handle them later.

### DIFF
--- a/third_party/arcs/proto/manifest.proto
+++ b/third_party/arcs/proto/manifest.proto
@@ -26,7 +26,7 @@ message RecipeProto {
 
 // Recipe-level representation of a store.
 message HandleProto {
-  // Next Id: 5
+  // Next Id: 6
   enum Fate {
     UNSPECIFIED = 0;
     CREATE = 1;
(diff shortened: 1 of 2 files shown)


third_party/arcs/proto/manifest.proto
```diff
@@ -109,9 +113,11 @@ message TypeProto {
     SingletonTypeProto singleton = 7;
     CountTypeProto count = 8;
     ListTypeProto list = 9;
+    NullableTypeProto nullable = 13;
   }
-  bool optional = 10;
   RefinementExpressionProto refinement = 11;
+  repeated AnnotationProto annotations = 12;
```
bgogul, line 119: does the parser still work correctly?

bgogul (approved):
looks good modulo question about parser

third_party/arcs/proto/manifest.proto
markww: It appears to on the tests that we are running. I think that all comes down to the fact that most of the changes in `manifest.proto` have been append-only changes, only adding proto fields. That means that the parser, using an old version of the proto, just fills in a subset of the possible fields.

The one exception is the `optional` field. Because we aren't relying upon the `optional` field in our code nor in any of the examples that we are currently compiling, we seem to be getting away with that one as well.

PR #362, opened 2022-02-10
Remove Cargo.lock

aferr:


aferr:
@markww @bgogul 

It looks like when the cargo crate is meant to be a standalone binary (which is the situation here), the recommended best practice is to include Cargo.lock in the git repository:

[source](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)

which says:

>If you’re building a non-end product, such as a rust library that other rust [packages](https://doc.rust-lang.org/cargo/appendix/glossary.html#package) will depend on, put Cargo.lock in your .gitignore. If you’re building an end product, which are executable like command-line tool or an application, or a system library with crate-type of staticlib or cdylib, check Cargo.lock into git.

PR #366, opened 2022-02-11
Add block builder class and tests.

bgogul:


src/ir/block_builder.h
```diff
@@ -0,0 +1,84 @@
+//-----------------------------------------------------------------------------
+// Copyright 2021 Google LLC
```
markww, line 2: Copyright year.

src/ir/block_builder.h
```diff
@@ -0,0 +1,84 @@
...
+  // Adds an operation to the block and returns the operation.
+  const Operation& AddOperation(const Operator& op,
+                                NamedAttributeMap attributes,
+                                NamedValueMap inputs) {
+    std::unique_ptr<Operation> operation(new Operation(
+        block_.get(), op, std::move(attributes), std::move(inputs)));
+    const Operation* result = operation.get();
+    block_->operations_.push_back(std::move(operation));
+    return *result;
+  }
```
markww, line 58: Just to make the pointer never be constructed outside of its `unique_ptr` and move around less, this might be better phrased like this:

```C++
block->operations_.push_back(
  std::make_unique<Operation>(block_.get(), op, std::move(attributes), std::move(inputs));
return **block->operations_.back();
```

Not critical though.

src/ir/block_builder_test.cc
```diff
@@ -0,0 +1,185 @@
...
+namespace raksha::ir {
+namespace {
+
+class BlockBuilderTest : public testing::Test {
+ public:
+  BlockBuilderTest() {
+    context_.RegisterOperator(std::make_unique<Operator>("core.plus"));
+    context_.RegisterOperator(std::make_unique<Operator>("core.minus"));
+    context_.RegisterOperator(std::make_unique<Operator>("core.merge"));
+    context_.RegisterOperator(std::make_unique<Operator>("core.choose"));
```
markww, line 36: Probably not relevant to this PR, but seeing this usage is making me wonder if `Operator` should sent into `RegisterOperator` as a value and moved into the `unique_ptr` space via the move ctor. Would make it a little more concise to call `RegisterOperator`.

markww (approved):
LGTM. Left a few optional suggestions.

src/ir/block_builder.h
bgogul: argh, not again. I solemnly promise that I really created this file in 2021. (I actually did! :) )

Jokes aside, I will change the year.

src/ir/block_builder_test.cc
bgogul: I am not sure I quite follow. Are you saying an usage like this? `context_.RegisterOperator(Operator("core.plus"))`

markww: Yes, and then within `RegisterOperator`, we could have something like so:

```C++
void RegsiterOperator(Operator op) {
  auto insert_result = operator_map_.insert({op.name, std::make_unique<Operator>(std::move(op)));
  /*...*/
}
```

(The above may be slightly wrong, I didn't look at `RegisterOperator`'s decl or defn).

PR #371, opened 2022-02-11
Add a small comment to test copybara import.

bgogul:


markww (approved):
LGTM (But not really, because this is a test of Copybara).

markww:
Commenting this to prevent someone from accidentally hitting the big green merge button.

src/ir/data_decl_test.cc
```diff
@@ -20,6 +20,8 @@
 #include "src/ir/types/type.h"
 #include "src/ir/types/type_factory.h"
 
+// This is a test for DataDecl.
+
```
markww, line 24: Just trying to add a comment to block accidental merging.

PR #374, opened 2022-02-11
comparison operator for intrusive_ptr

harsha-mandadi:


src/utils/intrusive_ptr.h
```diff
@@ -41,6 +41,10 @@ class intrusive_ptr {
     this->Swap(other);
     return *this;
   }
+  
+  bool operator==(intrusive_ptr<T> other) {
+    return this->ptr_ == other.ptr_;
+  }
```
markww, line 47: `operator==` should be declared as taking a `const` pointer to `this`. This is done by putting a `const` modifier trailing the signature. Making no other changes, this would be written like so:

```C++
bool operator==(intrusive_ptr<T> other) const {
```

In addition, the use of the parameter as a value type rather than a const reference type introduces a copy. That would be fine (indeed, preferable!) for many types, but `intrusive_ptr` does a pointer dereference and modification on every copy, which makes this potentially have surprising effects. I'd say to change the parameter type to

```C++
const intrusive_ptr<T> &other
````

which is the more standard form for operator-overloading parameters anyway.

File: src/utils/intrusive_ptr_test.cc [hunk 2]
```diff
@@ -181,5 +181,20 @@ TEST(IntrusivePtrTest, DestructorReducesCount) {
   EXPECT_EQ(ptr->count(), 1);
 }
 
+TEST(IntrusivePtrTest, EqualReturnsForIdenticalPtrs) {
+  RefCountedTypePtr ptr = make_intrusive_ptr<RefCountedType>();
+
+  RefCountedTypePtr copy = ptr;
+  EXPECT_TRUE(ptr==copy);
```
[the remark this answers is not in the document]
harsha-mandadi: My initial draft was using EXPECT_EQ. However, I felt that EXPECT_TRUE is better for readability. If we want to be consistent in our code, I can make the changes.

src/utils/intrusive_ptr.h
harsha-mandadi: That makes sense. I will update the code.

src/utils/intrusive_ptr.h
```diff
@@ -75,10 +75,19 @@ class intrusive_ptr {
...
 }
 
+template <class T, class U>
+bool operator==(const intrusive_ptr<T> &lhs, const intrusive_ptr<U> &rhs) {
+  return lhs.get() == rhs.get();
+}
+
+template <class T, class U>
+bool operator!=(const intrusive_ptr<T> &lhs, const intrusive_ptr<U> &rhs) {
+  return !(lhs.get() == rhs.get());
```
bgogul, line 89: Invoke the `== operator` defined above: 

```suggestion
  return !(lhs == rhs);
```

src/utils/intrusive_ptr_test.cc
```diff
@@ -181,5 +186,26 @@ TEST(IntrusivePtrTest, DestructorReducesCount) {
...
+  EXPECT_EQ(ptr, another_copy);
+}
+
+TEST(IntrusivePtrTest, ComparePtrsOfDifferentTypes) {
+  intrusive_ptr<RefCountedTypeDerived> derived_ptr = make_intrusive_ptr<RefCountedTypeDerived>();
+  intrusive_ptr<RefCountedType> base_ptr(derived_ptr.get());
+  EXPECT_EQ(base_ptr, derived_ptr);
+}
+
+TEST(IntrusivePtrTest, UnEqualReturnsForDifferentPtrs) {
```
bgogul, line 205:
```suggestion
TEST(IntrusivePtrTest, NotEqualReturnsTrueForDifferentPtrs) {
```

src/utils/intrusive_ptr_test.cc
```diff
@@ -181,5 +186,26 @@ TEST(IntrusivePtrTest, DestructorReducesCount) {
   EXPECT_EQ(ptr->count(), 1);
 }
 
+TEST(IntrusivePtrTest, EqualReturnsForIdenticalPtrs) {
```
bgogul, line 189: Ideally, the test name should describe what expectation is being tested.

```suggestion
TEST(IntrusivePtrTest, EqualReturnsTrueForIdenticalPtrs) {
```

src/utils/intrusive_ptr_test.cc
```diff
@@ -46,6 +46,11 @@ class RefCountedType {
   unsigned int count_;
 };
 
+class RefCountedTypeDerived: public RefCountedType {
+ public:
+  RefCountedTypeDerived() {}
+};
```
bgogul, line 52: I _think_ you can remove the constructor: 

```suggestion
class RefCountedTypeDerived: public RefCountedType {};
```

bgogul (approved):
Looks good modulo some style issues. Please take care of them before merging this PR in.

Issue #321, opened 2022-02-03
Add constructor for intrusive_ptr that allows conversions between base and derived ptrs.


bgogul:
Suppose we have the following classes:

```
class Base {}

class Derived : Base {}
```
We should be able to initialize `intrusive_ptr<Base>` with an `intrusive_ptr<Derived>` instance.

PR #376, opened 2022-02-14
intrusive ptr constructor that can initialize intrusive_ptr<base_class> from intrusive_ptr<derived_class>

harsha-mandadi:
issue #321

### DIFF
--- a/src/utils/intrusive_ptr.h
+++ b/src/utils/intrusive_ptr.h
@@ -33,6 +33,10 @@ class intrusive_ptr {
   intrusive_ptr(intrusive_ptr<T>&& other) : ptr_(other.ptr_) {
     other.ptr_ = nullptr;
   }
+  
+  template <class X, std::enable_if_t<std::is_convertible<X *, T *>::value, bool> = true>
+  intrusive_ptr(const intrusive_ptr<X>& other) : ptr_(Retain(other.get())) {}
+  
   ~intrusive_ptr() { Release(ptr_); }
 
   intrusive_ptr<T>& operator=(intrusive_ptr<T> other) {
--- a/src/utils/intrusive_ptr_test.cc
+++ b/src/utils/intrusive_ptr_test.cc
@@ -165,6 +165,19 @@ TEST(IntrusivePtrTest, FactoryConstructionWorks) {
   // All objects will be destroyed when the intrusive_pointers go out of scope.
 }
 
+TEST(IntrusivePtrTest, CanConvertBetweenConvertablePtrs){
+  auto derived_ptr = make_intrusive_ptr<RefCountedTypeDerived>();
+  EXPECT_EQ(derived_ptr->count(), 1);
+  intrusive_ptr<RefCountedType> base_ptr(derived_ptr);
+  EXPECT_EQ(derived_ptr.get(), base_ptr.get());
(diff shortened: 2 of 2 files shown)


src/utils/intrusive_ptr.h
```diff
@@ -33,6 +33,12 @@ class intrusive_ptr {
   intrusive_ptr(intrusive_ptr<T>&& other) : ptr_(other.ptr_) {
     other.ptr_ = nullptr;
   }
+  
+  template <class X, std::enable_if_t<std::is_convertible<X *, T *>::value, bool> = true>
+  intrusive_ptr(intrusive_ptr<X> other) : ptr_(other.get()) {
+    other.ptr_ = nullptr;
```
markww, line 39: This doesn't seem right. This is the behavior I would expect from a `move` constructor, where the new pointer "steals" the contents of the old pointer without incrementing the count.

This constructor's signature is closer to that of a copy constructor, and so it should increment the pointed-to reference counter.

In addition, if this is supposed to be a copy ctor, then the parameter type should be `const intrusive_ptr<X> &other`, so that the parameter points to the existing copy rather than silently making a copy, which would require altering the refcount.

src/utils/intrusive_ptr_test.cc
```diff
@@ -80,6 +85,13 @@ TEST(IntrusivePtrTest, CopyConstructorIncreasesCountAndSharesPtr) {
   RefCountedTypePtr copy = ptr;
   EXPECT_EQ(value->count(), 2);
   EXPECT_EQ(ptr.get(), copy.get());
+
+  auto derived_ptr = make_intrusive_ptr<RefCountedTypeDerived>();
+  EXPECT_EQ(derived_ptr->count(), 1);
+  intrusive_ptr<RefCountedType> base_ptr(derived_ptr);
+  EXPECT_EQ(derived_ptr.get(), base_ptr.get());
+  EXPECT_EQ(derived_ptr->count(), 2);
+  EXPECT_EQ(base_ptr->count(), 2);
```
bgogul, line 94: Move this to a separate test say `TEST(IntrusivePtrTest, CanConvertBetweenConvertiblePtrs)` or something like that.  

Also check that the underlying pointers are the same. `EXPECT_EQ(base_ptr, derived_ptr);`?

src/utils/intrusive_ptr_test.cc
```diff
@@ -46,6 +46,11 @@ class RefCountedType {
   unsigned int count_;
 };
 
+class RefCountedTypeDerived: public RefCountedType {
+ public:
+  RefCountedTypeDerived() {}
```
bgogul, line 51: I don't think you need to declare the constructor.

src/utils/intrusive_ptr_test.cc
```diff
@@ -165,6 +165,16 @@ TEST(IntrusivePtrTest, FactoryConstructionWorks) {
   // All objects will be destroyed when the intrusive_pointers go out of scope.
 }
 
+TEST(IntrusivePtrTest, CanConvertBetweenConvertablePtrs){
```
bgogul, line 168: minor typo:
```suggestion
TEST(IntrusivePtrTest, CanConvertBetweenConvertiblePtrs){
```

bgogul (approved):

Issue #375, opened 2022-02-14
[Authorization Logic // Rust] Fix frontend for constants vs symbolic IDs


aferr:
The grammar for authorization logic currently has just one rule for identifiers that is meant to cover: parameter names, variable argument names, and constant argument names. This rule looks like `ID : ('"')? [_a-zA-Z0-9/.#:]* ('"')?;`. 

Because all of these are lumped together, this places unneeded syntactic restrictions such as requiring an extra space in type declarations for predicates. It also likely allows more code to parse than should be allowed.

It would be better to more cleanly separate parameter/symbolic names from constants. This might instead look like:
```
id
    : '"' PARAM_NAME (NAME_EXTRA_CHARS PARAM_NAME)* '"'
    | PARAM_NAME
    ;
...
PARAM_NAME: [_a-zA-Z][_a-zA-Z0-9]*;
NAME_EXTRA_CHARS: [/.#:];
```

or something similar. However, antlr is pretty finicky about lexer rules that might overlap, so pushing this through is a bit more work than it sounds. We can get by with the syntax as it is for now, so this is an issue instead of a PR :)

aferr:
This is [fixed for the C++ version](https://github.com/google-research/raksha/blob/main/src/policy/auth_logic/AuthLogic.g4). Rust version is no longer relevant.

PR #377, opened 2022-02-14
Authorization Logic :: Type Declarations

aferr:
This implements type declarations for authorization logic. Type declarations are needed to support both:
https://b.corp.google.com/issues/219015959
and
https://b.corp.google.com/issues/219015796
for transparent release. The types supported are symbols, numbers, and principals. The syntax is directly borrowed from Souffle for now. An extra space is needed between type declarations and parameter names, but there is https://github.com/google-research/raksha/issues/375 to track fixing this eventually.


rust/tools/authorization-logic/src/souffle/souffle_emitter.rs
```diff
@@ -15,116 +15,81 @@
  */
 
 use crate::{ast::*, souffle::datalog_ir::*};
-use std::collections::HashSet;
 
-pub struct SouffleEmitter {
-    // This is a set of predicates that need to be declared. This set is
-    // populated by side effect while traversing the program body.
-    decls: HashSet<AstPredicate>,
```
aferr, line 23: Declarations used to be generated by keeping around a set of all the predicates that were encountered and then translating every parameter into a symbol.  This is removed now that there are type declarations.

bgogul:
> This implements type declarations for authorization logic. Type declarations are needed to support both: https://b.corp.google.com/issues/219015959 and https://b.corp.google.com/issues/219015796 for transparent release. 

Links to internal bugs don't make sense in open source. When you get a chance add a brief description of what this PR is about. I have some context and will review it in any case.

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
```diff
@@ -349,7 +492,15 @@ impl LoweringToDatalogPass {
         };
         dlir_assertions.append(&mut dlir_queries);
         dlir_assertions.push(dummy_assertion);
+
+        // type declarations
```
bgogul, line 496: Replicate the comment about how this should be called after body is emitted?

rust/tools/authorization-logic/src/souffle/souffle_emitter.rs
bgogul: Thanks for the heads up. In github, you can hide whitespace during review if needed.

bgogul (commented):
Mostly clarification questions and some minor suggestions. 

Haven't reviewed rust code for style issues.

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -226,8 +228,57 @@ fn construct_import(ctx: &ImportAssertionContext) -> AstImport {
...
+            AstType::SymbolType,
+        AuthLogicTypeContextAll::PrincipalTypeContext(ctx_prime) =>
+            AstType::PrincipalType,
+        _ => {
+            panic!("construct_type tried to build error");
+        }
+    }
+}
+
+fn construct_type_declaration(ctx: &TypeDeclarationContext) -> AstTypeDeclaration {
```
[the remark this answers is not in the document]
aferr: `astconstructionvisitor.rs` is the interface between the code that we have written and the parser that is generated by Antlr. This type is defined in code generated by antlr that is produced in `AUTH_LOGIC/src/parsing/antlr_gen`. 

(As you may have noticed, `astconstructionvisitor.rs` is technically not a visitor -- in general ANTLR backends can generate visitors, but the visitor generation for rust is buggy, so we can't use it in this rust code.)

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
aferr: ok

rust/tools/authorization-logic/test_inputs/conditions
```diff
@@ -1,3 +1,8 @@
+.decl fact1(x : symbol)
+.decl fact2(x : symbol)
+.decl cond1(x : symbol)
+.decl cond2(x : symbol, y : symbol)
+
```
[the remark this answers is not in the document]
aferr: I think these are going to stay mandatory even for the C++ version as long as we have any notion of types. For languages that have type inference, inference still relies on explicit declarations of types somewhere (for example they might appear in the function signature but not the function body). I don't yet see how type inference would make sense for a language like our authorization logic. So in short, I just think "these are mandatory going forward".

rust/tools/authorization-logic/test_inputs/typeDeclarations
```diff
@@ -0,0 +1,7 @@
+.decl someFact(x : symbol)
+.decl someOtherFact(x : symbol, y : number)
+.decl thirdFact(p : principal)
```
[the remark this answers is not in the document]
aferr: I think technically since this comment is on the frontend of the auth logic language it does not matter because the backend could still emit `Principal`... but perhaps it's best to keep it stylistically consistent anyway.

src/xform_to_datalog/authorization_logic_datalog_facts.cc
```diff
@@ -23,38 +23,29 @@
...
-      "says_canSay_hasTag",
-      "says_removeTag",
-      "says_canSay_removeTag",
-      "says_may",
-      "says_will",
-      "isAccessPath",
-      "isTag",
-      "isPrincipal"
-  };
+      "says_ownsTag", "says_ownsAccessPath", "says_hasTag",  "says_removeTag",
```
aferr, line 43: The only change in this file was that the _canSay_ relations are removed from this list because now auth logic can generate those with the types that the C++ side is expecting.

src/xform_to_datalog/datalog_facts_test.cc
```diff
@@ -154,9 +154,12 @@ says_hasTag("particle", "recipe.particle.out", owner, "tag") :- ownsAccessPath(o
 
 
 // Authorization Logic
-.decl grounded_dummy(x0: symbol)
-.decl says_cond1(x0: symbol, x1: symbol)
-.decl says_fact1(x0: symbol, x1: symbol)
+.type SomeType <: symbol
```
aferr, line 157: @bgogul Is there any way we can make the tests in `xform_to_datalog` any less brittle? I think I'll have to change all of these tests every time the authorization logic backend is changed.

aferr: There might not be... and even if there is maybe it isn't worth the effort. It might just be that the next PR is the last one where the backed for the rust implementation changes.

rust/tools/authorization-logic/Cargo.lock
```diff
@@ -2,6 +2,15 @@
 # It is not intended for manual editing.
 version = 3
 
+[[package]]
```
bgogul, line 5: I am still not sure why these changes are needed. Did you change your `Cargo.toml` file? If these changes are needed, can you split them off to a separate PR?

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -81,8 +81,19 @@ importAssertion
     : IMPORT principal SAYS ID
     ;
 
+// The name "type" would cause name collisions in the code that antlr generates
+authLogicType
+    : NUMBERTYPE #numberType
+    | PRINCIPALTYPE #principalType
```
bgogul, line 87: Any reason to keep this separate from `ID #customType`?

bgogul (approved):
Looks good except for the following caveats: 
 - Did not look closely at rust code for style. 
 - Split off Cargo.lock changes to a separate PR.

rust/tools/authorization-logic/Cargo.lock
aferr: I didn't change `Cargo.toml`, but some other earlier commit changes it without updating `Cargo.lock`. Sure I can make the update to `Cargo.lock` a separate PR.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: Not too many strong reasons at the moment. Some weak reasons though:
 - Extensibility --> extending the grammar of principals is one thing we might naturally do (add and/or on principals for example), and making this a builtin type could help with that. (A bit moot in this case since this is not the implementation we will keep around).
 - Principals can show up in more parts of the grammar than objects of the other types. For example in `canSay` or `canActAs`, so in the implementation when these need to be generated we can call PrincipalType instead of CustomType.

aferr: I'm assuming that the approval means that this is OK to leave for now? If you'd rather get rid of this LMK. We can also just keep this in mind for the C++ impl.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -81,8 +81,19 @@ importAssertion
     : IMPORT principal SAYS ID
     ;
 
+// The name "type" would cause name collisions in the code that antlr generates
+authLogicType
+    : NUMBERTYPE #numberType
+    | SYMBOLTYPE #symbolType
```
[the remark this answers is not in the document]
aferr: The tests on both the rust side and the C++ side both pass currently... I think I am a bit confused by your question. What change to the grammar are you anticipating? The `SYMBOLTYPE` is already removed and now just NUMBER, PRINCIPAL, and CUSTOMTYPE are there.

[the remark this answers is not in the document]
aferr: And `AccessPath` is a `CUSTOMTYPE`.

rust/tools/authorization-logic/Cargo.lock
aferr: fixed

PR #378, opened 2022-02-15
Decode the protos describing a source column from the SQL verifier.

markww:
This change sets up decoding the protos describing a source column from the SQL verifier. We expand the `SourceTableColumn` proto, which describes a column and the tags that the policy verifier statically assigns to it, to a `Storage` node indicating the column, a `StoredValue` indicating the value read from that column, and a number of `tag_claim` `Operation`s that "cast" the expected tags onto the value read from the column.

### DIFF
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -37,13 +37,13 @@ class IRContext {
 
   // Register an operator and return the stored operator instances.
   const Operator &RegisterOperator(std::unique_ptr<Operator> op) {
-    return operators_.RegisterNode(std::move(op), "operator");
+    return operators_.RegisterNode(std::move(op));
   }
 
   // Returns the Operator with a particular name. If there is no
   // operator with that name, fail.
   const Operator &GetOperator(absl::string_view operators_name) const {
-    return operators_.GetNode(operators_name, "operator");
+    return operators_.GetNode(operators_name);
   }
 
   // Returns true if the operator is registered with this context.
@@ -53,13 +53,13 @@ class IRContext {
 
     // Register an operator and return the stored operator instances.
   const Storage &RegisterStorage(std::unique_ptr<Storage> op) {
-    return storages_.RegisterNode(std::move(op), "storage");
+    return storages_.RegisterNode(std::move(op));
   }
 
   // Returns the Storage with a particular name. If there is no
   // operator with that name, fail.
   const Storage &GetStorage(absl::string_view operators_name) const {
-    return storages_.GetNode(operators_name, "storage");
+    return storages_.GetNode(operators_name);
   }
 
   // Returns true if the operator is registered with this context.
@@ -68,30 +68,34 @@ class IRContext {
   }
 
  private:
-  template<class NamedIrNode>
-  class NamedIrNodeRegistry {
+  // Names for the kinds of nodes we will manage with `NamedIRNodeRegistry`.
+  // These constants are used to work around the fact that string literals
+  // cannot directly be used as template arguments.
+  static constexpr char kOperatorName[] = "operator";
+  static constexpr char kStorageName[] = "storage";
+
+  template<class Node, const char *kNodeKindName>
+  class NamedIRNodeRegistry {
    public:
       // Register a node and return the stored instance.
-    const NamedIrNode &RegisterNode(std::unique_ptr<NamedIrNode> node,
-                                    absl::string_view node_kind_name) {
+    const Node &RegisterNode(std::unique_ptr<Node> node) {
         // Note: using node->name() here is ok because C++ list initialization
         // guarantees that evaluation of initializers earlier in the list
         // are sequenced before initializers later in the list.
         auto ins_res = nodes_.insert(
             {std::string(node->name()), std::move(node)});
         CHECK(ins_res.second)
-          << "Cannot register duplicate " << node_kind_name << " with name '"
+          << "Cannot register duplicate " << kNodeKindName << " with name '"
           << ins_res.first->first << "'.";
         return *ins_res.first->second.get();
     }
 
     // Returns the node with a particular name. If there is no
     // node with that name, fail.
-    const NamedIrNode &GetNode(
-        absl::string_view node_name, absl::string_view node_kind_name) const {
+    const Node &GetNode(absl::string_view node_name) const {
       auto find_res = nodes_.find(node_name);
       CHECK(find_res != nodes_.end())
-          << "Looking up an unregistered " << node_kind_name
+          << "Looking up an unregistered " << kNodeKindName
           <<  " '" << node_name << "'.";
       return *find_res->second;
     }
@@ -101,13 +105,13 @@ class IRContext {
       return nodes_.find(node_name) != nodes_.end();
     }
    private:
-    absl::flat_hash_map<std::string, std::unique_ptr<NamedIrNode>> nodes_;
+    absl::flat_hash_map<std::string, std::unique_ptr<Node>> nodes_;
   };
 
   // List of registered operators.
-  NamedIrNodeRegistry<Operator> operators_;
+  NamedIRNodeRegistry<Operator, kOperatorName> operators_;
   // List of registered storages.
-  NamedIrNodeRegistry<Storage> storages_;
+  NamedIRNodeRegistry<Storage, kStorageName> storages_;
   // TypeFactory for this context.
   types::TypeFactory type_factory_;
 };
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -16,9 +16,13 @@
 #ifndef SRC_IR_IR_CONTEXT_H_
 #define SRC_IR_IR_CONTEXT_H_
 
+#include <memory>
+#include <vector>
+
 #include "absl/container/flat_hash_map.h"
 #include "src/ir/operator.h"
 #include "src/ir/types/type_factory.h"
+#include "src/ir/module.h"
 
 namespace raksha::ir {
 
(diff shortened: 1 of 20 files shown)


src/ir/ir_context.h
```diff
@@ -33,29 +37,84 @@ class IRContext {
...
+  const Storage &GetStorage(absl::string_view operators_name) const {
+    return storages_.GetNode(operators_name, "storage");
+  }
+
+  // Returns true if the operator is registered with this context.
+  bool IsRegisteredStorage(absl::string_view operators_name) const {
+    return storages_.IsRegisteredNode(operators_name);
+  }
+
+  const Operation &RegisterOperation(Operation op) {
```
bgogul, line 70: Why do you need to register these individual operations? I am wondering a better way is to add a global module and keep the operations within there?

bgogul (commented):
Noticed a few things.

src/ir/ir_context.h
```diff
@@ -33,29 +37,84 @@ class IRContext {
...
+    operations_.push_back(std::make_unique<Operation>(std::move(op)));
+    return *operations_.back();
   }
 
  private:
+  template<class NamedIrNode>
+  class NamedIrNodeRegistry {
```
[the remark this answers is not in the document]
markww: This is to share code for retrieving `Operator`s and `Storage`s. They both follow the same pattern: they are IR nodes that contain a name that should be unique, and we maintain a map to ensure the unique mapping of name to node and to look up the nodes by name.

src/ir/ir_context.h
markww: You're right, that's probably cleaner.

src/ir/operation.cc
```diff
@@ -23,22 +23,61 @@
...
+      names, ", ",
+      [&](std::string* out, const absl::string_view name) {
+        auto find_res = map_to_print.find(name);
+        CHECK(find_res != map_to_print.end())
+          << "Name " << name << " not found in map_to_print";
+        absl::StrAppend(out, name, ": ", value_printing_fn(find_res->second));
+      });
+}
+
+std::string Operation::NoSsaToString() const {
```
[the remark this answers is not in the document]
bgogul: I started refactoring the `ToString` method as a visitor here: https://github.com/google-research/raksha/blob/arcs_to_ir%40bgogul/src/ir/ir_printer.h

src/ir/ir_context.h
markww: Added a `BlockBuilder` to the decode context, which eliminates the need for this.

src/ir/operation.cc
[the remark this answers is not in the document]
markww: I have re-written the test in a more visitor style, but I didn't use the visitor you wrote. I realized, while trying to adapt your visitor to my problem, that it isn't the right tool for the job because I wanted to visit values in dataflow order. In general, that starts to look a lot like a dataflow analysis. For this problem specifically, however, the values should just linearly nest. So I wrote a special purpose visitor and used some downcast methods.

markww:
I split up this PR, so it should be easier to review it (and its diffbases) now.

src/ir/proto/sql/decode_test.cc
```diff
@@ -0,0 +1,152 @@
...
+//
+// In addition, the process of traversing the value tree largely overlaps
+// with checking the properties that we want to check anyway.
+//
+// NOTE: It would be nicer to return the result rather than use an output
+// parameter, but that would preclude using the `ASSERT` macros (which expect
+// to be in a function returning void).
+void GatherSourceColumnInfoFromValues(
+    const Value val, ComparableSourceTableColumn &result) {
+  if (const value::StoredValue *storage_val = val.If<value::StoredValue>()) {
```
bgogul, line 83: A visitor is probably better than `if..else if.. else`? You could also consider `utils::overloaded`?

src/ir/proto/sql/decode.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <vector>
+
+#include "src/ir/value.h"
+#include "src/ir/proto/sql/decoder_context.h"
+#include "src/ir/proto/sql/sql_ir.pb.h"
+
+namespace raksha::ir::proto::sql {
+
+ir::Value DecodeSourceTableColumn(
```
bgogul, line 28: I think I am missing a lot of context to where this fits in the overall scheme of things. I think documentation of this method will help.

src/ir/proto/sql/decoder_context.h
```diff
@@ -0,0 +1,93 @@
...
+  static constexpr absl::string_view kTagClaimOperatorName = "tag_claim";
+  static constexpr absl::string_view kAddedTagAttributeName = "added_tag";
+  static constexpr absl::string_view kTagKindAttributeName = "tag_kind";
+  static constexpr absl::string_view kConfidentialityTagAttrValue =
+      "confidentiality";
+  static constexpr absl::string_view kIntegrityTagAttrValue = "integrity";
+  static constexpr absl::string_view kTagClaimInputValueName = "input";
+  static constexpr absl::string_view kTagClaimOutputName = "out";
+
+  explicit DecoderContext(IRContext &ir_context) : ir_context_(ir_context) { }
```
bgogul, line 45: Why do you need this class at all? Simply use `IRContext` directly?

src/ir/proto/sql/decoder_context.h
```diff
@@ -0,0 +1,93 @@
...
+                        : kConfidentialityTagAttrValue)}},
+                NamedValueMap{{std::string(kTagClaimInputValueName), value}}),
+        kTagClaimOutputName));
+  }
+
+  std::unique_ptr<Block> FinishGlobalSqlBlock() {
+    return global_sql_block_builder_.build();
+  }
+ private:
+  BlockBuilder global_sql_block_builder_;
```
bgogul, line 87: Do you actually need a block builder or just a list of operations, which can be added to a block later?

src/ir/proto/sql/decoder_context.h
```diff
@@ -0,0 +1,93 @@
...
+      return ir_context_.RegisterStorage(
+          std::make_unique<Storage>(
+              storage_name,
+              // Always use PrimitiveType, as we are ignoring types for the
+              // moment.
+              ir_context_.type_factory().MakePrimitiveType()));
+    }
+  }
+
+  ir::Value ClaimTagOnValue(
```
bgogul, line 60: why are you returning a value instead of an operation? How does this fit into the overall translation? Is there some sample uses of this function somewhere?

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,34 @@
...
+// This value corresponds to some source table for the query being analyzed.
+// It may have some confidentiality tags or integrity tags annotated onto it.
+// We entangle the tag claim aspect and the column identification because
+// that fits well with the policy matching procedure in the SQL verifier.
+message SourceTableColumn {
+  // The fully-qualified path to the column, including table name and (if it
+  // exists) schema name.
+  string column_path = 1;
+  repeated string confidentiality_tags = 2;
+  repeated string integrity_tags = 3;
```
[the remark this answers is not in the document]
markww: Yes, that's correct. I'll add a comment to clarify this.

src/ir/proto/sql/decoder_context.h
```diff
@@ -0,0 +1,93 @@
...
+      ir::Value value, absl::string_view tag, bool is_integrity) {
+    const Operator &tag_claim_op =
+        (ir_context_.IsRegisteredOperator(kTagClaimOperatorName))
+          ? ir_context_.GetOperator(kTagClaimOperatorName)
+          : ir_context_.RegisterOperator(
+              std::make_unique<Operator>(kTagClaimOperatorName));
```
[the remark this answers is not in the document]
markww: Good idea. I'll do that.

src/ir/proto/sql/decode_test.cc
```diff
@@ -0,0 +1,152 @@
...
+// be traversed right here.
+//
+// In addition, the process of traversing the value tree largely overlaps
+// with checking the properties that we want to check anyway.
+//
+// NOTE: It would be nicer to return the result rather than use an output
+// parameter, but that would preclude using the `ASSERT` macros (which expect
+// to be in a function returning void).
+void GatherSourceColumnInfoFromValues(
+    const Value val, ComparableSourceTableColumn &result) {
```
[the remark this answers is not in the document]
markww: No reason. I can take it out.

src/ir/proto/sql/decoder_context.h
markww: The latter. I can make this just a vector.

src/ir/proto/sql/decoder_context.h
markww: I was trying to put context-y type things that we would only use for the SQL MVP in here to not pollute the `IRContext` class. I can try to slim this down/eliminate it though.

src/ir/proto/sql/decode_test.cc
markww: OK, I'll check it out.

src/ir/ir_context.h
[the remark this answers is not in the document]
markww: This is now the domain of a different PR, so resolving.

src/ir/proto/sql/sql_ir.proto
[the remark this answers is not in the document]
markww: Added comment.

src/ir/proto/sql/decoder_context.h
markww: The reason why I am returning a `Value` instead of an `Operation` is that I am modeling columns as `Storage`s. While this drags more IR mechanism into this MVP implementation, I think it's worth it in this case, as SQL tables are very naturally `Storage`. In a more mature implementation, we'd probably make the whole table a `Storage` and select columns from that `Storage` using other `Operation`s, but making each individual column a `Storage` works for now. In addition, it prevents us from potentially having to put in some hacky Datalog to treat an `Operation` as a persistent data source when, in all other circumstances, they should be considered stateless.

src/ir/proto/sql/decode_test.cc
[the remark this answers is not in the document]

src/ir/proto/sql/decode_test.cc
markww: Implemented this using a visitor using `utils::overloaded`.

markww:
Converting this one to draft, as the underlying protos have changed so dramatically that it probably makes more sense to just open new PRs. Keeping the PR alive so I can keep stealing code from it.

markww:
This draft has served its purpose.

PR #380, opened 2022-02-16
Add and test new accessors to `Module` and `Operation`.

bgogul:


bgogul:
@markww I am not sure if this interferes with your work, but I needed this to write some tests for the visitor.

src/ir/module.h
```diff
@@ -34,11 +34,13 @@ class Block;
 class Operation {
  public:
   Operation(const Block* parent, const Operator& op,
-            NamedAttributeMap attributes, NamedValueMap inputs)
+            NamedAttributeMap attributes, NamedValueMap inputs,
+            std::unique_ptr<Module> module = nullptr)
       : parent_(parent),
-        op_(&op),
+        op_(std::addressof(op)),
```
markww, line 40: Why `std::addressof` instead of just `&`? I've never had to use it before, it looks like it gets the actual address when `&` is overloaded. Is that case coming up here?

src/ir/module.h
```diff
@@ -34,11 +34,13 @@ class Block;
 class Operation {
  public:
   Operation(const Block* parent, const Operator& op,
-            NamedAttributeMap attributes, NamedValueMap inputs)
+            NamedAttributeMap attributes, NamedValueMap inputs,
+            std::unique_ptr<Module> module = nullptr)
```
markww, line 38: My first thought was that, aesthetically, it's a little odd to have the `parent` `Block` be the first parameter and the parent `Module` be the last parameter. Then I realized that this `Module` is not the parent, but the implementation. To reduce confusion, maybe name the module something like `impl_module`?

markww, line 38: Would it make for cleaner calls to pass `module` in as type `Module` and have `Operation` put it in a `std::unique_ptr`... or even just embed it in a `std::optional<Module>`?

src/ir/module.h
```diff
@@ -48,6 +50,7 @@ class Operation {
   const Block* parent() const { return parent_; }
   const NamedValueMap& inputs() const { return inputs_; }
   const NamedAttributeMap& attributes() const { return attributes_; }
+  const Module* module() const { return module_.get(); }
```
markww, line 53: I'm a little worried about null safety here, as it's pretty subtle that the implementation module not being present is just wrapped in a `std::unique_ptr`. I'm not sure I see solutions that are extremely better though.

src/ir/module.h
```diff
@@ -104,20 +107,23 @@ class Block {
...
+
   Module(const Module&) = delete;
   Module& operator=(const Module&) = delete;
   Module(Module&&) = default;
   Module& operator=(Module&&) = default;
 
   // Adds a block to the module and returns a pointer to it.
-  const Block* AddBlock(std::unique_ptr<Block> block) {
-    const Block* result = block.get();
+  const Block& AddBlock(std::unique_ptr<Block> block) {
```
markww, line 119: It might be more concise at call sites to just take the block by value (like `AddBlock(Block block)`) then use it to move-initialize the unique pointer (like `blocks_.push_back(std::make_unique<Block>(block))`).

markww (approved):
I have style questions, but nothing major. LGTM.

src/ir/module.h
markww, line 38: One more thought about this `module`: if this corresponds to the implementation, shouldn't it be attached to the `Operator`?

src/ir/module.h
bgogul: It seemed more readable and cleaner. That is why I started shifting to this everywhere.

src/ir/module.h
bgogul: Isn't it more efficient just to pass in a `std::unique_ptr<>`?  Otherwise, we create something on the stack and then move it to the heap? Granted that most of the contents of `Block` is in the heap anyway and it may not be that much more expensive. However,  I am wondering if we really gain much in terms of conciseness by changing signature to  `AddBlock(Block block)`. WDYT?

src/ir/module.h
bgogul: Good idea. Will fix it.  

I think there is a related problem with the use of `module` in `Block`. It should really be called `parent` there. I will fix that issue in a separate PR.

src/ir/module.h
bgogul: I agree with the overall sentiment around null safety. However, I don't think things are that much better if we return `const &`. The main advantage of returning `const &` is that we indicate to the caller that the value cannot be `null`. 

Other than I am not sure how better is it in terms of memory safety. The underlying object should still exist until the lifetime of the `const &`. It is true that there is automatic extension of the lifetime of the underlying object, but it is quite limited: https://abseil.io/tips/107

In this specific case, the API suggests that the `module()` could be `nullptr` and the callers have to take care of that situation appropriately. So, it seems OK to return `const Module*`?

src/ir/module.h
bgogul: This allows more flexibility. For example, we don't have define and register an operator for every function or particle that we want to define. We will have an operator called `MakeFunction` (or `MakeParticle`) to define a function. The `Operation` (not `Operator`) instance of a `MakeFunction` will define the body of the function using the module. 

We will invoke the function with `CallFunction` (or `CallParticle`) that will take the `MakeFunction` operation instance as an argument (or perhaps attribute).

bgogul: This is similar to your question about `AddBlock(Block block)`. Let us discuss there?

src/ir/module.h
markww: Not necessarily. If we create the `Block` as a temporary in the argument list to `AddBlock` like this:

```C++
const Block *canonical_block = module.AddBlock(Block(/*...*/));
```

Then the created `Block` will likely never touch the stack. Copy elision will cause it to be created directly into the parameter. If that parameter is `std::move`d, into the `std::unique_ptr`, that means that it will likely be created directly into the allocated memory.

Performance isn't my main concern here, I mostly just think it's more concise to, for some type `T`, pass in an argument of `T(/*params*/)` rather than `std::make_unique<T>(/*params*/)`. But if you like it the other way, I think it's fine as is.

src/ir/module.h
markww: That sounds like a level of dynamism that could reduce the effectiveness of the static analysis. It may require a conservative overapproximation on what call sites can call what function bodies. If we wanted something like that, we could create a new kind of `Value` (`LambdaValue`?) to wrap dynamically-passed functions and pass them as arguments to other functions.

In that case, every `Operation` would have one non-optional `Operator` and the `Operator` would have an optional implementing `Module`, depending on whether the implementation is known. If we need dynamism like what you are describing, that could be indicated by an `Operator` that returns a `LambdaValue`, which may be passed to a function that calls that `LambdaValue`.

I believe having the implementing `Module` owned by the `Operator` could cause significant savings in memory, as static calls are usually the common case. There are generally many more call sites (`Operation`s) than function definitions (`Operator`s). Putting the `Module` on the `Operation` and pointing to it with a `unique_ptr` would require multiple "calls" to the same `Operator` to have duplicate copies of the implementation at different memory addresses, causing more space to be used and the cache to be hit less.

src/ir/module.h
bgogul: I see. We cannot initialize a `Block` with a single constructor call like `Block(...)`. e.g. An `Operation` in a block refers to the parent block. That is why we use `BlockBuilder` which returns `std::unique_ptr<Block>`.
  
Let us leave this for now and can revisit if it becomes a major pain point.

src/ir/module.h
bgogul: I don't think this will necessarily reduce the effectiveness of static analysis. If the argument to a `CallFunction` is the result of `MakeFunction` operation, it is similar to a direct call? 

Also, I am not sure there will be a difference in memory usage.  There will be one `MakeFunction` operation which will hold on to `Module`. The `CallFunction` operation will refer to that instance of `MakeFunction` operation?  

In any case, there is so much to unpack here and is probably a topic for a broader discussion. Let us discuss this separately?

bgogul: Done.

bgogul: Resolving this for now. We should have a longer conversation about this direction.

PR #384, opened 2022-02-18
Make attributes and values of operations be printed out in name order.

markww:
Previously, the attributes and names of operations would be printed out
in arbitrary order. As of this commit, they are always printed out in
the alphabetical order of their associated name.

### DIFF
--- a/src/ir/operation.cc
+++ b/src/ir/operation.cc
@@ -23,22 +23,43 @@
 
 namespace raksha::ir {
 
+// Note: I tried to make this only defined in terms of the type parameter T, but
+// the compiler was unhappy with me writing std::function<std::string(const T&)>
+// and I could not figure out how to placate it.
+template<class T, class U>
+static std::string PrintNamedMapInNameOrder(
+    const absl::flat_hash_map<std::string, T> &map_to_print,
+    U value_printing_fn) {
+  std::vector<absl::string_view> names;
+  names.reserve(map_to_print.size());
+  for (auto &key_value_pair : map_to_print) {
+    names.push_back(key_value_pair.first);
+  }
+  std::sort(names.begin(), names.end());
+  return absl::StrJoin(
+      names, ", ",
+      [&](std::string* out, const absl::string_view name) {
+        auto find_res = map_to_print.find(name);
+        CHECK(find_res != map_to_print.end())
+          << "Name " << name << " not found in map_to_print";
+        absl::StrAppend(out, name, ": ", value_printing_fn(find_res->second));
+      });
+}
+
 std::string Operation::ToString(SsaNames& ssa_names) const {
   constexpr absl::string_view kOperationFormat = "%%%d = %s [%s](%s)";
   SsaNames::ID this_ssa_name = ssa_names.GetOrCreateID(*this);
-  std::string attributes_string = absl::StrJoin(
-      attributes_, ", ",
-      [](std::string* out, const NamedAttributeMap::value_type& v) {
-        const auto& [name, attribute] = v;
-        absl::StrAppend(out, name, ": ", attribute->ToString());
-      });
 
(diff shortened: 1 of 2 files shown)


src/ir/operation.cc
```diff
@@ -23,22 +23,43 @@
 
 namespace raksha::ir {
 
+// Note: I tried to make this only defined in terms of the type parameter T, but
+// the compiler was unhappy with me writing std::function<std::string(const T&)>
+// and I could not figure out how to placate it.
+template<class T, class U>
+static std::string PrintNamedMapInNameOrder(
```
bgogul, line 30: This should be a visitor, but I can take care of it once these changes are in.

src/ir/operation.cc
```diff
@@ -23,22 +23,43 @@
 
 namespace raksha::ir {
 
+// Note: I tried to make this only defined in terms of the type parameter T, but
+// the compiler was unhappy with me writing std::function<std::string(const T&)>
+// and I could not figure out how to placate it.
+template<class T, class U>
```
bgogul, line 29: `F` is typically used for functors and lambdas?
```suggestion
template<class T, class F>
```

src/ir/operation.cc
```diff
@@ -23,22 +23,43 @@
...
+static std::string PrintNamedMapInNameOrder(
+    const absl::flat_hash_map<std::string, T> &map_to_print,
+    U value_printing_fn) {
```
bgogul, line 32: `F value_pretty_printer`?

src/ir/operation.cc
```diff
@@ -23,22 +23,43 @@
...
+  std::vector<absl::string_view> names;
+  names.reserve(map_to_print.size());
+  for (auto &key_value_pair : map_to_print) {
+    names.push_back(key_value_pair.first);
+  }
+  std::sort(names.begin(), names.end());
```
bgogul, line 38: Insert into an  absl::btree_set to get sorted right away?

src/ir/operation.cc
```diff
@@ -23,22 +23,43 @@
...
+  return absl::StrJoin(
+      names, ", ",
+      [&](std::string* out, const absl::string_view name) {
+        auto find_res = map_to_print.find(name);
+        CHECK(find_res != map_to_print.end())
+          << "Name " << name << " not found in map_to_print";
```
bgogul, line 44: You could simply use `map_to_print.at(name)` as it is guaranteed to exist?

bgogul (commented):
Left some mostly style suggestions.

src/ir/operation.cc
markww: Usually, I like to do offline sorts when possible, as it tends to perform better. But in this circumstance, where it is plausible that the value is printed out multiple times and it is much more concise, I think perhaps your suggestion of a `btree_set` makes more sense.

src/ir/operation.cc
markww: I did not know that `at` existed on `unordered_map`! Thanks for pointing that out.

src/ir/operation.cc
markww: OK, I can do that.

src/ir/operation.cc
markww: OK, can do.

src/ir/operation.cc
markww: Done.

src/ir/operation.cc
markww: Done.

src/ir/operation.cc
markww: Done.

src/ir/operation.cc
bgogul: Sounds good.

PR #385, opened 2022-02-18
Make IRContext manage Storages in addition to Operators.

markww:
Previously, IRContext managed the mapping of names to Operators. This
change makes it also manage the mapping of names to Storages and keep
track of canonical copies of Operations. Because the logic for Operators
and Storages is so similar, an inner templatized class is used to
serve both kinds of IrNodes.

### DIFF
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -37,13 +37,13 @@ class IRContext {
 
   // Register an operator and return the stored operator instances.
   const Operator &RegisterOperator(std::unique_ptr<Operator> op) {
-    return operators_.RegisterNode(std::move(op), "operator");
+    return operators_.RegisterNode(std::move(op));
   }
 
   // Returns the Operator with a particular name. If there is no
   // operator with that name, fail.
   const Operator &GetOperator(absl::string_view operators_name) const {
-    return operators_.GetNode(operators_name, "operator");
+    return operators_.GetNode(operators_name);
   }
 
   // Returns true if the operator is registered with this context.
@@ -53,13 +53,13 @@ class IRContext {
 
     // Register an operator and return the stored operator instances.
   const Storage &RegisterStorage(std::unique_ptr<Storage> op) {
-    return storages_.RegisterNode(std::move(op), "storage");
+    return storages_.RegisterNode(std::move(op));
   }
 
   // Returns the Storage with a particular name. If there is no
   // operator with that name, fail.
   const Storage &GetStorage(absl::string_view operators_name) const {
-    return storages_.GetNode(operators_name, "storage");
+    return storages_.GetNode(operators_name);
   }
 
   // Returns true if the operator is registered with this context.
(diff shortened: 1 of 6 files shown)


src/ir/ir_context.h
```diff
@@ -33,29 +37,77 @@ class IRContext {
...
+  }
+
+  // Returns true if the operator is registered with this context.
+  bool IsRegisteredStorage(absl::string_view operators_name) const {
+    return storages_.IsRegisteredNode(operators_name);
   }
 
  private:
+  template<class NamedIrNode>
+  class NamedIrNodeRegistry {
```
bgogul, line 72: I know `Ir` is according to style guide. Nevertheless, let us stick with `IR` as the capitalization if `IR` is used in class/function names? 

I think we already have `IRContext`. If you strongly prefer `Ir` then we should also change `IrContext`. Let us sync up more on this.

src/ir/ir_context.h
```diff
@@ -33,29 +37,77 @@ class IRContext {
...
+    return storages_.GetNode(operators_name, "storage");
+  }
+
+  // Returns true if the operator is registered with this context.
+  bool IsRegisteredStorage(absl::string_view operators_name) const {
+    return storages_.IsRegisteredNode(operators_name);
   }
 
  private:
+  template<class NamedIrNode>
```
bgogul, line 71: You can add `node_kind_name` as a template argument? Also `T` or `Node` might be OK instead of `NamedIrNode`.

```suggestion
  template<class T, const char* node_kind>
```

src/ir/ir_context.h
```diff
@@ -33,29 +37,77 @@ class IRContext {
...
+  class NamedIrNodeRegistry {
+   public:
+      // Register a node and return the stored instance.
+    const NamedIrNode &RegisterNode(std::unique_ptr<NamedIrNode> node,
+                                    absl::string_view node_kind_name) {
+        // Note: using node->name() here is ok because C++ list initialization
```
bgogul, line 77: Is this in the standard?

src/ir/ir_context.h
markww: I don't have strong preferences, I'll use `IR`.

src/ir/ir_context.h
markww: Yes! I looked it up because I was worried that they could be reordered freely, like the evaluation of function call arguments. Turns out, there's a sequenced-before guarantee here: https://en.cppreference.com/w/cpp/language/list_initialization#Notes

src/ir/ir_context.h
markww: This isn't allowed until C++20. String literals are not allowed as template parameter arguments because the C++ standard has too loose of requirements upon the uniquification of string literals with identical contents (if I recall correctly, it mandates it at the TU granularity, but not for the whole binary).

src/ir/ir_context.h
markww: Done.

src/ir/ir_context.h
markww: Changed type param name.

src/ir/module.h
```diff
@@ -43,9 +43,9 @@ class Operation {
         inputs_(std::move(inputs)),
         impl_module_(std::move(impl_module)) {}
 
-  // Disable copy (and move) semantics.
+  // Disable copy semantics.
   Operation(const Operation&) = delete;
-  Operation& operator=(const Operation&) = delete;
+  Operation(Operation &&) = default;
```
bgogul, line 48: Why do you allow `operator=`?

src/ir/module.h
```diff
@@ -83,7 +83,7 @@ class Block {
 
   Block() : module_(nullptr) {}
 
-  // Disable copy (and move) semantics.
+  // Disable copy semantics.
```
bgogul, line 86: When you remove copy semantics, it also disables move.

src/ir/ir_context.h
```diff
@@ -33,29 +37,77 @@ class IRContext {
...
+    return storages_.IsRegisteredNode(operators_name);
   }
 
  private:
+  template<class Node>
+  class NamedIRNodeRegistry {
+   public:
+      // Register a node and return the stored instance.
+    const Node &RegisterNode(std::unique_ptr<Node> node,
+                                    absl::string_view node_kind_name) {
```
bgogul, line 76: Make `node_kind_name` a member field of the class and initialize it on construction?

src/ir/ir_context.h
bgogul: Sounds good.

bgogul (commented):
Seems to have some unrelated changes. Other than that looks good.

src/ir/ir_context.h
bgogul: Ah, I thought if you define a constexpr, you could use it:

```C++
constexpr const char* kOperatorNodeKind[] = "operator";

NamedIRNodeRegistry<Operator, kOperatorNodeKind> ...;
``` 

It is possible that I am misremembering this.

markww: I'll give it a try, but I think C++17 doesn't like it.

src/ir/module.h
markww: This was for another change, don't need this anymore. I've removed it.

src/ir/module.h
markww: This was a misfire in updating the comment, meant to update the other one. I've backed this change out.

src/ir/ir_context.h
markww: Wow, you were right, it does allow `constexpr` `char[]` globals in the template, but not the unwrapped string literal. Good catch!

markww: Done.

src/ir/ir_context.h
markww: I did this the other way, using a template argument.

markww:
I think I addressed the comments, PTAL. Your `constexpr` trick for the strings worked!

bgogul (approved):

PR #388, opened 2022-02-23
Add support for modules in  SsaNames class.

bgogul:


src/ir/ssa_names_test.cc
```diff
@@ -22,34 +22,35 @@
...
+SsaNamesTest<Operation>::SsaNamesTest()
+    : test_op_("test"),
+      first_entity_(nullptr, test_op_, {}, {}),
+      second_entity_(nullptr, test_op_, {}, {}) {}
 
-  EXPECT_NE(first_operation_id, second_operation_id);
-  EXPECT_EQ(names.GetOrCreateID(first_operation), first_operation_id);
-}
+using MyTypes = ::testing::Types<Operation, Block, Module>;
+TYPED_TEST_SUITE(SsaNamesTest, MyTypes);
```
markww, line 45: I think a more specific alias than `MyTypes` is called for here, that's a pretty generic name. Maybe `SsaTestTypes`? Alternatively, you could just inline the definition of the `MyTypes` alias into the one place where it is used.

src/ir/ssa_names_test.cc
```diff
@@ -22,34 +22,35 @@
...
-  Block second_block;
-  SsaNames::ID first_block_id = names.GetOrCreateID(first_block);
-  SsaNames::ID second_block_id = names.GetOrCreateID(second_block);
+  SsaNames::ID first_id = names.GetOrCreateID(this->first_entity_);
+  SsaNames::ID second_id = names.GetOrCreateID(this->second_entity_);
 
-  EXPECT_NE(first_block_id, second_block_id);
-  EXPECT_EQ(names.GetOrCreateID(first_block), first_block_id);
+  EXPECT_NE(first_id, second_id);
+  EXPECT_EQ(names.GetOrCreateID(this->first_entity_), first_id);
```
markww, line 53: I know it doesn't add that much value to the test, but since you have the `second_id` lying around as well and there are only two IDs, it might make sense to just throw in a

```C++
EXPECT_EQ(names.GetOrCreateID(this->second_entity_), second_id);
```

just for completeness's sake.

src/ir/ssa_names.h
```diff
@@ -54,6 +59,7 @@ class SsaNames {
     absl::flat_hash_map<const T *, ID> item_ids_;
   };
 
+  IDManager<Module> module_ids_;
```
markww, line 62: Nit: It looks like you generally use the order {`Block`, `Operation`, `Module`} in this file, but here you are using {`Module`, `Operation`, `Block`}. Move this to the end of the field decls to make this a bit more uniform?

markww (approved):
Looks good, just a few nits.

src/ir/ssa_names_test.cc
bgogul: done

src/ir/ssa_names_test.cc
bgogul: Done.

src/ir/ssa_names.h
bgogul: Chose `{Operation, Block, Module}` to match `module.h` and made this consistent.

PR #389, opened 2022-02-23
Add protos used for communicating with SQL verifier for MVP impl.

markww:
Add the protos that we will use to communicate with the SQL verifier
for the MVP. These protos are built to make it very easy to export the
minimum amount of state possible from the SQL verifier to provide some
dataflow analysis value. A mature implementation should probably throw
these protos out entirely and rewrite them to capture more information
and be better factored.

### DIFF
--- a/src/ir/proto/sql/sql_ir.proto
+++ b/src/ir/proto/sql/sql_ir.proto
@@ -22,79 +22,87 @@ syntax = "proto3";
 package raksha.ir.proto.sql;
 
 // This value corresponds to some source table for the query being analyzed.
-// It may have some confidentiality tags or integrity tags annotated onto it.
-// We entangle the tag claim aspect and the column identification because
-// that fits well with the policy matching procedure in the SQL verifier.
 message SourceTableColumn {
   // The fully-qualified path to the column, including table name and (if it
   // exists) schema name.
   string column_path = 1;
-  repeated string confidentiality_tags = 2;
-  repeated string integrity_tags = 3;
+}
+
+// A literal. The SQL verifier currently just passes on the string form of the
+// literal as seen by the parser. For the MVP, we just continue passing this
+// along.
+message Literal {
+  string literal_str = 1;
 }
 
 // This `MergeOperation` is an all-purpose combination of inputs into a single
-// output. For the MVP, this is the only operation we will use to model our SQL
-// dataflow other than policy rule pieces.
+// output.
 message MergeOperation {
-  // We may optionally associate merges with names to help with building
-  // AccessPaths. This name may include column and/or table aliases provided
-  // by the user.
-  string name = 1;
   // Direct dataflow dependencies. Relevant for confidentiality and integrity
   // tags.
-  repeated Value inputs = 2;
+  repeated Expression inputs = 2;
   // Control dependencies. Confidentiality tags can flow from these inputs to
   // the outputs. They are irrelevant to integrity tags.
-  repeated Value control_inputs = 3;
-  // Whether or not this merge actually transforms the data in some way or
-  // whether it just merges the data, keeping all possible integrity
-  // properties.
-  bool xforms_data = 4;
-  // A unique ID assigned to each MergeOperation. Allows PolicyTransform
-  // nodes to point at MergeOperations, as the need to be able to check the
-  // tag state of particular MergeOperations to decide whether to perform the
-  // transform.
-  uint64 id = 5;
+  repeated Expression control_inputs = 3;
 }
 
-// The action that is taken by a policy if all preconditions match.
-enum PolicyTransformAction {
+// The kind of a tag.
+enum TagKind {
+  UNKNOWN = 0;
+  INTEGRITY = 1;
+  CONFIDENTIALITY = 2;
+}
+
+// The action that is taken upon a tag by a policy rule.
+enum TagAlteration {
   NONE = 0;
-  ADD_INTEGRITY = 1;
-  REMOVE_CONFIDENTIALITY = 2;
-}
-
-// A value is one of a `SourceTableColumn`, which represents a column in one
-// of the input tables to the SQL query, or a `MergeOperation`, which is any
-// other operation that is performed upon data.
-message Value {
-  oneof value_variant {
-    SourceTableColumn source_table_column = 1;
-    MergeOperation merge_operation = 2;
-  }
+  ADD = 1;
+  REMOVE = 2;
 }
 
 // A transform upon the state of a program as dictated by a policy rule.
 //
-// A `PolicyTransform` refers to both the node that it will perform its
-// transform upon and the nodes that are considered inputs to the policy
-// transforms via ID rather than nesting the nodes directly underneath them.
-// This allows the SQL AST to be built where each AST node owns, and has
-// nested within it, its child node and the PolicyTransforms can be placed to
-// the side without altering that structure.
+// A `PolicyTransform` is structured like a cast upon some `transformed_node`
+// `Expression`. The preconditions of this transform, however, are presented as
+// IDs rather than nesting the nodes directly underneath them. This is
+// necessary because these inputs may not be direct children of the
+// transformed node. Because protos only allow each structure to have a
+// single nesting parent, we need a different way of referring to them for
+// these additional references.
 message PolicyTransform {
-  // The MergeOperation that is having its tag state changed by the
-  // PolicyTransform.
-  uint64 transformed_node_id = 1;
+  // The `Expression` that is having its tag state changed by the
+  // `PolicyTransform`.
+  Expression transformed_node = 1;
   // A map from the unique ID of a particular input `MergeOperation` to the
   // integrity tag that is required to be present upon that operation for
   // this transform to execute.
   map<uint64, string> required_integrity_tags = 2;
   // The action to be performed on the tag state of the node indicated by the
   // `transformed_node_id` if all of the required integrity tags are present
-  // on the nodes indicated in the `required_integrity_tags` map.
-  PolicyTransformAction action = 3;
+  // on the nodes indicated in the `required_integrity_tags` map. Can be
+  // added or removed.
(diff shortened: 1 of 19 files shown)


src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,100 @@
...
+  REMOVE_CONFIDENTIALITY = 2;
+}
+
+// A value is one of a `SourceTableColumn`, which represents a column in one
+// of the input tables to the SQL query, or a `MergeOperation`, which is any
+// other operation that is performed upon data.
+message Value {
+  oneof value_variant {
+    SourceTableColumn source_table_column = 1;
+    MergeOperation merge_operation = 2;
```
bgogul, line 74: Don't you need the index/name of the result? Or, is it the case that there is always one output of a merge operation?

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,100 @@
...
+  // We may optionally associate merges with names to help with building
+  // AccessPaths. This name may include column and/or table aliases provided
+  // by the user.
+  string name = 1;
+  // Direct dataflow dependencies. Relevant for confidentiality and integrity
+  // tags.
+  repeated Value inputs = 2;
+  // Control dependencies. Confidentiality tags can flow from these inputs to
+  // the outputs. They are irrelevant to integrity tags.
+  repeated Value control_inputs = 3;
```
bgogul, line 49: nit: control_dependencies

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,100 @@
...
+  // Whether or not this merge actually transforms the data in some way or
+  // whether it just merges the data, keeping all possible integrity
+  // properties.
+  bool xforms_data = 4;
```
bgogul, line 53: What is the advantage of keeping the `PolicyTransform` separately? Why can't we  remove the two fields (`xforms_data` and `id`) and replace them with `PolicyTransform`? 

If you do that, you don't need any ids as well?

src/ir/proto/sql/sql_ir.proto
bgogul: On second though, `control_inputs` might be fine as well.

src/ir/proto/sql/sql_ir.proto
markww: There will always be one output of a merge operation. SQLite doesn't have any temporary names, other than those introduced by a `WITH` clause, which the verifier does not currently handle.

src/ir/proto/sql/sql_ir.proto
markww: Two parts to this question:

1. `xforms_data` does not mean "performs a policy transform on the data", it means "combines the data in some way that alters values or layout". You could imagine a `UNION ALL`, the list of expressions used with an `IN` operator, or a `WHERE` clause as having `xforms_data` set to `false`, while most arithmetic operations would have it set to `true`. I can change the name to remove this source of confusion.

2. We need the IDs because a `PolicyTransform` does not need to occur over only a single `MergeOperation`. Consider a pattern that attempts to match an expression comparing a timestamp against a window, like so:

```
?a - ?b > ?window
```

Let's say that the verifier is run over some SQL containing the following code:

```
col1 - col2 > 1000
```

The above expression will contain two `MergeOperation`s: one for `-` and one for `>`. The policy rule will match this expression and produce from it a single `PolicyTransform`: one which takes as inputs `col1`, `col2`, and `col3` and presumably performs some matching of `IntegrityTag`s upon them.

In this case, there is no one `MergeOperation` that has, in its scope, the result and all of the inputs to the transform. It is for this reason that we need IDs to point at different values.

(Typing out this explanation has made me realize I need a third kind of value, a literal value, and that the ID and name need to be associated with `Value`s, not just `MergeOperation`s).

src/ir/proto/sql/sql_ir.proto
markww: I'll change this.

src/ir/proto/sql/sql_ir.proto
bgogul: IIUC, typically there would be two merge operations: 
 - `%0 = col1 - col2`
 - `%1 =  %0 > 0`
 
However, if there is a pattern match, we will also see this as a whole during the visit in addition to the two above:
 -  `%x = col1 - col2 > 0`
 
Isn't that the case? If so, I am saying that `%0` and `%1` will effectively be dropped becausese only `%x` will be passed on to the parent of `col1 - col2 > 0`.

bgogul: About your other comment about `xforms_data`, why can't we capture `UNION_ALL` as a `Transform` also?

markww: That would work if there were only one policy rule. But what if there were a separate policy rule that matched:

```
?a - ?b
```

for some integrity tags? In that case, two overlapping subtrees of the original AST could be matched by separate policy rules, and destructively modifying the AST upon which the policy rules operate could cause one of the matches to be dropped.

markww:
I've refactored this as we discussed in meetings to center the new `Expression` class more. PTAL, let me know what you think.

src/ir/proto/sql/sql_ir.proto
markww: I ended up not changing this, but am open to it if you still don't like it. Sounds like you are ok with it now though.

src/ir/proto/sql/sql_ir.proto
markww: As we discussed on the phone, I've gotten rid of the confusing `xforms_data` field and have made the `PolicyTransform` message an `Expression` that hangs off of the normal `Expression` tree. Keeping the IDs as inputs to `PolicyTransform` to prevent mutating the tree and/or duplicating subtrees.

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,108 @@
...
+// A literal. The SQL verifier currently just passes on the string form of the
+// literal as seen by the parser. For the MVP, we just continue passing this
+// along.
+message Literal {
+  string literal_str = 1;
+}
+
+// This `MergeOperation` is an all-purpose combination of inputs into a single
+// output.
+message MergeOperation {
```
bgogul, line 40: Is it right that this will always joins all the confidentiality labels and drops all the integrity labels? Document whatever the behavior is?

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,108 @@
...
+// A transform upon the state of a program as dictated by a policy rule.
+//
+// A `PolicyTransform` is structured like a cast upon some `transformed_node`
+// `Expression`. The preconditions of this transform, however, are presented as
+// IDs rather than nesting the nodes directly underneath them. This is
+// necessary because these inputs may not be direct children of the
+// transformed node. Because protos only allow each structure to have a
+// single nesting parent, we need a different way of referring to them for
+// these additional references.
+message PolicyTransform {
```
bgogul, line 72: Techincally, this does not have anything to do with `Policy` and thefore, we could call this something else. How about `TagAlterationExpression` or `TagTransformExpression` or something like that.

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,108 @@
...
+  // The `Expression` that is having its tag state changed by the
+  // `PolicyTransform`.
+  Expression transformed_node = 1;
```
bgogul, line 75: Is there always one expression? You might want to change this to

```
repeated Expression inputs = 1;
``` 
and use the  index of inputs as the key in `required_integrity_tags`?

Essentially, collect all the required context in this message as `inputs`.

src/ir/proto/sql/sql_ir.proto
```diff
@@ -0,0 +1,108 @@
...
+  // A map from the unique ID of a particular input `MergeOperation` to the
```
bgogul, line 76: Does this have to be just `MergeOperation`? It can be any `Expression`, right?

bgogul (approved):
I left one major comment about `transformed_node` and `required_integrity_tags` of policy transform. Other than that there are some minor naming suggestions.

src/ir/proto/sql/sql_ir.proto
markww: Will do.

src/ir/proto/sql/sql_ir.proto
markww: OK, I can do that.

src/ir/proto/sql/sql_ir.proto
markww: There is always one root expression for these transformations. The message currently known as `PolicyTransform` acts as a cast upon the `transformed_node`. It may consider whether `IntegrityTag`s are present upon other `Expression`s, but because they are structurally nested within other `Expression`s (either the `transformed_node` or one of its transitive children) we cannot embed them here as well.

src/ir/proto/sql/sql_ir.proto
markww: It can be any `Expression`. I thought I had gotten all of these old overly-specific uses of `MergeOperation`, but I missed this one. Thanks!

src/ir/proto/sql/sql_ir.proto
markww: Renamed `PolicyTranform` to `TagTransform`, got rid of all mentions of policy in this file.

src/ir/proto/sql/sql_ir.proto
markww: Fixed.

src/ir/proto/sql/sql_ir.proto
markww: Done.

PR #390, opened 2022-02-24
Add `DecoderContext`, a SQL-specific context. Add `GetOrCreateStorage` to this class.

markww:
Most of the time, we want to create `Storage`s in advance. For the SQL MVP, however, it is very useful to lazily create `Storage`s because of how SQL buries its context at the bottom of the queries. This PR creates a new `DecoderContext` to keep state and methods that are SQL specific, and adds a `GetOrCreateStorage` there.

### DIFF
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -51,7 +51,12 @@ class IRContext {
     return operators_.IsRegisteredNode(operators_name);
   }
 
-    // Register an operator and return the stored operator instances.
+  const Operator &GetOrCreateOperator(absl::string_view operator_name,
+                                      std::unique_ptr<Operator> new_operator) {
+    return operators_.GetOrCreateNode(operator_name, std::move(new_operator));
+  }
+
+  // Register an operator and return the stored operator instances.
   const Storage &RegisterStorage(std::unique_ptr<Storage> op) {
     return storages_.RegisterNode(std::move(op));
   }
@@ -65,6 +70,11 @@ class IRContext {
   // Returns true if the operator is registered with this context.
   bool IsRegisteredStorage(absl::string_view operators_name) const {
     return storages_.IsRegisteredNode(operators_name);
+  }
+
+  const Storage &GetOrCreateStorage(absl::string_view storage_name,
+                                    std::unique_ptr<Storage> new_storage) {
+    return storages_.GetOrCreateNode(storage_name, std::move(new_storage));
   }
 
  private:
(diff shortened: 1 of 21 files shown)


src/ir/ir_context.h
```diff
@@ -51,7 +51,12 @@ class IRContext {
     return operators_.IsRegisteredNode(operators_name);
   }
 
-    // Register an operator and return the stored operator instances.
+  const Operator &GetOrCreateOperator(absl::string_view operator_name,
```
bgogul, line 54: hmm.. `Operator` and `Storage` should be created in one place. e.g., declaration of storage or declaration of an operator. I am not sure where you are planning to use this, but this API can lead to silent errors that would become hard to track.  

It looks like you are passing in a constructed `Operator` or `Storage` anyway. Why not use `RegisterOperator` and `GetOperator`? What advantages do you get with a `GetOrCreateOperator` and other variants?

markww: You might be right about `Operator`; we know all of the `Operator`s that we want to use for the SQL MVP in advance, and so we can just create them.

For `Storage`s, which are used for columns currently (and will most likely be used for tables in the future) something like `GetOrCreate` looks much more appealing because we do not know in advance the columns that a user will request. Therefore, it is appealing to be able to `GetOrCreate` a `Storage` in a single call when a column reference is being deserialized.

Perhaps the way to solve this is to keep `GetOrCreateNode`, but eliminate `GetOrCreateOperator`, thus enabling it for `Storage`s and disabling it for `Operator`s.

markww: Done.

markww:
Refactored the `GetOrCreateStorage` so it is on `DecoderContext` rather than `IRContext`. PTAL.

bgogul (approved):
Just wonder if you need `DecoderContext` at all or a helper function might be sufficient?

src/ir/proto/sql/decoder_context.h
```diff
@@ -0,0 +1,57 @@
...
+#define SRC_IR_PROTO_SQL_DECODER_CONTEXT_H_
+
+#include <memory>
+
+#include "absl/container/flat_hash_map.h"
+#include "src/ir/ir_context.h"
+#include "src/ir/value.h"
+#include "src/ir/proto/sql/sql_ir.pb.h"
+
+namespace raksha::ir::proto::sql {
```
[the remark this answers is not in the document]
markww: Interesting, will consider this for follow-on PR.

PR #391, opened 2022-02-24
Decode SourceTableColumn SQL proto.

markww:
This PR decodes the SourceTableColumn SQL proto. The decoding of this
proto is fairly straightforward, as it just involves getting the storage
indicated by the proto (creating it if it does not yet exist) and
wrapping the whole thing in a `StorageValue` and a `Value`.

### DIFF
--- a/src/ir/proto/sql/decode.cc
+++ b/src/ir/proto/sql/decode.cc
@@ -20,7 +20,7 @@ namespace raksha::ir::proto::sql {
 
 ir::Value DecodeSourceTableColumn(
     const SourceTableColumn &source_table_column,
-    IRContext &ir_context) {
+    DecoderContext &decoder_context) {
   const std::string &column_path = source_table_column.column_path();
   CHECK(!column_path.empty()) << "Required field column_path was empty.";
   // Note: In the future, we will probably want to make a `Storage` per table
@@ -32,11 +32,7 @@ ir::Value DecodeSourceTableColumn(
   // probably want to change that when we start handling types in a
   // non-trivial fashion.
   return ir::Value{value::StoredValue(
-      ir_context.GetOrCreateStorage(
-          column_path,
-          std::make_unique<Storage>(
-              column_path,
-              ir_context.type_factory().MakePrimitiveType())))};
+      decoder_context.GetOrCreateStorage(column_path))};
 }
 
 }  // namespace raksha::ir::proto::sql
--- a/src/ir/proto/sql/decode.cc
+++ b/src/ir/proto/sql/decode.cc
@@ -18,7 +18,7 @@
 
 namespace raksha::ir::proto::sql {
 
-ir::Value DecodeSourceTableColumn(
+Value DecodeSourceTableColumn(
     const SourceTableColumn &source_table_column,
     DecoderContext &decoder_context) {
   const std::string &column_path = source_table_column.column_path();
@@ -31,8 +31,8 @@ ir::Value DecodeSourceTableColumn(
   // Also, for now, we consider all storages to have primitive type. We will
   // probably want to change that when we start handling types in a
   // non-trivial fashion.
-  return ir::Value{value::StoredValue(
-      decoder_context.GetOrCreateStorage(column_path))};
+  return Value(value::StoredValue(
+      decoder_context.GetOrCreateStorage(column_path)));
 }
 
 }  // namespace raksha::ir::proto::sql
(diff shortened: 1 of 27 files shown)


src/ir/proto/sql/decode.cc
```diff
@@ -0,0 +1,42 @@
...
+  // Note: In the future, we will probably want to make a `Storage` per table
+  // and grab columns from that table using operators. For now, however,
+  // making each column an independent storage works for the MVP and is much
+  // simpler.
+  //
+  // Also, for now, we consider all storages to have primitive type. We will
+  // probably want to change that when we start handling types in a
+  // non-trivial fashion.
+  return ir::Value{value::StoredValue(
+      ir_context.GetOrCreateStorage(
```
bgogul, line 35: I commented on the parent PR. Ideally, we should setup the `IRContext` with all the `Operators` and `Storage` and use it here. Adding things on the fly can be error prone. e.g., if we add more attributes to `Storage`, how do we ensure that all uses are consistent. 

With `GetOrCreateXYZ` variants, it looks like the first occurrence wherever it is wins?

src/ir/value.h
```diff
@@ -114,6 +114,15 @@ class Value {
...
 
+  // A downcast operation. Just delegates directly to std::get on variants.
+  template<class T>
+  const T &As() const { return std::get<T>(value_); }
+
+  // A dynamic downcast operation. Just delegates directly to std::get_if on
+  // variants.
+  template<class T>
+  const T *If() const { return std::get_if<T>(&value_); }
+
```
bgogul, line 125: Are these needed anymore?

src/ir/proto/sql/decode.cc
markww: Yes. One way to guard against the errors that you are describing might be to add a `isStaticallyIdentical` option to `Value`s, which indicates if the item is known at static time to be an identical value. Then, if we attempt to call `GetOrCreateNode` with a candidate `Node` and that `Node` is not `isStaticallyIdentical` to the value currently in the map, we can error.

For `Operator` and `Storage`, `isStaticallyIdentical` would currently just be that the names (and in the case of `Storage`, types) match. We can default it to `false` for everything else until we need this concept there.

src/ir/value.h
markww: I'm using `As` in some tests, either in this PR or follow-on PRs. I don't think I'm using `If` at the moment, but it seems like a useful thing to have if we have `As`. While visitors are indeed a better pattern for full visitation of a tree, sometimes it is nice to just get the single-step downcast to the specific type of value.

src/ir/proto/sql/decode.cc
markww: This is now fixed.

src/ir/proto/sql/decode.cc
```diff
@@ -0,0 +1,38 @@
...
+  // Note: In the future, we will probably want to make a `Storage` per table
+  // and grab columns from that table using operators. For now, however,
+  // making each column an independent storage works for the MVP and is much
+  // simpler.
+  //
+  // Also, for now, we consider all storages to have primitive type. We will
+  // probably want to change that when we start handling types in a
+  // non-trivial fashion.
+  return ir::Value{value::StoredValue(
+      decoder_context.GetOrCreateStorage(column_path))};
```
bgogul, line 35: Don't need `ir::` prefix. Also, `{` should `(`?
```suggestion
  return Value(value::StoredValue(
      decoder_context.GetOrCreateStorage(column_path)));
```

src/ir/value.h
bgogul: A visitor would be ideal, but to unblock let us use `If<T>` variant. The `As<T>` variant can throw an exception and it is not obvious at use sites. With `If` variant, you will get a `nullptr`, and we can easily document use `CHECK_NOTNULL` to document expectations.

Also, add tests for `If` (in a follow up or separate PR?)

bgogul (approved):
Looks good assuming you get rid of the `As<T>` variant and replace uses of `As` with `If<T>` variant.

src/ir/value.h
markww: OK, that sounds good.

The issue with a `Visitor` here is that these are `Value`s; as such, the `Value`s that are the predecessors to this `Value` are not a tree, but a graph - the dataflow graph. Creating a `Visitor` on the `Value`s would be akin to performing a dataflow graph traversal, and I'm not sure it makes sense to do so when we want to lean on Souffle for that.

src/ir/proto/sql/decode.cc
markww: Done.

src/ir/value.h
markww: Using `If` instead of `As` in the test now.

PR #392, opened 2022-02-24
Decode `Literal` SQL IR protos.

markww:
This PR decodes `Literal` SQL IR protos. `Literal` protos are decoded
almost identically to `SourceTableColumn`s, even down to representing
them as `Storage`s. This is done because the IR does not currently have
a representation for `Storage`s and because in a verifier that does not
allow for modifications (as with the SQL verifier we are targeting),
there actually isn't that much difference between a named `Storage` and
a `Literal` associated with some canonical name.

### DIFF
--- a/src/ir/proto/sql/decode.cc
+++ b/src/ir/proto/sql/decode.cc
@@ -39,4 +39,32 @@ ir::Value DecodeSourceTableColumn(
               ir_context.type_factory().MakePrimitiveType())))};
 }
 
+ir::Value DecodeLiteral(const Literal &literal, IRContext &ir_context) {
+  const std::string &literal_str = literal.literal_str();
+  CHECK(!literal_str.empty()) << "required field literal_str was empty.";
+
+  // For now, we are abusing the `Storage` IR node to encode literals. This is
+  // not as crazy as it may initially look: just like a `Storage`, a `Literal`
+  // is a global name that should always return the same value if no
+  // modifications have been made to it. The only difference between a true
+  // `Storage` and a `Literal` is that it is illegal to modify a `Literal`.
+  // This difference is even less important for the SQL verifier we are
+  // currently trying to integrate with, as the SQL verifier considers a
+  // single query upon a table at a time. While this single query is running,
+  // the transactionality guarantees of the database should force the view onto
+  // the table to be constant through the life of that query, making the
+  // columns we are also capturing in `Storage`s constant anyway.
+  //
+  // Prefix the literal string with "literal:" to reduce the chance of a
+  // collision.
+  std::string qualified_literal_str =
+      absl::StrCat("literal:", literal_str);
+  return ir::Value{value::StoredValue(
+      ir_context.GetOrCreateStorage(
+        qualified_literal_str,
+        std::make_unique<Storage>(
+            qualified_literal_str,
+            ir_context.type_factory().MakePrimitiveType())))};
+}
+
 }  // namespace raksha::ir::proto::sql
--- a/src/ir/proto/sql/decode_test.cc
+++ b/src/ir/proto/sql/decode_test.cc
@@ -56,4 +56,36 @@ INSTANTIATE_TEST_SUITE_P(
     DecodeSourceTableColumnTest, DecodeSourceTableColumnTest,
     testing::ValuesIn(kSourceColumnPaths));
 
+class DecodeLiteralTest :
+    public testing::TestWithParam<absl::string_view> {};
+
+TEST_P(DecodeLiteralTest, DecodeLiteralTest) {
+  absl::string_view literal_str = GetParam();
+  std::string textproto = absl::StrFormat(R"(literal_str: "%s")", literal_str);
+  Literal literal_proto;
+
+  ASSERT_TRUE(google::protobuf::TextFormat::ParseFromString(
+      textproto, &literal_proto))
+      << "Could not decode Literal";
+
+  IRContext ir_context;
+  ir::Value result = DecodeLiteral(literal_proto, ir_context);
+  value::StoredValue stored_value = result.As<value::StoredValue>();
+  const Storage &storage = stored_value.storage();
+  EXPECT_EQ(storage.name(), absl::StrCat("literal:", literal_str));
+  EXPECT_EQ(storage.type().type_base().kind(),
+            types::TypeBase::Kind::kPrimitive);
+}
+
+absl::string_view kLiteralStrs[] = {
+    "100",
+    "0",
+    "3.1415",
+    "hello world!",
+    "null",
+};
+
+INSTANTIATE_TEST_SUITE_P(
+    DecodeLiteralTest, DecodeLiteralTest, testing::ValuesIn(kLiteralStrs));
+
 }  // namespace raksha::ir::proto::sql
--- a/src/ir/proto/sql/decode.h
+++ b/src/ir/proto/sql/decode.h
@@ -29,6 +29,8 @@ ir::Value DecodeSourceTableColumn(
     const SourceTableColumn &source_table_column,
     IRContext &ir_context);
 
+ir::Value DecodeLiteral(const Literal &literal, IRContext &ir_context);
+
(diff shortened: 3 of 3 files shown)


src/ir/proto/sql/decode.cc
```diff
@@ -35,4 +36,27 @@ Value DecodeSourceTableColumn(
       decoder_context.GetOrCreateStorage(column_path)));
 }
 
+Value DecodeLiteral(
```
bgogul, line 39: Why not use an `StringAttribute`?

src/ir/proto/sql/decode.cc
```diff
@@ -35,4 +36,27 @@ Value DecodeSourceTableColumn(
...
+    const Literal &literal, DecoderContext &decoder_context) {
+  const std::string &literal_str = literal.literal_str();
+  CHECK(!literal_str.empty()) << "required field literal_str was empty.";
+
+  // For now, we are abusing the `Storage` IR node to encode literals. This is
```
bgogul, line 44: Instead of using `Storage`, why don't you use a `const` operation with the string as a `StringAttribute`?

markww: Ah, that's a good point. I'll change the approach to do that.

markww: Done.

markww:
I think this is ready for another look @bgogul , thanks!

src/ir/proto/sql/decode_test.cc
```diff
@@ -58,4 +58,47 @@ INSTANTIATE_TEST_SUITE_P(
...
+      << "Could not decode Literal";
+
+  IRContext ir_context;
+  DecoderContext decoder_context(ir_context);
+  ir::Value result = DecodeLiteral(literal_proto, decoder_context);
+  const value::OperationResult *op_result = result.If<value::OperationResult>();
+  EXPECT_THAT(op_result, testing::NotNull());
+  EXPECT_EQ(op_result->name(), "out");
+  const Operation &operation = op_result->operation();
+  EXPECT_THAT(operation.parent(), testing::IsNull());
```
bgogul, line 80: Adding using declarations like `using testing::isNull;` in this file could help readability. e.g., 
```C++
EXPECT_THAT(operation.parent(), IsNull());
```
instead of
```C++ 
EXPECT_THAT(operation.parent(), testing::IsNull());
```

src/ir/proto/sql/decode.cc
```diff
@@ -35,4 +36,12 @@ Value DecodeSourceTableColumn(
       decoder_context.GetOrCreateStorage(column_path)));
 }
 
+Value DecodeLiteral(
+    const Literal &literal, DecoderContext &decoder_context) {
+  const absl::string_view literal_str = literal.literal_str();
+  CHECK(!literal_str.empty()) << "required field literal_str was empty.";
+  return Value(value::OperationResult(
+      decoder_context.MakeLiteralOperation(literal_str), "out"));
```
bgogul, line 44: Perhaps in a separate PR, I think it will be useful to have some default names for singleton results like `"out"`

bgogul (approved):
Mostly stylistic comments.

src/ir/proto/sql/decoder_context.h
```diff
@@ -46,10 +49,25 @@ class DecoderContext {
...
 
+  // Create an `Operation` providing the literal value indicated by
+  // `literal_str`. This shall create an `Operation` of `Operator` "sql.literal"
+  // and with the literal value indicated in a `literal_str` attribute.
+  const Operation &MakeLiteralOperation(absl::string_view literal_str) {
+    operations_.push_back(std::make_unique<Operation>(
+        /*parent=*/nullptr,
+        literal_operator_,
+        NamedAttributeMap{
+          {"literal_str", StringAttribute::Create(literal_str)}},
```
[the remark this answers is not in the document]
markww: Yes. The SQL verifier we are integrating with currently does not actually parse its literals. It just makes a note of the string from the original source file that describes the literal. We are continuing this treatment here.

src/ir/proto/sql/decoder_context.h
```diff
@@ -31,7 +31,10 @@ namespace raksha::ir::proto::sql {
 // inappropriate to put in IRContext.
 class DecoderContext {
  public:
-  DecoderContext(IRContext &ir_context) : ir_context_(ir_context) {}
+  DecoderContext(IRContext &ir_context)
+    : ir_context_(ir_context),
+      literal_operator_(ir_context_.RegisterOperator(
+          std::make_unique<Operator>("sql.literal"))) { }
```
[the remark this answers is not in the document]
markww: I can do the `constexpr` global to contain `"sql.literal"`. On the `string_literal` suggestion, see other comment: this `Operation` is for a literal of any type.

src/ir/proto/sql/decoder_context.h
```diff
@@ -46,10 +49,25 @@ class DecoderContext {
...
+        NamedValueMap{}));
+    return *operations_.back();
+  }
+
   IRContext &ir_context() { return ir_context_; }
 
  private:
   IRContext &ir_context_;
+  const Operator &literal_operator_;
+  std::vector<std::unique_ptr<Operation>> operations_;
```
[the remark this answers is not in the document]
markww: Sounds good, I will check this out.

src/ir/proto/sql/decode_test.cc
```diff
@@ -58,4 +58,47 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+  ASSERT_TRUE(google::protobuf::TextFormat::ParseFromString(
+      textproto, &literal_proto))
+      << "Could not decode Literal";
+
+  IRContext ir_context;
+  DecoderContext decoder_context(ir_context);
+  ir::Value result = DecodeLiteral(literal_proto, decoder_context);
+  const value::OperationResult *op_result = result.If<value::OperationResult>();
+  EXPECT_THAT(op_result, testing::NotNull());
```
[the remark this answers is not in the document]
markww: I thought there was, but I don't see it here: https://google.github.io/googletest/reference/assertions.html

src/ir/proto/sql/decode_test.cc
markww: Good point, will do.

src/ir/proto/sql/decoder_context.h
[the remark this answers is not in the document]
markww: Created the `constexpr` global.

src/ir/proto/sql/decode.cc
markww: I've moved all of the names that I'm using for SQL IR construction into `constexpr` globals in the `DecoderContext` class. It might be useful, however, to move this particular `constexpr` global into `Operation` itself in another PR.

src/ir/proto/sql/decode_test.cc

src/ir/proto/sql/decoder_context_test.cc
```diff
@@ -38,4 +38,31 @@ TEST(IRContextTest, GetOrCreateStorageIdempotence) {
...
+  EXPECT_THAT(lit_op.inputs(), testing::IsEmpty());
+  EXPECT_EQ(lit_op.op().name(), "sql.literal");
+
+  // Check that "attributes" is exactly "literal_str" associated with the
+  // parameter value.
+  const NamedAttributeMap &attributes = lit_op.attributes();
+  EXPECT_EQ(attributes.size(), 1);
+  auto find_result = attributes.find("literal_str");
+  EXPECT_NE(find_result, attributes.end());
+  EXPECT_EQ(find_result->second->ToString(), literal_str);
```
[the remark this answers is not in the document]
markww: Done.

src/ir/proto/sql/decoder_context.h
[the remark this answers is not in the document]
markww: Yes, I think this could be a good pattern. I'll move towards it in a follow-on PR.

PR #393, opened 2022-02-25
Add an IRPrinter class as a visitor.

bgogul:
This does not yet print all aspects of `Block` (e.g., inputs, outputs, and results). We will do this in follow up PRs as needed.
The goal of this PR is mainly to move the `Operation::ToString` logic  out to a visitor.


src/ir/ir_printer.h
```diff
@@ -0,0 +1,133 @@
...
+        << absl::StreamFormat(kOperationFormat, this_ssa_name,
+                              operation.op().name(), attributes_string,
+                              inputs_string);
+    if (operation.impl_module()) {
+      out << " {\n";
+      IncreaseIndent();
+    } else {
+      out << "\n";
+    }
+  }
```
markww, line 92: Rather than reporting the lack of a known implementation with the lack of a field, perhaps there should be some concise indication of the impl being unknown? Maybe something like `[impl_unknown]` or `= 0`.

src/ir/ir_printer.h
```diff
@@ -0,0 +1,133 @@
...
+
+  void PostVisit(const Operation& operation) override {
+    if (operation.impl_module()) {
+      DecreaseIndent();
+      out << Indent() << "}\n";
+    }
+  }
+
+ private:
+  std::string Indent() { return std::string(indent_ << 1, ' '); }
```
markww, line 102: I think `indent_ * 2` would be clearer than `indent_ << 1`. Especially in this context where `<<` might be initially mistaken for the stream operator.

src/ir/ir_printer.h
```diff
@@ -0,0 +1,133 @@
...
+
+// A visitor that prints the IR.
+class IRPrinter : public IRTraversingVisitor<IRPrinter> {
+ public:
+  template <typename T>
+  static std::string ToString(const T& entity) {
+    IRPrinter printer;
+    entity.Accept(printer);
+    return printer.out.str();
+  }
```
markww, line 37: I think, in the common case that this is printed to a log file or the console, it will be friendlier to have an interface like

```C++
static void ToString(const T& entity, ostream &out)
```

where `out` is used to construct the `IRPrinter` which ends up holding an `ostream &` field. That will allow everything to be printed out directly to the console/file rather than buffering in an `ostringstream` first. You can still have an alternative with the above interface by just having an implementation like this:

```C++
static std::string ToString(const T& entity) {
  std::ostringstream out;
  ToString(entity, out);
  return out.str;
}
```

src/ir/ir_printer_test.cc
```diff
@@ -0,0 +1,91 @@
...
+      nullptr, minus_op(),
+      {{"access", StringAttribute::Create("private")},
+       {"name", StringAttribute::Create("addition")}},
+      {{"const", Value(value::Any())}, {"arg", Value(value::Any())}});
+
+  EXPECT_EQ(IRPrinter::ToString(operation),
+            "%0 = core.minus [access: private, name: addition](arg: <<ANY>>, "
+            "const: <<ANY>>)\n");
+}
+
```
markww, line 89: I don't think we're exercising the case where an `Operation` has an `impl` `Module` in these tests. Should we add a test exercising that case?

markww (approved):
Looks good. Just have a suggestion about a possible interface change and additional test.

src/ir/ir_printer.h
bgogul: That was what I had in mind originally, but I did not like having to hold a reference to `ostream` because the constructor was public at that time. Now that the constructor is private, I think it is fine to do so. Will do.

src/ir/ir_printer.h
bgogul: I chose this representation to keep things compact as `impl_module() == nullptr` is the common case. Otherwise, every operation will have a `[impl_unknown]` or `=0` suffix.

src/ir/ir_printer.h
bgogul: Will do.

src/ir/ir_printer_test.cc
bgogul: `PrettyPrintsOperationWithProperIndentation` does that?

src/ir/ir_printer.h
bgogul: done.

src/ir/ir_printer_test.cc
markww: ok, sounds good.

PR #394, opened 2022-02-25
Add impl_module to BlockBuilder::AddOperation arguments

bgogul:


src/ir/block_builder.h
```diff
@@ -49,9 +49,11 @@ class BlockBuilder {
...
   const Operation& AddOperation(const Operator& op,
                                 NamedAttributeMap attributes,
-                                NamedValueMap inputs) {
-    block_->operations_.push_back(std::make_unique<Operation>(
-        block_.get(), op, std::move(attributes), std::move(inputs)));
+                                NamedValueMap inputs,
+                                std::unique_ptr<Module> impl_module = nullptr) {
+    block_->operations_.push_back(
+        std::make_unique<Operation>(block_.get(), op, std::move(attributes),
+                                    std::move(inputs), std::move(impl_module)));
```
markww, line 56: I remember previously that I said that I thought the `impl_module` should go on the `Operator` rather than the `Operation`, and you had a reason why it should go on the `Operation` instead, but I don't think we ever fleshed that out together. Why should the implementation live on the `Operation` rather than the `Operator`?

src/ir/block_builder.h
```diff
@@ -49,9 +49,11 @@ class BlockBuilder {
   // Adds an operation to the block and returns the operation.
   const Operation& AddOperation(const Operator& op,
                                 NamedAttributeMap attributes,
-                                NamedValueMap inputs) {
-    block_->operations_.push_back(std::make_unique<Operation>(
-        block_.get(), op, std::move(attributes), std::move(inputs)));
+                                NamedValueMap inputs,
+                                std::unique_ptr<Module> impl_module = nullptr) {
```
markww, line 53: Should we have this be a default argument, or would it be better to make callers explicitly state that the implementation is null?

src/ir/block_builder_test.cc
```diff
@@ -130,11 +130,16 @@ TEST_F(BlockBuilderTest, AddOperationUpdatesOperationList) {
       builder.AddOperation(context_.GetOperator("core.minus"), {}, {}));
   const Operation* merge_op = std::addressof(
       builder.AddOperation(context_.GetOperator("core.merge"), {}, {}));
+  const Operation* merge_op_with_module = std::addressof(builder.AddOperation(
+      context_.GetOperator("core.merge"), {}, {}, std::make_unique<Module>()));
```
markww, line 134: What does an empty implementation `Module` mean when we drop this to the analysis? Does it mean that the data doesn't go through the module at all? Or does it mean the default behavior of every input going to every output?

src/ir/block_builder.h
bgogul: Contrary to what the comment associated with `impl_module` suggest, the intention is not to break down an operation into smaller operations. Rather, this is a mechanism to associate lambdas with operations and was something that I pivoted to based on the proposal you had to model operations like `sql.select` and `sql.where` with lambdas.  I left the comment unchanged after the pivot and forgot this context. :( Anyway, having a mechanism to associate lambdas with operations simplifies the modeling of such operations as well other common ones like `if` and `while`.  In fact, it might make sense to have more than one module associated with an operation. e.g,. An `if` operation could have two modules correspnding to the `if` and `else` branches. However,  we can tackle it later.

src/ir/block_builder.h
bgogul: I think having the default argument as `nullptr` is fine as it it will be most common usage.

src/ir/block_builder_test.cc
bgogul: It is really up to the semantics of the operation. (See my other comment about the `impl_module` being a mechanism to associate lambda's with operation.)

src/ir/block_builder.h
bgogul: We should probably change the name of the field as well.

bgogul:
@markww, I replied to your comments. PTAL.

src/ir/block_builder.h
markww: Interesting, ok. So this is more akin to associating a dynamic call instruction with its possible static targets than associating a function name with its implementation.

It does seem like hanging a `Module` off of an `Operation` could be useful in that circumstance. You could even imagine a kind of attribute called `ResolutionNumberAttribute` that, for some named attribute indicating a resolution to be filled in, associates that attribute with a position in a `vector` of `Module`s. But as you said, that can be tackled later.

src/ir/block_builder.h
markww: OK, sounds good.

src/ir/block_builder_test.cc
markww: Sounds good.

markww (approved):
LGTM

PR #395, opened 2022-02-25
Make types::Type copyable by using intrusive_ptr.

bgogul:


src/ir/types/type.h
```diff
@@ -31,13 +46,13 @@ class Type {
   const TypeBase& type_base() const { return *type_.get(); }
 
   friend class TypeFactory;
+
  private:
-  Type(std::unique_ptr<TypeBase> type) : type_(std::move(type)) {}
+  Type(std::unique_ptr<TypeBase> type) : type_(type.release()) {}
```
markww, line 51: I'm not sure we should accept a `std::unique_ptr` as a parameter here. It seems like some confusing mixing of proofs. Perhaps `Type` should just accept a `TypeBase` value, and then the constructor for `type_` can just be

```C++
type_(new TypeBase(std::move(type))
```

src/ir/types/type.h
```diff
@@ -31,13 +46,13 @@ class Type {
   const TypeBase& type_base() const { return *type_.get(); }
```
markww, line 46: Is this PR the right time to add `operator*` and `operator->` instead of using `type_base`?

src/ir/types/type.h
bgogul: This is is a `unique_ptr` where we explicitly transfer ownership to the object? I agree this would be a problem if it is a `shared_ptr`. Moreover, https://en.cppreference.com/w/cpp/memory/unique_ptr/release explicitly releases ownership?

Also, this is a quick incremental change with having to change the places where `Type` is constructed.

src/ir/types/type.h
bgogul: Let us do it in a separate PR. I want to get rid of `type_base` if possible and `Type` as the type-erased (no pun intended :) ) version of `TypeBase`.

src/ir/types/type.h
markww: ok, sounds good then.

src/ir/types/type.h
markww: Sounds good :+1:

markww (approved):
LGTM

src/ir/types/type.h
bgogul: On second thought, even though this is a `unique_ptr`, it may not be a good idea to mix these two. Things will be OK as long as the pointer cleanly changes ownership from `unique_ptr` to `intrusive_ptr`. However, if these ever intertwine, it won't be pretty!

Nevertheless, let us get this as it is safe _currently_. We should fix this soon. I have filed #406.

PR #399, opened 2022-02-25
Add DecoderContext that tracks ID to value mapping.

markww:
Add a `DecoderContext` class that tracks the mapping from IDs to values.
This will be necessary for linking tag transforms to their inputs, which
are referenced by IDs instead of being nested directly under the
transform.


src/ir/proto/sql/decoder_context_test.cc
```diff
@@ -39,4 +39,30 @@ TEST(IRContextTest, GetOrCreateStorageIdempotence) {
...
+  EXPECT_EQ(&get_value2, &value2);
+
+  EXPECT_DEATH({ decoder_context.RegisterValue(1, Value(value::Any())); },
+               "id_to_value map has more than one value associated with the id "
+               "1.");
+  EXPECT_DEATH({ decoder_context.RegisterValue(2, Value(value::Any())); },
+               "id_to_value map has more than one value associated with the id "
+               "2.");
+  EXPECT_DEATH({ decoder_context.GetValue(3); },
+               "Could not find a value with id 3.");
```
bgogul, line 64: Split this out to a separate test:

```C++
using DecoderContextDeathTest = DecoderContextTest;
TEST(DecoderContextDeathTest, ...) {
  ...
}
```

src/ir/proto/sql/decoder_context_test.cc
```diff
@@ -39,4 +39,30 @@ TEST(IRContextTest, GetOrCreateStorageIdempotence) {
...
+
+  Storage storage("storage1", ir_context.type_factory().MakePrimitiveType());
+
+  const Value &value1 = decoder_context.RegisterValue(1, Value(value::Any()));
+  EXPECT_THAT(value1.If<value::Any>(), testing::NotNull());
+  const Value &value2 =
+      decoder_context.RegisterValue(2, Value(value::StoredValue(storage)));
+  const Value &get_value1 = decoder_context.GetValue(1);
+  const Value &get_value2 = decoder_context.GetValue(2);
+  EXPECT_EQ(&get_value1, &value1);
```
bgogul, line 54: Not related to this PR, but we should add a comparator for `Value`.

bgogul (approved):
Looks good on its own, but not sure where it fits. Left some style comments as well.

src/ir/proto/sql/decoder_context.h
```diff
@@ -50,6 +72,7 @@ class DecoderContext {
 
  private:
   IRContext &ir_context_;
+  absl::flat_hash_map<uint64_t, std::unique_ptr<Value>> id_to_value;
```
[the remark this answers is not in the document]
markww: I don't think so. The `uint64_t` IDs here are assigned in the SQL verifier frontend and used to link nodes at long range. There's no real guarantee that we will visit the nodes in the same order in both places, so we have to use a map.

src/ir/proto/sql/decoder_context.h
```diff
@@ -33,6 +33,28 @@ class DecoderContext {
  public:
   DecoderContext(IRContext &ir_context) : ir_context_(ir_context) {}
 
+  // Register a value as associated with a particular id. This stores that
+  // value and provides a reference to its canonical value.
+  const Value &RegisterValue(uint64_t id, Value value) {
+    auto insert_result =
+        id_to_value.insert({id, std::make_unique<Value>(std::move(value))});
```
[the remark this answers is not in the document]
markww: See other comment about why `std::vector` won't work here.

src/ir/proto/sql/decoder_context_test.cc
markww: Agreed.

src/ir/proto/sql/decoder_context_test.cc
markww: Done.

PR #401, opened 2022-02-28
Add decoding of SQL Expression proto.

markww:
The SQL Expression proto is the main class to be used for providing SQL
ASTs to Raksha. We can only partially implement it right now, but this
is necessary for handling the recursive cases (such as `TagTransform`
and `MergeOperation`).


src/ir/proto/sql/decode.h
```diff
@@ -25,6 +25,9 @@
 
 namespace raksha::ir::proto::sql {
 
+const Value &DecodeExpression(
+    const Expression &expr, DecoderContext &decoder_context);
+
```
bgogul, line 30: This comment is not related to this PR, but why can't we return an `Operation` for `DecodeSourceTableColumnXYZ` methods? Then we can wrap them in a `Value` as needed?

src/ir/proto/sql/decode_test.cc
```diff
@@ -113,4 +114,96 @@ absl::string_view kLiteralStrs[] = {
...
+  Expression expr;
+  ASSERT_TRUE(google::protobuf::TextFormat::ParseFromString(kTextproto, &expr))
+    << "Could not decode expr";
+  const ir::Value &result = DecodeExpression(expr, decoder_context);
+  const value::StoredValue *stored_value = result.If<value::StoredValue>();
+  EXPECT_THAT(stored_value, testing::NotNull());
+  const Storage &storage = stored_value->storage();
+  EXPECT_EQ(storage.name(), "table1.col");
+  EXPECT_EQ(storage.type().type_base().kind(),
+            types::TypeBase::Kind::kPrimitive);
```
bgogul, line 131: seems like a repeat of the `DecodeSourceColumnTableTest`s. Is there anyway to reuse them via typed parameterized or value parameterized tests?

src/ir/proto/sql/decode.h
markww: I'm using a `Storage` to represent the column, as that seemed like a pretty natural way to represent the source of the column (clearly a more mature system would represent the table with the `Storage` and use `Operation`s to grab columns out of that table). Once I chose a `Storage`, a `StoredValue` seemed like the natural way to represent the value coming out of that `Storage`.

Your question makes me wonder if `StoredValue` is necessary. If we introduced a new `StorageAttribute`, we could introduce a `load` `Operation`, hang the `Storage` from which it `load`ed onto the attribute, and just use the result of that `OperationResult` instead of `StoredValue`.

The way that we are leaning more and more on `Attribute`s is making me wonder if `Operator`s should further have on them a list of `expected_attribute`s, indicating the `Attribute`s that should be present upon an `Operation` associated with that `Operator` and the types we expect them to have. This could be useful for writing datalog rules to handle all `Operation`s associated with a particular `Operator` without constantly testing for `Attribute` presence.

src/ir/proto/sql/decode_test.cc
```diff
@@ -113,4 +114,96 @@ absl::string_view kLiteralStrs[] = {
...
+  EXPECT_EQ(&result, &decoder_context.GetValue(1));
+}
+
+TEST(DecodeLiteralExprTest, DecodeLiteralExprTest) {
+  const std::string kTextproto =
+      R"(id: 5 name: "alias" literal: { literal_str: "9" } )";
+  IRContext ir_context;
+  DecoderContext decoder_context(ir_context);
+  Expression expr;
```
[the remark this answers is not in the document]
markww: Sure, I can do that.

src/ir/proto/sql/decode_test.cc
markww: I'll see what I can do here.

markww: I think the source of this duplication is that I have been testing the variant decoding methods as if they were public-facing methods. But really, they shouldn't be: the only public-facing method will be the method that decodes an `Expression`, and the others should be `static` helper functions for decoding an `Expression`. I'll make those methods not part of the public API, which will remove the need to test them individually, and will test them only through `DecodeExpression`.

src/ir/proto/sql/decode.h
bgogul: Yes, the `Operator` should have metadata like `expected_attributes` and `expected_inputs`. It should not change the current `Operation` API and will mostly serve as a way to validate the instances of operations. In my initial version of the `Operator`, I did have some metadata but removed them as there were some unanswered questions and wanted to think it through. e.g. how do we model variadic arguments like `Merge` operation and so on.

Yes, it looks like we can get rid of `StoredValue` as well. We could even get rid of ANY and make it an operation. Then value will be quite simple  as it is either a block arg or an operation result.

src/ir/proto/sql/decode_test.cc
bgogul: SG. Please do that.

src/ir/proto/sql/decode.h
bgogul: Just to be clear, let us do this in  follow up PRs.

bgogul (approved):
Looks good modulo refactoring the tests.

src/ir/proto/sql/decode.h
markww: Sounds good, resolving.

src/ir/proto/sql/decode_test.cc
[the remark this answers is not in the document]

src/ir/proto/sql/decode_test.cc

PR #403, opened 2022-02-28
Remove unnecessary header value.h

markww:


bgogul (approved):
Do you need to update the corresponding BUILD file?

markww:
No, because the corresponding `BUILD` file didn't have the necessary dependency already :grimacing:

PR #411, opened 2022-03-02
Catch the case where a block builder is double-built.

markww:


bgogul (approved):
Thanks for taking care of this!

PR #415, opened 2022-03-03
Add `operator==` for the `Value` class and its children.

### LINKED ISSUE #337: Add comparator for values.

markww:
These check whether or not the `Value`s have identical representation,
not whether they evaluate to the same value (which we cannot really know
at static time).

This change fixes #337 .


markww:
Realized I missed an opportunity to take advantage of `std::variant`'s builtin `operator==`, update incoming.

src/ir/value.h
```diff
@@ -115,13 +139,22 @@ class Value {
   }
 
   // A downcast operation. Just delegates directly to std::get on variants.
-  template<class T>
-  const T &As() const { return std::get<T>(value_); }
+  template <class T>
```
bgogul, line 142: Didn't we decide to get rid of this method?

src/ir/value.h
```diff
@@ -115,13 +139,22 @@ class Value {
...
-  const T *If() const { return std::get_if<T>(&value_); }
+  template <class T>
+  const T* If() const {
+    return std::get_if<T>(&value_);
+  }
+
+  // Return whether the two `Value`s are identical representations. If they
+  // are different, this does not necessarily mean that they do not
+  // *evaluate* to the same value.
+  bool operator==(const Value& other) const { return value_ == other.value_; }
```
bgogul, line 157: I presume this takes care of the variant kinds as well?

src/ir/value.h
```diff
@@ -63,13 +73,19 @@ class BlockArgument : public NamedValue<Block> {
...
+    return NamedValuesEq(*this, other);
+  }
 };
 
 // Indicates the result of an operation.
 class OperationResult : public NamedValue<Operation> {
  public:
   using NamedValue<Operation>::NamedValue;
   const Operation& operation() const { return element(); }
+  bool operator==(const OperationResult& other) const {
```
bgogul, line 86: Not a big deal, but you might be able to inherit the operator from `NamedValue<T>` with an `using` declaration.

bgogul (approved):

src/ir/value.h
markww: Yes. From here: https://en.cppreference.com/w/cpp/utility/variant/operator_cmp

"""
If v.index() != w.index(), returns false;
"""

where `index()` returns the index of the alternative in the template parameter pack of alternatives.

src/ir/value.h
markww: Maybe. It does seem that `If` is proving to be more useful. I can do that in a separate PR; the changes here only occurred due to formatting.

src/ir/value.h
markww: I like this suggestion, I'm doing it.

markww: Done.

PR #419, opened 2022-03-03
Make `Value` ctor `explicit` and more concise.

markww:
Previously, the `Value` ctor was not explicit, which meant that the
various `Value` variants could implicitly become a `Value`. This is not
encouraged by the C++ style guide.

Further, this commit takes the various constructors defined for each
`variant` alternative and provides a single constructor that takes as
its argument the `variant` type. Because a given `variant` defines an
implicit conversion from each of its variants to itself, this is a more
concise way of accomplishing the same thing. To keep the restatement of
the `variant` type in sync between the field and the ctor, define
a `ValueVariant` `using` alias to use in both the field and the ctor
parameter.


src/ir/value.h
```diff
@@ -101,10 +101,11 @@ class Any {
 // A class that represents a data value.
 class Value {
  public:
-  Value(value::BlockArgument arg) : value_(std::move(arg)) {}
-  Value(value::OperationResult arg) : value_(std::move(arg)) {}
-  Value(value::Any arg) : value_(std::move(arg)) {}
-  Value(value::StoredValue arg) : value_(std::move(arg)) {}
+  using ValueVariants =
```
bgogul, line 104: Could simply be `Variants` as in `Value::Variants`.

bgogul (approved):

src/ir/value.h
markww: Done.

PR #420, opened 2022-03-03
Make canonical `Value`s be returned by value, not by reference.

markww:
Previously, we held canonical `Value`s in a map of `unique_ptr`s and
returned pointers/references to these `Value`s. This caused problems
when we wished to place these `Value`s into an `InputValueMap`, which
expects to place value-typed `Value`s as the value of the map.

This commit makes us just return the canonical `Value`s as value types.
This causes a copy when we return them from the map. This should not be
a huge problem: the largest `Value`s currently contain a `std::string`
and a pointer. The most naive encoding of a `std::string`, containing a
pointer to the string contents, a `size_t` size, and a `size_t`
capacity, is 3 register widths. Add the other pointer, and `Value` is
4 register widths in size. Not ideal for copying, but not bad either.

If we wanted to make this copy even lighter-weight, we could change the
`std::string` on `Value` to an `absl::string_view`, or even have an
register-width `Identifier` signifier that could be looked up in a map
when needed. But it's likely that these optimizations won't be
necessary: the `id_to_value` map is used only to link SQL
`TagTransform`s to the `Value`s that they must consider, which will only
happen once and only for the number of inputs to said transforms.
I don't think that will be our inner loop, and will likely not cause
performance issues.


src/frontends/sql/decode.cc
```diff
@@ -74,12 +74,14 @@ static Value GetExprValue(const Expression &expr,
   CHECK(false) << "Unreachable!";
 }
 
-const Value &DecodeExpression(const Expression &expr,
-                              DecoderContext &decoder_context) {
+Value DecodeExpression(const Expression &expr,
```
bgogul, line 77: Is memory safety the reason for this change?

bgogul (approved):

src/frontends/sql/decode.cc
markww: It's more about the fact that `NamedValueMap` has value-typed `Value`s as its value and that `Value` is pretty small and naively-copyable anyway.

PR #421, opened 2022-03-04
Decode the `MergeOperator`.

markww:
The `MergeOperator` just decodes into a single merge operator, with the
direct and control inputs respectively provided via names with a common
prefix.


src/frontends/sql/decode_test.cc
```diff
@@ -202,9 +203,104 @@ const TextprotoDeathMessagePair kTextprotoDeathMessagePairs[] = {
...
+};
+
+TEST_P(DecodeMergeOpTest, DecodeMergeOpTest) {
+  auto &[opt_name, direct_size, control_size] = GetParam();
+
+  Expression expr;
+  EXPECT_TRUE(
+      google::protobuf::TextFormat::ParseFromString(GetTextproto(), &expr))
+      << "Could not decode expr";
+  Value value = DecodeExpression(expr, decoder_context_);
```
bgogul, line 269: Why are you not using `GetDecodedValue()`?

src/frontends/sql/decode_test.cc
```diff
@@ -202,9 +203,104 @@ const TextprotoDeathMessagePair kTextprotoDeathMessagePairs[] = {
...
+  EXPECT_EQ(value, decoder_context_.GetValue(1));
```
bgogul, line 270: Why `1` in `GetValue(1)`?

src/frontends/sql/decode_test.cc
```diff
@@ -202,9 +203,104 @@ const TextprotoDeathMessagePair kTextprotoDeathMessagePairs[] = {
...
+  const Block &top_level_block = decoder_context_.BuildTopLevelBlock();
```
bgogul, line 271: Just a new line to separate checking of expectation from set up .

src/frontends/sql/decode_test.cc
```diff
@@ -202,9 +203,104 @@ const TextprotoDeathMessagePair kTextprotoDeathMessagePairs[] = {
...
+  EXPECT_THAT(op.attributes(), IsEmpty());
+  std::vector<std::string> expected_keys;
+  for (uint64_t i = 0; i < direct_size; ++i) {
+    expected_keys.push_back(
+        absl::StrCat(DecoderContext::kMergeOpDirectInputPrefix, i));
+  }
+  for (uint64_t i = 0; i < control_size; ++i) {
+    expected_keys.push_back(
+        absl::StrCat(DecoderContext::kMergeOpControlInputPrefix, i));
+  }
```
bgogul, line 287: Let us inline the strings as test parameters. It makes the test easier to understand and will make it easy to debug in case of failures: 
```C++
Combine(
  ValuesIn(kSampleExprNames),
  ValuesIn({"direct_input_0"}, {"direct_input_0", "direct_input_1", ... "direct_input_4"}, {...}),
  ValuesIn({}, {"control_input_0"}, ...))
```

src/frontends/sql/decoder_context_test.cc
```diff
@@ -127,4 +135,95 @@ TEST(DecoderContextDeathTest, DecoderContextValueRegistryDeathTest) {
                "Could not find a value with id 3.");
 }
 
+class DecodeMergeOpTest
+    : public TestWithParam<std::tuple<std::vector<Value>, std::vector<Value>>> {
+};
+
+static bool BeginsWith(absl::string_view str1, absl::string_view str2) {
```
bgogul, line 142: Use `absl::StartsWith`? https://github.com/abseil/abseil-cpp/blob/04bde89e5cb33bf4a714a5496fac715481fc4831/absl/strings/match.h#L58

src/frontends/sql/decoder_context_test.cc
```diff
@@ -127,4 +135,95 @@ TEST(DecoderContextDeathTest, DecoderContextValueRegistryDeathTest) {
                "Could not find a value with id 3.");
 }
 
+class DecodeMergeOpTest
```
bgogul, line 138: What is the difference between the test here and the test with the same name  in `decode_context_test`?

src/frontends/sql/decoder_context_test.cc
```diff
@@ -127,4 +135,95 @@ TEST(DecoderContextDeathTest, DecoderContextValueRegistryDeathTest) {
...
+static std::optional<uint64_t> ExtractIdxAfterPrefix(absl::string_view str,
+                                                     absl::string_view prefix) {
+  if (!BeginsWith(str, prefix)) return std::nullopt;
+  uint64_t result = 0;
+  std::string index_str(str.substr(prefix.size()));
+  EXPECT_NO_THROW({ result = std::stoul(index_str); });
+  return result;
+}
+
+TEST_P(DecodeMergeOpTest, DecodeMergeOpTest) {
```
bgogul, line 156: Ideally, unit tests should not have this much logic as it makes debugging failures very difficult. Can we use any of the gmock matchers?

src/frontends/sql/decoder_context_test.cc
```diff
@@ -127,4 +135,95 @@ TEST(DecoderContextDeathTest, DecoderContextValueRegistryDeathTest) {
...
+                                  dummy_operation_result);
+  std::vector found_control_values(found_input_map.size(),
+                                   dummy_operation_result);
+
+  uint64_t num_direct_inputs_found = 0;
+  uint64_t num_control_inputs_found = 0;
+  for (auto &[key, value] : found_input_map) {
+    if (std::optional<uint64_t> direct_idx = ExtractIdxAfterPrefix(
+            key, DecoderContext::kMergeOpDirectInputPrefix)) {
+      found_direct_values.at(*direct_idx) = value;
```
bgogul, line 190: `found_direct_values[*direct_idx] = value`?

src/frontends/sql/decoder_context_test.cc
```diff
@@ -127,4 +135,95 @@ TEST(DecoderContextDeathTest, DecoderContextValueRegistryDeathTest) {
...
+  const NamedValueMap &found_input_map = op.inputs();
+
+  std::vector found_direct_values(found_input_map.size(),
+                                  dummy_operation_result);
+  std::vector found_control_values(found_input_map.size(),
+                                   dummy_operation_result);
+
+  uint64_t num_direct_inputs_found = 0;
+  uint64_t num_control_inputs_found = 0;
+  for (auto &[key, value] : found_input_map) {
```
bgogul, line 187: Not testing if the created merge operation has more inputs than it was passed?

src/frontends/sql/decoder_context.cc
```diff
@@ -0,0 +1,36 @@
+#include "src/frontends/sql/decoder_context.h"
+
+namespace raksha::frontends::sql {
+
+using ir::NamedValueMap;
+using ir::Operation;
+using ir::Value;
+
+static NamedValueMap MakeNamedValueMapFromVector(absl::string_view prefix,
```
bgogul, line 9: Also, MLIR code base has a pattern that would help here and elsewhere. Basically create a class as follows:

```
class MergeOperator : public Operator {
public:
  std::unique_ptr<Operation>  MakeOperation(std::vector<Value> inputs, std::vector<Value> outputs);
  // The logic that direct inputs have  "direct_input_" prefix will be capture here. 
  ValueRange GetDirectInputs(const Operation& operation) {
   } 

};
```
Then, we can use it to create and manipulate relevant operations in a more straightforward way:

```
MergeOperator::CreateOperation(....);
...
MergeOperator::GetControlInputs(operation);
```

Not suggesting you make these changes in this PR, but seems like a useful mechanism to build.

src/frontends/sql/BUILD
```diff
@@ -71,6 +72,7 @@ cc_library(
 cc_test(
     name = "decoder_context_test",
     srcs = ["decoder_context_test.cc"],
+    copts = ["-fexceptions"],
```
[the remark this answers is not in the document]
markww: It is because `std::stoul` throws an exception when it fails to parse a number from a string. I use `std::stoul` in this test and want to make sure it does not fail to parse a string as a number, and so I use `EXPECT_NO_THROW`, which requires exceptions to be enable.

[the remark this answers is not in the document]
markww: It looks like absl has a non-exception-throwing equivalent I can use, will refactor to eliminate `-fexceptions`.

src/frontends/sql/decoder_context.cc
markww: Interesting, that does seem useful! In MLIR, is this primarily used as a test helper or as an implementation thing? I think I might add this, but as a `testonly` class right now to prevent from having to write yet-more tests for it. If we ever decide that we want to use it in actual non-test code, we can unit test it then. Considering the MVP nature of the current work, I think that it's more likely to be stripped out entirely than to be moved to non-test status.

src/frontends/sql/decoder_context_test.cc
markww: `operator[]` is not bounds-checked in `std::vector`. As such, I never use it unless forced to by performance profiling.

src/frontends/sql/BUILD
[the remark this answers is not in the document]
markww: Done.

src/frontends/sql/decoder_context_test.cc
markww: Done.

src/frontends/sql/decode_test.cc
markww: I fixed the `ID` at `1` for the `MergeOperator` because, unlike `Literal` and `SourceTableColumn`, `MergeOperator` is an inner node and will have terminals under it with their own `ID`s. Therefore, I could not make the parameter dynamically chosen as a test parameter without risking a collision. Instead, I made the `MergeOperator` in this test always have the `ID` 1, and did not add the `ID` as one of the test parameters.

src/frontends/sql/decode_test.cc
markww: It isn't defined/declared for `DecodeMergeOpTest`. I used the same test pattern for `Literal`s and `SourceTableColumn`s because they were both terminals having about the same form. But there are difficulties that pop up for inner nodes when trying to use that pattern; see the comment about why the number `1` is fixed in the test.

src/frontends/sql/decoder_context_test.cc
markww: One tests the creation of the object from a proto, the other the creation of the object from input vectors. Using input vectors as inputs is more flexible and easier to test. But what I might do is create the `MergeOperator` helper class in the style that you recommended, put a `operator==` on it, and use that to remove some of the duplicated logic.

src/frontends/sql/decoder_context_test.cc
markww: `operator[]` on `vector`s is not bounds checked, so I try to use `at` instead.

src/frontends/sql/decode.cc
```diff
@@ -49,6 +49,34 @@ static Value DecodeLiteral(const Literal &literal,
...
+  // regular input is provided.
+  CHECK(merge_operation.inputs_size() >= 1)
+      << "Each MergeOperation is expected to have at least one non-control "
+         "input.";
+  std::vector<Value> direct_input_values;
+  direct_input_values.reserve(merge_operation.inputs_size());
+  for (Expression input_expr : merge_operation.inputs()) {
+    direct_input_values.push_back(
+        DecodeExpression(input_expr, decoder_context));
+  }
```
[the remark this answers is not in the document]
markww: I don't think that will work here, because `inputs` does not return a `std::vector`. It instead returns a `RepeatedPtrField`. But I will keep this utility in mind for the future.

src/frontends/sql/decoder_context_test.cc
markww: This test now has a very different form, resolving.

src/frontends/sql/decoder_context.cc
markww: I've created a class `MergeOperationView` that does something similar to this, but with composition rather than inheritance. In some ways, the inheritance-based solution would be better, but this prevents from having to modify the `BlockBuilder` and other classes that expect a regular `Operation`.

src/frontends/sql/decoder_context_test.cc
markww: A lot of the common logic between these tests has been moved to `testing::MergeOperationView`, so these tests should be more distinct now.

src/frontends/sql/decode_test.cc
markww: These lines no longer exist in the test, due to the refactoring with `MergeOperationView`.

src/frontends/sql/decode_test.cc
```diff
@@ -202,9 +203,104 @@ const TextprotoDeathMessagePair kTextprotoDeathMessagePairs[] = {
...
+        << "Could not decode expr";
+    return DecodeExpression(expr, decoder_context_);
+  }
+
+  IRContext ir_context_;
+  DecoderContext decoder_context_;
+};
+
+TEST_P(DecodeMergeOpTest, DecodeMergeOpTest) {
+  auto &[opt_name, direct_size, control_size] = GetParam();
```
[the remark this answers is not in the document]
markww: Renamed (to `optional_name`, not `operator_name`, which basically proves your point here) and moved.

src/frontends/sql/decode_test.cc
markww: Added a comment explaining this.

src/frontends/sql/decode_test.cc
markww: Done.

src/frontends/sql/decoder_context_test.cc
markww: This has been slimmed down considerably, as most of the logic has moved to `testing::MergeOperationView`.

src/frontends/sql/decode_test.cc
```diff
@@ -202,9 +203,104 @@ const TextprotoDeathMessagePair kTextprotoDeathMessagePairs[] = {
...
      .death_message = "Required field expr_variant not set."},
+    {.textproto = R"(id: 1 merge_operation: { })",
+     .death_message = "Each MergeOperation is expected to have at least one "
+                      "non-control input."},
 };
 
 INSTANTIATE_TEST_SUITE_P(DecodeExprDeathTest, DecodeExprDeathTest,
                          ValuesIn(kTextprotoDeathMessagePairs));
 
+// We test the decoding of the child `Operation`s in other tests. We also
```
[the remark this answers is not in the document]
markww: I am sharing the test format `IdNameAndStringTest` between `DecodeSourceTableColumnExprTest` and `DecodeLiteralExprTest`. That said, that's not much code and could be split off into its own header. I think splitting these up could make sense, but I think it would increase the footprint of this PR significantly, so I'd rather do it as a follow-on PR if that's ok.

markww:
I think this is ready for another look, @bgogul .

src/frontends/sql/decode_test.cc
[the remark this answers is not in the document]
bgogul: SG, but let us get it done in a follow up?

bgogul (approved):

src/frontends/sql/BUILD
```diff
@@ -37,7 +37,9 @@ cc_proto_library(
 
 cc_library(
     name = "decoder_context",
+    srcs = ["decoder_context.cc"],
     hdrs = ["decoder_context.h"],
+    visibility = [":__subpackages__"],
```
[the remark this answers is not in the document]
markww: I need it so that the `MergeOperationView` in the `testing` subpackage can have access to the string constants on `DecoderContext`. `MergeOperationView` doesn't need to see anything else from the `//src/frontends/sql` package though.

src/frontends/sql/decode_test.cc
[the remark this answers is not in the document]
markww: Absolutely, will do as my next PR.

PR #425, opened 2022-03-04
Remove `:decoder_context`'s dependency on SQL protos.

markww:
The `DecoderContext` class helps with the decoding of SQL protos, but it
should not know anything about the decoding of those SQL protos. That
should be the sole job of the SQL frontend's `:decode` library. This
change removes an unused dependency upon these protos in
`:decoder_context`.


bgogul (approved):
Oops! This fell through the cracks.

markww:
No prob :)

PR #386, opened 2022-02-22
Handling multiverses

aferr:

PR #426, opened 2022-03-07
[Authorization Logic :: Rust] Comparisons on Numbers

aferr:
This PR supports comparisons on numbers (such as "<") which is needed to launch for transparent release and is also useful in general. A test that closely follows the use-cases involving delegation and making use of the PR that supports populating universes across principals is given.


aferr:
This depends on [Handling multiverses #386](https://github.com/google-research/raksha/pull/386).

PR #427, opened 2022-03-07
[Authorization Logic :: Rust] Comparisons on Numbers

aferr:
This PR supports comparisons on numbers (such as "<") which is needed to launch for transparent release and is also useful in general. A test that closely follows the use-cases involving delegation and making use of the PR that supports populating universes across principals is given.


aferr:
OK, I can't seem to get github to drop the commits from the PR this depends on even after I rebase and attempt to force push... And since there are no reviews/discussions on the last PR anyway in this case I'll just close https://github.com/google-research/raksha/pull/426 and we can use this one instead.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -120,10 +139,19 @@ ATTRIBUTE: 'attribute';
 
 // Identifiers wrapped in quotes are constants whereas
 // identifiers without quotes are variables.
-ID : ('"')? [_a-zA-Z0-9/.#:]* ('"')?;
+ID : ('"')? [_a-zA-Z]+ [_a-zA-Z0-9/.#:]* ('"')?;
+NUMLITERAL : [0-9]+;
```
markww, line 143: Do you need negatives? Or is that handled elsewhere?

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -49,14 +49,15 @@ fn construct_predicate(ctx: &PredicateContext) -> AstPredicate {
...
-    let name_ = ctx.ID(0).unwrap().get_text();
+    let name_ = ctx.ID().unwrap().get_text();
     let mut args_ = Vec::new();
-    let mut idx = 1;
-    while let Some(id) = ctx.ID(idx) {
-        args_.push(id.get_text());
+    let mut idx = 0;
+    while let Some(arg) = ctx.pred_arg(idx) {
+        args_.push(arg.get_text());
         idx += 1;
```
markww, line 61: Rather than having `mut` locals and using a `while` loop, would it be possible to refactor this to use `map` on `pred_arg` to produce a `Vec` and remove the mutation?

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -136,13 +131,41 @@ fn construct_can_say_fact(ctx: &CanSayFactContext) -> AstFact {
...
+        RvalueContextAll::FlatFactRvalueContext(ffctx) => {
+            AstRValue::FlatFactRValue { 
+                flat_fact: construct_flat_fact(&ffctx.flatFact().unwrap())
+            }
+        },
+        RvalueContextAll::BinopRvalueContext(bctx) => {
+            AstRValue::BinopRValue {
+                lnum: bctx.pred_arg(0).unwrap().get_text(),
+                binop: construct_binop(&bctx.binop().unwrap()),
+                rnum: bctx.pred_arg(1).unwrap().get_text()
```
markww, line 157: I'm seeing a lot of `unwrap`s here; what is being `unwrap`ed? Should it be handled in a more principled fashion rather than `panic`ing if it is not present?

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -154,8 +177,8 @@ fn construct_fact_assertion(ctx: &FactAssertionContext) -> AstAssertion {
 fn construct_hornclause(ctx: &HornClauseAssertionContext) -> AstAssertion {
     let lhs = construct_fact(&ctx.fact().unwrap());
     let mut rhs = Vec::new();
-    for flat_fact_ctx in ctx.flatFact_all() {
-        rhs.push(construct_flat_fact(&flat_fact_ctx));
+    for rvalue_ctx in ctx.rvalue_all() {
+        rhs.push(construct_rvalue(&rvalue_ctx));
     }
```
markww, line 182: Should this use a higher-ordered function like `map` rather than mutation?

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -72,10 +72,26 @@ pub enum AstFact {
...
+    Equals,
+    NotEquals,
+    LessOrEquals,
+    GreaterOrEquals
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum AstRValue {
+    FlatFactRValue { flat_fact: AstFlatFact},
+    BinopRValue { lnum: String , binop: AstBinop, rnum: String }
```
markww, line 88: I think that numbers are different from, and more-special-than, regular `String`s and should be represented as such in the AST. You should have a special `struct` wrapper, something like `NumberLiteral`, and put the representation of the number in there. If we're not doing anything number-y with the `NumberLiteral`, a `String` is fine, but usually it's helpful to parse them and have the numeric interpreation in a first-class `int64` (or whatever is appropriate).

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
```diff
@@ -18,14 +18,20 @@
 //! this authorization logic into datalog simpler.
 use crate::ast::*;
 
+#[derive(Clone)]
+pub enum DLIRRvalue {
+    PredicateRvalue { predicate: AstPredicate },
+    BinopRValue { lnum: String, binop: AstBinop, rnum: String }
```
markww, line 24: It looks like this alternative is the same as the one in `AstRValue`. Should the `BinopRValue` definition be extracted and shared?

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
```diff
@@ -102,6 +102,10 @@
 use crate::{ast::*, souffle::datalog_ir::*};
 use std::collections::HashMap;
 
+fn pred_to_dlir_rvalue(pred: &AstPredicate) -> DLIRRvalue {
+    DLIRRvalue::PredicateRvalue { predicate: pred.clone() }
+}
```
markww, line 107: You could make this method require an `AstPredicate` rather than an `&AstPredicate`. That would allow you to skip cloning at call sites where the `pred` isn't needed for anything anymore anyway, which could save some copies. Also, might make the borrow checker easier to placate at callsites.

rust/tools/authorization-logic/src/test/test_num_compare.rs
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2021 Google LLC.
```
markww, line 2: New file; should this be 2022?

rust/tools/authorization-logic/src/souffle/universe_translation.rs
```diff
@@ -174,6 +187,25 @@ fn populate_constant_type_environment_prog(
     constant_type_environment
 }
 
+fn populate_constant_type_environment_rvalue(
+    relation_type_environment: &RelationTypeEnv,
+    constant_type_environment: &mut HashMap<String, AstType>,
```
markww, line 192: It would be more functional to take the map as a value and return the populated map rather than taking a mutable reference and returning void if that's possible.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -120,10 +139,19 @@ ATTRIBUTE: 'attribute';
 
 // Identifiers wrapped in quotes are constants whereas
 // identifiers without quotes are variables.
-ID : ('"')? [_a-zA-Z0-9/.#:]* ('"')?;
+ID : ('"')? [_a-zA-Z]+ [_a-zA-Z0-9/.#:]* ('"')?;
```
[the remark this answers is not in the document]
aferr: >  If so, you should drop the +

Yes, that's what I actually meant. Thanks! I had it this way a while ago but changed to what it was just prior to this PR as other quirky bits were needed in IDs for various use-cases. Good catch. It's fixed now.

>Also, many regex languages have special codes for common things like alphanumerics and digits

ANTLR does not have this. I actually originally looked into ANTLRs suggested way of describing the lex rule for variables. (And IIRC they gave the one that you just suggested now minus the other quirky characters we need).

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: It's not needed for the transparent release demo certainly. I don't recall seeing that souffle supports negative numeric literals. If souffle does not support this, I don't want to bother coming up with some way to support them now given that we don't need them yet. (Supporting negatives could be useful if we keep some extension of this grammar around maybe)

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
aferr: fixed

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
aferr: The antlr listener API that returns a `context` for the subexpressions of the parse rule returns an option type. Unwrapping is fine because this is a parse error if it tries to unwrap a `None` in this case, so panicking is right -- though if this were code that was going to live for a while giving better error messages rather than what the default panic gives you would be an improvement. (Though I think actually if there is a parse error it will cause a panic with a better descrption before it reaches this branch (or the other similar branches for which I give the default panics)).

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
aferr: fixed

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
aferr: The `BinOpRvalue` branch is the same yes, but the overall definition of `RValue` is different because here the other branch is a predicate but in the other case it's a flatfact. 

I guess it could be split into a separate class, but I think other parts of the code would become more verbose. I also think there is some merit to going the other way and making the other class which is shared between them (`AstPredicate`) distinct because then the layers of the IR are totally separate and then you get type errors if you try to run a compiler pass that was meant for one layer of the IR on another.

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
aferr: This gets called inside `map` combinators, and in rust the arguments to these are `&` so that makes things more awkward and doesn't allow the clone to be skipped (it just requires cloning in the map).

rust/tools/authorization-logic/src/test/test_num_compare.rs
aferr: I'm still not ready.

rust/tools/authorization-logic/src/souffle/universe_translation.rs
aferr: I think making this overall collection of functions work that way would entail making `add_typing` do the same, which would entail making a copy of the entire hashmap inside `add_typing` because the standard libarary calls for updating hashmaps work by mutation rather than by returning a value and passing ownership to the return.

rust/tools/authorization-logic/src/ast.rs
aferr: These can also be variables not just numeric literals.

markww: Doesn't that make it even more important to parse, discover, and add structured types around whatever lies underneath this `BinOpRValue`?

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
```diff
@@ -314,6 +322,26 @@ impl LoweringToDatalogPass {
...
+                pred_to_dlir_rvalue(&push_prin(String::from("says_"),
+                    &speaker, &flat_pred))
+            }
+            AstRValue::BinopRValue { lnum, binop, rnum } => {
+                DLIRRvalue::BinopRValue { 
+                    lnum: lnum.clone(), 
+                    binop: *binop, 
+                    rnum: rnum.clone() 
+                }
+            }
```
markww, line 342: Note that if `AstRValue` and `DLIRRvalue` shared the same subclass, this would just be something like:

```Rust
AstRValue::BinOp(binop_rval) => DLIRRvalue::BinopRValue { binop: binop_rval.clone() }
```

rust/tools/authorization-logic/src/ast.rs
aferr: A separate AST node for `pred_args` (which can be arguments to either predicates or these comparisons) does seem better. I opened an issue to track this: https://github.com/google-research/raksha/issues/429. Because the point of this rust implementation is to just work briefly for a demo, it's probably more important that this improvement makes its way into the C++ version. The issue can track both implementations.

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
markww, line 24: It looks like there is inconsistent casing of `Rvalue`/`RValue` here.

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
aferr: There's no subclasses in Rust, just traits.

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
markww: I've looked around, and I don't think it would be that much more verbose. I think it would end up saving space if you end up adding more structure under `lnum` and `rnum`. That said, none of this code is going to live very long, so I'll defer to your judgement on this one.

rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs
markww: Sorry, not subclass, `struct` embedded within analogous branch of the `enum`.

aferr: OK I get it now thanks. I took your other suggestion which is relevant and this is fixed now.

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
aferr: fied

aferr: fixed

rust/tools/authorization-logic/src/ast.rs
markww: OK, that works for me :+1:

markww (approved):
LGTM

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -54,6 +54,13 @@ impl PartialEq for AstPredicate {
 // See https://doc.rust-lang.org/std/cmp/trait.Eq.html
 impl Eq for AstPredicate {}
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AstBinaryOperation {
```
bgogul, line 58: nit: Name could reflect it is an arithmetic operator. e.g., `AstArithmeticComparison` (Similarly for others related structs.)

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -72,10 +79,26 @@ pub enum AstFact {
...
+    GreaterThan,
+    Equals,
+    NotEquals,
+    LessOrEquals,
+    GreaterOrEquals
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum AstRValue {
+    FlatFactRValue { flat_fact: AstFlatFact},
```
bgogul, line 94: What is a `FlatFact`?

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -72,10 +79,26 @@ pub enum AstFact {
...
+    LessThan,
+    GreaterThan,
+    Equals,
+    NotEquals,
+    LessOrEquals,
+    GreaterOrEquals
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum AstRValue {
```
bgogul, line 93: What is an `RValue`?

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
```diff
@@ -18,14 +18,20 @@
 //! this authorization logic into datalog simpler.
 use crate::ast::*;
 
+#[derive(Clone)]
+pub enum DLIRRValue {
+    PredicateRValue { predicate: AstPredicate },
+    BinopRValue { binop: AstBinaryOperation }
```
bgogul, line 24: This comment is mostly about naming. It looks `DLIRRValue` is actually a predicate and you have two kinds of predicates:
   - Datalog Relations.
   - Arithmetic comparisons. 

Why can't you call it `DLIRPredicate` or something like that?

rust/tools/authorization-logic/src/ast.rs
```diff
@@ -72,10 +79,26 @@ pub enum AstFact {
     AstCanSayFact { p: AstPrincipal, f: Box<AstFact> },
 }
 
+#[derive(Copy, Debug, Clone, Serialize, Deserialize)]
+pub enum AstOperator {
```
bgogul, line 83: `AstComparsionOperator`?

bgogul (commented):
My comments are mostly around naming of the newly-introduced struct. Either name them appropriately or document them well.

rust/tools/authorization-logic/src/ast.rs
aferr: FlatFact has been here a while. This is the part of the grammar of facts that can't be nested --  they do not contain `canSay` which is the part that can be nested. These need to be separated at the grammar level because allowing nesting on the RHS would make the language undecidable (and untranslatable to datalog).

rust/tools/authorization-logic/src/ast.rs
aferr: Here a `RValue`s are the parts of the grammar that can appear on the RHS. The names `rvalue` and `lvalue` are commonly used in programming language designs to describe parts of grammars that can appear on the right hand side and left hand side of assignment statements respectively. So for example in `x := y + 7` the rvalue is `x` and the lvalue is `y + 7`. [C++ is one example of a language that used to use this terminology -- the grammar has gotten more complicated though apparently](https://stackoverflow.com/questions/9406121/what-exactly-is-an-r-value-in-c#:~:text=In%20C%2B%2B%2C%20an%20rvalue,which%20is%20not%20a%20reference.&text=In%20(1)%2C%20the%20temporary,string%20literal%20is%20an%20rvalue.) It's not just C++ that uses this terminology in its language design, though.

So here the name rvalue is used to describe the part of the grammar that can appear on the RHS of a conditional horn clause, which is not exactly the same as appearing on the RHS of an assignment, but I thought it would be intuitive for people familiar with the term from other language designs. I'm open to other names.

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
aferr: I think we need some name other than predicate to capture the whole syntax of objects that can appear on the RHS. Predicate is just one of those kinds of values and the arithmetic operations are another. Maybe in the future we extend this syntax and there are even more kinds of objects that can appear on the RHS of a conditional. I thought `RValue` was an intuitive way to name this part of the grammar for the reasons described in the other thread.

rust/tools/authorization-logic/src/ast.rs
aferr: fixed

rust/tools/authorization-logic/src/ast.rs
aferr: fixed

aferr:
@bgogul left 2 conversations unresolved

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
bgogul: Sounds good. Don't want to push back too much on this, but I understand what these are. Can you add some brief documentation?

rust/tools/authorization-logic/src/ast.rs
bgogul: I am aware of `rvalue` and  I think it fits the usage here. Can you just document it briefly thouugh?

bgogul (approved):

rust/tools/authorization-logic/src/souffle/datalog_ir.rs
aferr: fixed

PR #430, opened 2022-03-08
Small bug fix for match on sign & lint fixes

Cypher1:
Noticed a set of easy to clean up warnings on a first compile.

One, accidentally binding the variable `Negative` instead of using the `Sign::Negated` case could have caused a bug but happened to correctly match all cases of sign other than `Sign::Positive` so is just a cosmetic change unless other signs are added later.


aferr:
Thanks for the fixes!

bgogul (approved):

PR #438, opened 2022-03-09
Simplify `TagTransform` proto message.

markww:
After a conversation with @bgogul yesterday, I realized I had not split
things correctly across the dataflow graph and policy aspects. This
change simplifies the `TagTransform` structure to tighten its structural
focus and adds a name to allow it to link with the policy rule to which
it connects.

### DIFF
--- a/src/frontends/sql/sql_ir.proto
+++ b/src/frontends/sql/sql_ir.proto
@@ -47,27 +47,6 @@ message MergeOperation {
   repeated Expression control_inputs = 3;
 }
 
-// The kind of a tag.
-enum TagKind {
-  TAG_KIND_UNKNOWN = 0;
-  TAG_KIND_INTEGRITY = 1;
-  TAG_KIND_CONFIDENTIALITY = 2;
-}
-
-// The action that is taken upon a tag by a TagTransform expression.
-enum TagAlteration {
-  TAG_ALTERATION_NONE = 0;
-  TAG_ALTERATION_ADD = 1;
-  TAG_ALTERATION_REMOVE = 2;
-}
-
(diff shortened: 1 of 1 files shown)


src/frontends/sql/sql_ir.proto
```diff
@@ -81,21 +60,14 @@ message TagTransform {
...
-  // integrity tags that are required to be present upon that operation for
-  // this transform to execute. Because it is possible that a `TagTransform`
-  // could require a single expression to have multiple integrity tag values,
-  // we express this `map` as a list of pairs.
-  repeated ExprTagPrecondition tag_preconditions = 2;
-  // The action to be performed on the tag state of the node indicated by the
-  // `transformed_node_id` if all of the required integrity tags are present
-  // on the nodes indicated in the `required_integrity_tags` map. Can be
-  // added or removed.
-  TagAlteration tag_alteration = 3;
```
bgogul, line 94: Don't you need to extract this information from policy rules? Are you planning to do generate Kotlin -> Datalog directly? 

In any case, the simplification in this PR is good.

bgogul (approved):
Nic

src/frontends/sql/sql_ir.proto
markww: Yes, I'm planning on directly printing from Kotlin to Datalog. Initially, I was thinking about having the protos be used to communicate about rule trees and structural trees, and we would handle the policy translation on the Raksha side. However, this is probably more work than was necessary for an implementation we know will be an MVP and it was confusing my thinking about generating dataflow graph Datalog. I think this is a better separation of concerns.

PR #439, opened 2022-03-09
Remove cargo tests from CI as they are covered by bazel tests.

bgogul:
We can now make use of bazel cache to speed rust tests in CI.


markww (approved):
LGTM.

PR #440, opened 2022-03-10
Enable layering_check in all packages.

bgogul:
This PR also fixes some formatting errors (e.g., `load` should be in the beginning) in the BUILD files.

### DIFF
--- a/src/common/proto/BUILD
+++ b/src/common/proto/BUILD
@@ -20,14 +20,16 @@ package(
     licenses = ["notice"],
 )
 
-# These files contain dummy marker libraries for indicating a dependency upon
-# protobuf libraries. We have to use them because Bazel cannot be configured
-# to just force explicit inclusion of protobuf dependencies. This causes
-# problems when we try to import Raksha into Google's internal repository.
-cc_library(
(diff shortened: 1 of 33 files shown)


src/common/proto/BUILD
```diff
@@ -14,19 +14,22 @@
...
-
-# These files contain dummy marker libraries for indicating a dependency upon
-# protobuf libraries. We have to use them because Bazel cannot be configured
-# to just force explicit inclusion of protobuf dependencies. This causes
-# problems when we try to import Raksha into Google's internal repository.
-cc_library(
+# These files contain marker libraries for indicating a dependency upon
+# protobuf libraries. We have to use them because the external protobuf
+# library provides a single BUILD target that encompasses all protobuf
+# related headers. However, the internal targets are much more fine grained.
```
markww, line 26: Should probably be "Google-internal" rather than just "internal", as we expect that our readers in an open source repository may have their own default idea of "internal".

markww (approved):
One small nit on a comment. Otherwise LGTM.

src/common/proto/BUILD
bgogul: Done.

PR #451, opened 2022-03-15
Replace uses of utils::MoveAppend with absl variants.

bgogul:


aferr (approved):
OK good to know that absl has this. I'll also use this one from now on. Thanks.

PR #452, opened 2022-03-15
Fix formatting in src/ir/auth_logic files.

bgogul:


aferr (approved):
LGTM thanks for fixing this.

I have been using `clang-format --style=Google`, and my text editor also runs this when I save. Is there any reason why that would not catch and fix these style issues automatically? (Maybe I just didn't set that up yet when I started writing this).

bgogul:
> LGTM thanks for fixing this.
> 
> I have been using `clang-format --style=Google`, and my text editor also runs this when I save. Is there any reason why that would not catch and fix these style issues automatically? (Maybe I just didn't set that up yet when I started writing this).

These files might have been created before you set it up.

PR #453, opened 2022-03-15
Generalize MapIter to take an iterable container, use std::transform, and reorder template arguments.

bgogul:
This commit has the following changes:
   - Generalizes MapIter a bit to accept any iterable container.
   - Uses std::transform instead of a `for`-loop.
   - Reordering template arguments helps argument deduction at call sites.
   - Updates the test to use appropriate matchers to help debugging test failures.

### DIFF
--- a/src/common/utils/map_iter.h
+++ b/src/common/utils/map_iter.h
@@ -23,13 +23,12 @@
 
 namespace raksha::utils {
 
-template <typename T, typename U, typename F>
-std::vector<U> MapIter(const std::vector<T>& input, F f) {
+template <typename U, typename C, typename F>
+std::vector<U> MapIter(const C& input, F map_fn) {
   std::vector<U> result;
   result.reserve(input.size());
-  for (auto& entity : input) {
-    result.push_back(f(entity));
-  }
+  std::transform(std::begin(input), std::end(input), std::back_inserter(result),
+                 map_fn);
   return result;
 }
 
--- a/src/common/utils/map_iter.h
+++ b/src/common/utils/map_iter.h
@@ -19,10 +19,13 @@
 #ifndef SRC_COMMON_UTILS_MAP_ITER_H_
 #define SRC_COMMON_UTILS_MAP_ITER_H_
 
+#include <algorithm>
 #include <vector>
 
 namespace raksha::utils {
 
+// Returns a vector that is generating by applying `map_fn: (C::value_type ->
+// U)` to every elemnent of the input container `C`.
 template <typename U, typename C, typename F>
 std::vector<U> MapIter(const C& input, F map_fn) {
   std::vector<U> result;
--- a/src/ir/auth_logic/souffle_emitter.h
+++ b/src/ir/auth_logic/souffle_emitter.h
@@ -50,14 +50,14 @@ class SouffleEmitter {
   // from generating two declarations, the sign here is always positive.
   datalog::Predicate PredToDeclaration(const datalog::Predicate& predicate) {
     int i = 0;
-    return datalog::Predicate(predicate.name(),
-                              utils::MapIter<std::string, std::string>(
-                                  predicate.args(),
-                                  [i](const std::string& arg) mutable {
-                                    return absl::StrCat("x",
-                                                        std::to_string(i++));
-                                  }),
-                              datalog::kPositive);
+    return datalog::Predicate(
+        predicate.name(),
+        utils::MapIter<std::string>(predicate.args(),
+                                    [i](const std::string& arg) mutable {
+                                      return absl::StrCat("x",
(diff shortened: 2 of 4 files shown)


src/common/utils/map_iter.h
```diff
@@ -23,13 +23,12 @@
 
 namespace raksha::utils {
 
-template <typename T, typename U, typename F>
-std::vector<U> MapIter(const std::vector<T>& input, F f) {
+template <typename U, typename C, typename F>
+std::vector<U> MapIter(const C& input, F map_fn) {
```
aferr, line 27: It looks like now that the first argument is a reference to C, it works for other kinds of collections of type C? It's a bit non-obvious to me that this works. Is it worth putting a comment that the intention is for the first argument to be some kind of collection?

aferr (approved):

src/common/utils/map_iter.h
bgogul: Good point. Done.

src/common/utils/map_iter.h
```diff
@@ -19,17 +19,19 @@
...
 #include <vector>
 
 namespace raksha::utils {
 
-template <typename T, typename U, typename F>
-std::vector<U> MapIter(const std::vector<T>& input, F f) {
+// Returns a vector that is generating by applying `map_fn: (C::value_type ->
+// U)` to every elemnent of the input container `C`.
+template <typename U, typename C, typename F>
+std::vector<U> MapIter(const C& input, F map_fn) {
```
markww, line 30: I think having `MapIter` accept a `const` reference parameter could be a bit of a mistake. It prevents copy elision from making the result of `MapIter` be constructed directly into the

markww:
Also, can we make the title of the PR a bit more comprehensive? I was looking around for the PR where the `std::transform` refactoring occurred and didn't realize it was in this PR.

src/common/utils/map_iter.h
markww: Note that this shift would also make it preferred to have lambdas accepting an element type `E` to have their parameter type be `E` rather than `const E &`.

bgogul: > I don't think it will. The guidance from the style guide is to, whenever we have a function that is performing a copy of a parameter in its body, take the parameter by value instead of by reference to move the copy to the callsite to open the door for copy elision.

We are not always making a copy of the input container `C`. Rather, we usually transform it to something else. Won't we make unnecessary copies in such cases? 

A couple of other concerns with taking the container `C` by value: 
 - If we really want to move the values, we should be able to use `std::transform` directly to do an in-place update.
 - Moreover, if we really want to take advantage of the `move_iterator`, the `map_fn` also needs to take its argument by value. 
 
IMO, all of this unnecessary complicates the API. Think of this as a `map` in functional programming, where you map one sequence of values to another.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -50,14 +50,14 @@ class SouffleEmitter {
...
-                                    return absl::StrCat("x",
-                                                        std::to_string(i++));
-                                  }),
-                              datalog::kPositive);
+    return datalog::Predicate(
+        predicate.name(),
+        utils::MapIter<std::string>(predicate.args(),
+                                    [i](const std::string& arg) mutable {
+                                      return absl::StrCat("x",
+                                                          std::to_string(i++));
```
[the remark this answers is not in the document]
bgogul: We can do this a separate PR.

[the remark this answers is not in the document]
markww: Sure, sounds good.

src/common/utils/map_iter.h
bgogul: Just for context, I was really going for an incremental change here. Further, after discovering `absl` helpers, I have started thinking if we should even have this API.  That said, I understand the spirit of your concern, but I think it will make the common case more expensive. It is possible that I am misunderstanding something here. Let us sync in person. 

> The body of this map function uses std::transform to read each element from the input and write a corresponding element into the result local, which is then returned. That means that the body of this function makes a copy of its input.

Given that we are using iterators, we won't be making a copy of the input if we use `std::transform`.  Am I wrong? The possible implementations in https://en.cppreference.com/w/cpp/algorithm/transform suggests no copies are made as well.

markww: Gogul and I talked offline. I hadn't taken into account that we are often considering only a few fields from the input vector when performing a `map`, so my argument about the equivalence of copies doesn't hold up as well as I thought.

markww (approved):
Looks good, but I'd still say change the PR title as this covers more than template parameter order.

Issue #364, opened 2022-02-10
Make comparisons in aut_logic::ast.h work by implementing a `compare` method.


aferr:
suggestion via @markww 

Make comparisons in `ast.h` work by implementing a `compare` method. Full context:

> One thing you might want to do: since you have an operator< and an operator==, maybe just make a single strcmp-style compare method that, given another Predicate, returns -1 if the current Predicate is less than the other Predicate, +1 if greater, and 0 if equal. Then you can just make all of the comparison operators use that, which ensures they have the same sense of ordering. 

[DLIR to Souffle Emitter #278](https://github.com/google-research/raksha/pull/278)

PR #456, opened 2022-03-15
Simply datalog::Predicate comparator

bgogul:


markww (approved):

src/ir/datalog/program.h
```diff
@@ -58,7 +58,7 @@ class Predicate {
     if (this->args().size() != otherPredicate.args().size()) {
       return false;
     }
-    for (uint64_t i = 0; i < this->args().size(); i++) {
+    for (uint64_t i = 0; i < this->args().size(); ++i) {
```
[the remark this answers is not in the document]
markww: Oh, I see, it's iterating over two lists together. Sounds good.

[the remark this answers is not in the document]
bgogul: Ah, your comment let me to simplify the comparison operator using `absl::c_equal`. PTAL. (There was also a problem with the BUILD target of the test.)

bgogul:
@harsha-mandadi FYI

src/ir/datalog/program.h
```diff
@@ -49,19 +51,9 @@ class Predicate {
...
-    if (this->args().size() != otherPredicate.args().size()) {
-      return false;
-    }
-    for (uint64_t i = 0; i < this->args().size(); i++) {
-      if (this->args().at(i) != otherPredicate.args().at(i)) return false;
-    }
-    return true;
+    return this->name() == otherPredicate.name() &&
+           this->sign() == otherPredicate.sign() &&
+           absl::c_equal(this->args(), otherPredicate.args());
```
markww, line 56: While I think this is a good way to simplify `operator==`, it doesn't make progress towards an even bigger refactoring that I think would be valuable. That is, the `operator==` and `operator<` have separate implementations, but it would be nice to just define the order on predicates with a `strcmp`-style `compare` function and have `operator==`'s body be just `compare(other) == 0` and `operator<`'s body be just `compare(other) < 0`. Should we just create the `compare` function?

markww: Relevant issue: https://github.com/google-research/raksha/issues/364

harsha-mandadi: @markww Just curious..Does compare(other) compare between 2 predicates? I can imagine a scenario where we check if 2 predicates are same. What is the use case for operator< ??

markww: See here: https://github.com/google-research/raksha/blob/main/src/ir/auth_logic/souffle_emitter.h#L116 . That said, maybe the `Predicate`s should be stored in a `vector` rather than a `set`, which would make the canonical order just "the order in which the `Predicate`s were encountered", which would make `<` unnecessary.

bgogul: I think there is some implicit de-duping based on `name`. In any case that use needs to be fixed and we should get rid of `operator <`. I will add the discussion here to #364 and resolving this comment.

markww: Wait, before you do, are we sure that `this->args() == otherPredicate.args()` wouldn't work?

markww: Yes, actually: `args` is a `std::vector` which, if its value type has an `operator==`, also has an `operator==`. So we should probably just use `std::vector` `operator==` instead of `c_equal`.

PR #459, opened 2022-03-16
Add structopt as Rust dependency

aferr:
This PR is the first PR that will split
[#245](https://github.com/google-research/raksha/pull/245) into two.
This one just adds structop as a dependency to cargo/BUILD.bazel and
the result of running `cargo.raze`. The next PR will make use of
structopt.

Co-authored-by: bMacSwigg <mcswiggen@google.com>


aferr:
@bgogul any idea why github actions are not triggering on this?

aferr:
I don't know why new commits are not pushing to this. I'll just close this one and we can use the one that mirrors this.

PR #466, opened 2022-03-18
Add path information to Raksha errors

Cypher1:
First step in providing more diagnostic information in errors in the souffle interface.


bgogul (approved):

Cypher1:
@bgogul " Import didn't affect any internal file" so I think this can be merged safely. (Sorry I did try to use the "readytopull" label but I don't think it works in this case).

bgogul:
Thanks @Cypher1 I have merged it now.

PR #467, opened 2022-03-18
Handle importing programs from multiple files

aferr:


rust/tools/authorization-logic/src/compilation_top_level.rs
```diff
@@ -33,6 +33,41 @@ use crate::{
...
+            .map(|prog| prog.priv_binds.clone())
+            .flatten()
+            .collect(),
+        pub_binds: multiprog
+            .iter()
+            .map(|prog| prog.pub_binds.clone())
+            .flatten()
+            .collect(),
+    }
+}
```
markww, line 69: This looks like it could be more naturally expressed as a `fold` upon the `multiprog` vector, where the `fold` lambda just merges the contents of each subcollection of two `AstProgram`s, producing one resultant `AstProgram`.

rust/tools/authorization-logic/src/compilation_top_level.rs
```diff
@@ -46,11 +81,21 @@ pub fn source_file_to_ast_test_only(input_file_path: &str) -> AstProgram {
...
+    let resolved_in_files = input_file_paths.into_iter()
+        .map(|f| utils::get_resolved_path(f));
     let resolved_out_file = utils::get_resolved_path(&output_file_path);
-    let prog = source_file_to_ast(&resolved_in_file);
-    let prog_with_imports = import_assertions::handle_imports(&prog);
+    let progs = resolved_in_files
+        .map(|f| source_file_to_ast(&f))
+        .collect();
+    let merged_prog = merge_multiprogram(progs);
+    let prog_with_imports = import_assertions::handle_imports(&merged_prog);
```
markww, line 98: This might factor the complexity better to consider compiling a multiprogram to be a composition of combining the programs and compiling that one program. That would leave `compile` largely unchanged, but would add a `compile_multiprogram` function as an optional previous step.

rust/tools/authorization-logic/src/lib.rs
```diff
@@ -41,7 +41,12 @@ pub extern "C" fn generate_datalog_facts_from_authorization_logic(
     let decl_skip_vec = unsafe {
         CStr::from_ptr(c_decl_skip)
     }.to_str().unwrap().split(',').map(|s| s.to_string()).collect();
-    compilation_top_level::compile(input_filename,
+    // The CLI interface has been extended to take multiple files as inputs, but the C library does 
+    // not make use of this because it is not needed, yet. Doing so would just entail changing this 
+    // interface and the code that consumes it; the internals of the auth logic compiler should not 
+    // need to change -- the following line can just call compile(...)
+    // instead of compile_one_prgram(...)
+    compilation_top_level::compile_one_program(input_filename,
```
markww, line 49: Is this information even important? Is, from the C program's perspective, compiling a multiprogram any different from preprocessing a multiprogram to be a single program?

Also, line wrapping.

markww (commented):
It looks like this PR is trying to merge several files that each contain authorization logic language portions into a single file and compiling it as if it were a single file. That will probably work in the short term, but it does raise longer-term questions. For instance, if there is a syntax error in one of the files to be combined, how to we report that error in a fashion that a user can easily find and fix?

In C/C++, which also combines source files, preprocessor directives such as `#file` and `#line` allow source location in the original files to be preserved. However, a better approach would probably be to separately compile each of the input files and combine them in their compiled form. That, I think, would currently be as simple as concatenating the resultant Datalog together.

aferr:
> It looks like this PR is trying to merge several files that each contain authorization logic language portions into a single file and compiling it as if it were a single file. That will probably work in the short term, but it does raise longer-term questions. For instance, if there is a syntax error in one of the files to be combined, how to we report that error in a fashion that a user can easily find and fix?
> 
> In C/C++, which also combines source files, preprocessor directives such as `#file` and `#line` allow source location in the original files to be preserved. However, a better approach would probably be to separately compile each of the input files and combine them in their compiled form. That, I think, would currently be as simple as concatenating the resultant Datalog together.

These are good points, and we should consider them for the long-term. At the moment, this implementation does not do much checking or error reporting on the source files to begin with, even without merging multiple files.

As you know, this rust implementation will not live long, and at the moment this is meant to target a specific feature set to reach a milestone.

Perhaps you should open an issue so that we address this when the time is right for the "real" C++ implementation?

rust/tools/authorization-logic/src/compilation_top_level.rs
aferr: Defining this with `fold` would require first defining an `AstProgram -> AstProgram -> AstProgram` function, and that function would look pretty similar to this one, so I think it is better to leave this as is.

rust/tools/authorization-logic/src/lib.rs
aferr: I left this comment because it could be useful to have this library handle combining multiple imported files, (but didn't finish implementing that because it is not strictly needed yet). I don't think including this comment hurts anything.

Actually [100 characters is the correct line wrapping](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md#indentation-and-line-width) even though the rest of this code tends to use 80 because I prefer 80.

rust/tools/authorization-logic/src/compilation_top_level.rs
aferr: I think this suggestion is about the same as what's here except the names of functions, and the code that previously made use of `compile` has already been changed to use `compile_one_program`, so I'd rather leave it as-is.

rust/tools/authorization-logic/src/lib.rs
aferr: OK. fixed to make it 80. I guess it's better to at least be consistent.

rust/tools/authorization-logic/src/compilation_top_level.rs
markww: You could do it entirely in a lambda in the fold call. It would look something like this:

```Rust
multiprog.iter().fold(None, |acc, currentAstProg|
  Some(match acc {
    None => currentAstProg
    Some(combinedAstProg) => AstProgram {
      acc.relation_declarations.append(currentAstProg.relation_declarations);
      acc.assertions.append(currentAstProg.assertions);
      acc.queries.append(currentAstProg.queries);
      acc.imports.append(currentAstProg.imports);
      acc.priv_binds.append(currentAstProg.priv_binds);
      acc.pub_binds.append(currentAstProg.pub_binds);
      acc
    })
  }).collect().unwrap(); // At least one program required.
```

The benefit of the above is that you don't have to perform the collection iteration stuff for each array, which makes it more readable.

markww:
I think that since this is throwaway code, using this approach and creating an issue to fix it in the future is fine. You are the one writing the code making this tradeoff, however, and this should be annotated in the code with a `TODO`, so I'll leave it to you to create the issue and add the comment.

rust/tools/authorization-logic/src/compilation_top_level.rs
aferr: fixed

aferr:
> I think that since this is throwaway code, using this approach and creating an issue to fix it in the future is fine. You are the one writing the code making this tradeoff, however, and this should be annotated in the code with a `TODO`, so I'll leave it to you to create the issue and add the comment.

OK here is [an issue](https://github.com/google-research/raksha/issues/485). I don' think adding a TODO to the code makes sense because there actually are no checks on the source files to begin with so importing from multiple files does not make a difference yet. I also think there are multiple ways to address this problem, and changing the order of compilation steps may not be the best solution (and we do not need to figure out how to solce this now).

markww (approved):
LGTM

PR #469, opened 2022-03-18
Move AuthLogic.g4 to src/frontends/

harsha-mandadi:
Grammar file consumed by both antlr with C++ and Rust backend points to a location in //rust direcyory. We intended to have a common grammar file inside raksha and all its reference would point to this grammar file. But, when importing to google3, this might be a problem as we are hiding all of auth_logic under rust/*. 

This PR takes care of moving AuthLogic.g4 from rust/tools/authorization-logic/src/parsing/ to src/frontends/auth_logic


rust/tools/authorization-logic/BUILD
```diff
@@ -118,6 +118,7 @@ rust_library(
         "//rust/tools/authorization-logic/cargo:antlr_rust",
         "//rust/tools/authorization-logic/cargo:bincode",
         "//rust/tools/authorization-logic/cargo:serde",
+        "//rust/tools/authorization-logic/cargo:structopt",
```
markww: You should probably make the diffbase of this PR the PR that introduces `structopt` so reviewers can focus on your new changes. It also adds a barrier to help prevent submitting the contents of the dependent PR before that PR.

src/frontends/auth_logic/BUILD
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#-------------------------------------------------------------------------------
+package(
+    default_visibility = ["//visibility:public"],
```
markww, line 18: Generally, truly-public visibility is discouraged. Instead, I'd say take out the default visibility and manually expose the file to the one package that needs it using `export_files`.

src/frontends/auth_logic/BUILD
```diff
@@ -0,0 +1,25 @@
...
+    licenses = ["notice"],
+)
+
+filegroup(
+    name = "authlogic_grammar",
+    srcs = ["//src/frontends/auth_logic:AuthLogic.g4"],
+)
```
markww, line 25: It's a little non-idiomatic to use a filegroup to expose a single file. You might want to just use `exports_files` to export this single file. You can use `exports_files` to set the visibility on that file and refer to that file's dependency path directly from the package(s) to which it is exposed.

rust/tools/authorization-logic/BUILD
markww: Looks like you fixed this.

markww (approved):
Looks good other than one possibly-outdated comment.

src/frontends/auth_logic/BUILD
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#-------------------------------------------------------------------------------
+package(
+    licenses = ["notice"],
+)
+
+#from //src/frontends/auth_logic
```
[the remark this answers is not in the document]
harsha-mandadi: Removed!

rust/tools/authorization-logic/BUILD
```diff
@@ -34,7 +34,7 @@ java_binary(
 # Generate the parsing-related files using antlr4rust
 genrule(
     name = "antlr_gen_srcs",
-    srcs = ["src/parsing/AuthLogic.g4"],
+    srcs = ["//src/frontends/auth_logic:AuthLogic.g4"],
```
bgogul, line 37: Let us put this under `src/policy/auth_logic` unless there is strong objection from others.

bgogul (approved):

PR #474, opened 2022-03-21
Introduce an sql::OpTraits class to associate information with SqlOps

bgogul:
`OpTraits<T>` will be needed to support classes that provide `operator`-specific view of the underlying `ir::Operation`.

### DIFF
--- a/src/frontends/sql/ops/op_traits.h
+++ b/src/frontends/sql/ops/op_traits.h
@@ -28,7 +28,7 @@ struct OpTraits {};
   template <>                                           \
   struct OpTraits<CLASS> {                              \
     static constexpr absl::string_view kName = OP_NAME; \
-  };
+  }
 
 DEFINE_OP_TRAITS(LiteralOp, "sql.literal");
 
--- a/src/frontends/sql/ops/op_traits.h
+++ b/src/frontends/sql/ops/op_traits.h
@@ -1,0 +1,39 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/frontends/sql/ops/op_traits.h
```diff
@@ -0,0 +1,39 @@
...
+namespace raksha::frontends::sql {
+
+template <typename Op>
+struct OpTraits {};
+
+#define DEFINE_OP_TRAITS(CLASS, OP_NAME)                \
+  class CLASS;                                          \
+  template <>                                           \
+  struct OpTraits<CLASS> {                              \
+    static constexpr absl::string_view kName = OP_NAME; \
```
markww, line 30: I think the macro version is fine, I've seen this idiom before. But macros are recommended against by the style guide, so just want to challenge it a little bit.

What if `OpTraits` were just a regular non-templatized `struct` and every class for which we want `OpTraits` had a `static OpTraits GetOpTraits()` method upon it that just returned a designated-initializer-created `OpTraits` struct? That would also work about the same and wouldn't require a macro, right?

bgogul: Actually, I started with adding static methods to the classes. However, when I wanted to ensure that we don't have duplicate operator names, it was not easy to enforce. So, I just moved to a centralized file where any duplicates will become obvious. 

As an additional benefit, it also provides a single place where all operators for a particular frontend is defined. We could use this to define a helper to register all the operators into an IRContext.

src/frontends/sql/ops/op_traits.h
```diff
@@ -0,0 +1,39 @@
...
+  };
```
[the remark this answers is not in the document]
bgogul: Will do.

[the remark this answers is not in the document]
bgogul: DOne.

bgogul:
@markww please take a look at the discussion around macros.

src/frontends/sql/ops/op_traits.h
markww: I see, ok. You could make the name collision guard even stronger with some macro token-pasting magic. For instance, imagine that `OP_NAME` was passed without quotes and the macro looked like this:

```C++
#define DEFINE_OP_TRAITS(CLASS, OP_NAME)                \
  bool op_name_unique_guard ## OP_NAME ##__ = false; \
  class CLASS;                                          \
  template <>                                           \
  struct OpTraits<CLASS> {                              \
    static constexpr absl::string_view kName = #OP_NAME; \
  }
```

That would make the invocation of `DEFINE_OP_TRAITS` look like this:

```C++
DEFINE_OP_TRAITS(LiteralOp, sql.literal);
```

Totally optional though.

bgogul: That is clever! Let me skip this for now as it seems compiler dependent? I don't quite like the lack of `"` around `sql.literal` either (not a big deal).

PR #476, opened 2022-03-21
Add a sql::LiteralOp class 

bgogul:


src/frontends/sql/ops/literal_op.h
```diff
@@ -0,0 +1,45 @@
...
+ public:
+  static constexpr absl::string_view kLiteralStringAttrName = "literal_string";
+
+  // Constructs a LiteralOp.
+  static std::unique_ptr<LiteralOp> Create(const ir::Block* parent_block,
+                                           const ir::IRContext& context,
+                                           absl::string_view literal);
+
+  // Inherit constructors.
+  using SqlOp::SqlOp;
```
markww, line 37: You're expecting to only create a `LiteralOp` through the `Create` method right? Doesn't that imply that the parent ctors should be provided in a non-`public` fashion? Otherwise we could use one of the less-restricted superclass ctors to create something of type `LiteralOp` that doesn't have the guarantees that the `Create` method provides by construction.

bgogul: Good catch. I meant to tighten it before sending out the PR, but forgot! Made this a private constructor as we need it in `LiteralOp::Create` method.

markww (approved):

PR #483, opened 2022-03-22
Remove LiteralOperationView class from sql/testing

bgogul:


markww (approved):
Looks good in isolation, but I wonder whether the constructor comment I left on the PR introducing the `LiteralOp` complicates the picture of this downstream PR, particularly the calls to `AddOperation`.

bgogul:
> Looks good in isolation, but I wonder whether the constructor comment I left on the PR introducing the `LiteralOp` complicates the picture of this downstream PR, particularly the calls to `AddOperation`.

That was an oversight. I have addressed the concern in  #476.

PR #487, opened 2022-03-24
Parser for authorization logic in C++

harsha-mandadi:
This PR takes care of following:
1. Generates parse trees from an input stream using antlr generated Parser.
2. Construct AST from the parse tree

### DIFF
--- a/src/ir/auth_logic/BUILD
+++ b/src/ir/auth_logic/BUILD
@@ -104,7 +104,7 @@ cc_library(
         "//src/common/utils:overloaded",
         "//src/ir/auth_logic:ast",
         "//src/ir/datalog:program",
-        "//src/policy/auth_logic:auth_logic_cc_generator_cc_parser",
+        "//src/policy/auth_logic:auth_logic_cc_generator",
     ],
 )
 
--- a/src/ir/auth_logic/BUILD
+++ b/src/ir/auth_logic/BUILD
@@ -91,3 +91,32 @@ cc_test(
         "//src/common/testing:gtest",
     ],
 )
+
+cc_library(
+    name = "ast_construction",
+    srcs = ["ast_construction.cc"],
+    hdrs = ["ast_construction.h"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],
+    deps = [
+        "//src/common/logging",
+        "//src/common/utils:map_iter",
+        "//src/common/utils:overloaded",
+        "//src/ir/auth_logic:ast",
+        "//src/ir/datalog:program",
+        "//src/policy/auth_logic:auth_logic_cc_generator_cc_parser",
+    ],
+)
+
+cc_test(
+    name = "ast_construction_test",
+    srcs = ["ast_construction_test.cc"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],
+    deps = [
+        ":ast",
+        ":ast_construction",
+        "//src/common/testing:gtest",
+        "//src/ir/datalog:program",
+    ],
+)
--- a/src/ir/auth_logic/ast_construction.cc
+++ b/src/ir/auth_logic/ast_construction.cc
@@ -1,0 +1,123 @@
+//-------------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//-------------------------------------------------------------------------------
+
+#include "src/ir/auth_logic/ast_construction.h"
+
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "src/common/logging/logging.h"
+#include "src/common/utils/map_iter.h"
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/datalog/program.h"
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicLexer.h"
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicParser.h"
+
+
+namespace raksha::ir::auth_logic {
+
+static Principal ConstructPrincipal(
+    AuthLogicParser::PrincipalContext& context) {
+  return Principal(context.ID()->getText());
+}
+
+static datalog::Predicate ConstructPredicate(
+    AuthLogicParser::PredicateContext& context) {
+  std::vector<std::string> args = utils::MapIter<std::string>(
+      context.pred_arg(),
+      [](auto* pred_arg) { return pred_arg->ID()->getText(); });
+  datalog::Sign sign = context.NEG() == nullptr ? datalog::Sign::kPositive
+                                                : datalog::Sign::kNegated;
+  return datalog::Predicate(context.ID()->getText(), args, sign);
+}
+
+static BaseFact ConstructVerbphrase(
+    Principal left_principal, AuthLogicParser::VerbphraseContext& context) {
+  if (auto* ctx = dynamic_cast<AuthLogicParser::PredphraseContext*>(
+          &context)) {  // PredicateContext
+    return BaseFact(Attribute(
+        left_principal, ConstructPredicate(*CHECK_NOTNULL(ctx->predicate()))));
+  }
+  auto& ctx =
+      *CHECK_NOTNULL(dynamic_cast<AuthLogicParser::ActsAsPhraseContext*>(
+          &context));  // ActsAsPhraseContext
+  return BaseFact(CanActAs(
+      left_principal, ConstructPrincipal(*CHECK_NOTNULL(ctx.principal()))));
+}
+
+static BaseFact ConstructFlatFact(AuthLogicParser::FlatFactContext& context) {
+  if (auto* ctx = dynamic_cast<AuthLogicParser::PrinFactContext*>(&context)) {
+    return ConstructVerbphrase(
+        ConstructPrincipal(*CHECK_NOTNULL(ctx->principal())),
+        *CHECK_NOTNULL(ctx->verbphrase()));
+  }
+  auto& ctx = *CHECK_NOTNULL(dynamic_cast<AuthLogicParser::PredFactContext*>(
+      &context));  // predFactContext
+  return BaseFact(ConstructPredicate(*CHECK_NOTNULL(ctx.predicate())));
+}
+
+static Fact ConstructFact(AuthLogicParser::FactContext& context) {
+  if (auto* flat_fact_fact_ctx =
+          dynamic_cast<AuthLogicParser::FlatFactFactContext*>(
+              &context)) {  // FlatFactFactContext
+    return Fact(
+        ConstructFlatFact(*CHECK_NOTNULL(flat_fact_fact_ctx->flatFact())));
+  }
+  auto& ctx = *CHECK_NOTNULL(
+      dynamic_cast<AuthLogicParser::CanSayFactContext*>(&context));
+  return ConstructFact(*CHECK_NOTNULL(ctx.fact()));
+}
+
+static Query ConstructQuery(AuthLogicParser::QueryContext& context) {
+  return Query(context.ID()->getText(),
+               ConstructPrincipal(*CHECK_NOTNULL(context.principal())),
+               ConstructFact(*CHECK_NOTNULL(context.fact())));
+}
+
+static Program ConstructProgram(AuthLogicParser::ProgramContext& context) {
+  std::vector<SaysAssertion> says_assertions;
+  std::vector<Query> queries = utils::MapIter<Query>(
+      context.query(), [](AuthLogicParser::QueryContext* query_context) {
+        return ConstructQuery(*CHECK_NOTNULL(query_context));
+      });
+  return Program(std::move(says_assertions), std::move(queries));
+}
+
+/// This function produces an abstract syntax tree (AST) rooted with a program
+/// node when given the textual representation of a program.
+Program ParseProgram(absl::string_view prog_text) {
+  // Provide the input text in a stream
+  antlr4::ANTLRInputStream input(prog_text);
+  // Creates a lexer from input
+  AuthLogicLexer lexer(&input);
+  // Creates a token stream from lexer
+  antlr4::CommonTokenStream tokens(&lexer);
+  // Creates a parser from the token stream
+  AuthLogicParser parser(&tokens);
+  // program_context points to the root of the parse tree
+  AuthLogicParser::ProgramContext& program_context =
+      *CHECK_NOTNULL(parser.program());
+  return ConstructProgram(program_context);
+}
+
+}  // namespace raksha::ir::auth_logic
--- a/src/ir/auth_logic/ast_construction.h
+++ b/src/ir/auth_logic/ast_construction.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2022 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// This file contains the datalog IR (DLIR) which makes the translation from
+// this authorization logic into datalog simpler.
+
+#ifndef SRC_IR_AUTH_LOGIC_AST_CONSTRUCTION_H_
+#define SRC_IR_AUTH_LOGIC_AST_CONSTRUCTION_H_
+
+#include "src/ir/auth_logic/ast.h"
+
+namespace raksha::ir::auth_logic {
+
+// Parses an input string
+Program ParseProgram(absl::string_view prog_text);
+
+}  // namespace raksha::ir::auth_logic
+
+#endif
--- a/src/ir/auth_logic/ast_construction_test.cc
+++ b/src/ir/auth_logic/ast_construction_test.cc
@@ -1,0 +1,76 @@
+//-------------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//-------------------------------------------------------------------------------
+
+#include "src/ir/auth_logic/ast_construction.h"
+
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "src/common/testing/gtest.h"
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
+
+struct QueryTestData {
+  std::string input_text;
+  std::string query_name;
+  std::string principal_name;
+  std::string fact_name;
+  datalog::Predicate predicate;
+};
+
+class AstConstructionTest : public testing::TestWithParam<QueryTestData> {};
+
+TEST_P(AstConstructionTest, SimpleTestWithQueryProgram) {
+  const auto& [input_text, query_name, principal_name, fact_name, predicate] =
+      GetParam();
+  Program prog_out = ParseProgram(input_text);
+  ASSERT_EQ(prog_out.queries().size(), 1);
+  const Query& query = prog_out.queries().front();
+  EXPECT_EQ(query.name(), query_name);
+  EXPECT_EQ(query.principal().name(), principal_name);
+  EXPECT_EQ(std::get<datalog::Predicate>(
+                std::get<BaseFact>(query.fact().GetValue()).GetValue())
+                .name(),
+            fact_name);
+  EXPECT_EQ(std::get<datalog::Predicate>(
+                std::get<BaseFact>(query.fact().GetValue()).GetValue()),
+            predicate);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    AstConstructionTest, AstConstructionTest,
+    testing::Values(
+        QueryTestData(
+            {R"(may_notifierA = query "UserA" says mayA("NotifierA"))",
+             "may_notifierA", R"("UserA")", "mayA",
+             datalog::Predicate("mayA", {R"("NotifierA")"},
+                                datalog::Sign::kPositive)}),
+        QueryTestData(
+            {R"(may_notifierB = query "UserB" says !mayB("NotifierB"))",
+             "may_notifierB", R"("UserB")", "mayB",
+             datalog::Predicate("mayB", {R"("NotifierB")"},
+                                datalog::Sign::kNegated)})));
+
+}  // namespace raksha::ir::auth_logic
--- a/src/policy/auth_logic/BUILD
+++ b/src/policy/auth_logic/BUILD
@@ -17,15 +17,15 @@
 load("//build_defs:antlr.bzl", "antlr4_cc_combined")
 
 package(
+    default_visibility = [
+        "//rust/tools/authorization-logic:__pkg__",
+        "//src/ir/auth_logic:__pkg__",
+    ],
     licenses = ["notice"],
 )
 
 exports_files(
     ["AuthLogic.g4"],
-    visibility = [
-        "//rust/tools/authorization-logic:__pkg__",
-        "//src/ir/auth_logic:__pkg__",
-    ],
 )
 
 #antlr
--- a/build_defs/antlr.bzl
+++ b/build_defs/antlr.bzl
@@ -39,7 +39,7 @@ def antlr4_cc_combined(name, src, listener = False, visitor = False):
     )
 
     native.cc_library(
-        name = name + "_cc_parser",
+        name = name,
         srcs = [generated],
         hdrs = [generated],
         copts = ["-fexceptions"],
--- a/build_defs/antlr.bzl
+++ b/build_defs/antlr.bzl
@@ -19,14 +19,14 @@ Generates C++ parser, visitor, listener, lexer from a grammar file.
 
 load("@rules_antlr//antlr:antlr4.bzl", "antlr")
 
-def antlr4_cc_combined(name, src, listener = False, visitor = True):
+def antlr4_cc_combined(name, src, listener = False, visitor = False):
     """Creates a C++ lexer, visitor, listener, parser from a source grammar.
     Args:
       name: Base name for the lexer and the parser rules.
       src: source ANTLR grammar file
       package: The namespace for the generated code
-      listener: generate ANTLR listener (default: True)
-      visitor: generate ANTLR visitor (default: True)
+      listener: generate ANTLR listener (default: False)
+      visitor: generate ANTLR visitor (default: False)
     """
     generated = name + "_grammar"
     antlr(
@@ -41,6 +41,7 @@ def antlr4_cc_combined(name, src, listener = False, visitor = True):
     native.cc_library(
         name = name + "_cc_parser",
         srcs = [generated],
+        hdrs = [generated],
         copts = ["-fexceptions"],
         linkopts = ["-fexceptions"],
         features = ["-use_header_modules"],


src/ir/auth_logic/ast_construction.h
```diff
@@ -0,0 +1,59 @@
...
+#define SRC_IR_AUTH_LOGIC_AST_CONSTRUCTION_H_
+
+#include "src/ir/datalog/program.h"
+#include "src/ir/auth_logic/ast.h"
+#include "absl/status/statusor.h"
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicParser.h"
+
+namespace raksha::ir::auth_logic {
+
+class parser {
```
bgogul, line 31: Naming style is wrong for classes and functions: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

bgogul (commented):
Seems like a reasonable apporach. Haven't looked too closely, but naming is off.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,173 @@
...
+     return dynamic_cast<T*>(context);
+}
+
+absl::StatusOr<Fact> construct_fact(AuthLogicParser::FactContext* context) {
+    if (auto* flat_fact_fact_ctx = instanceofFactContext<AuthLogicParser::FlatFactFactContext>(context)) {  
+        absl::StatusOr<BaseFact> result = construct_flatFact(flat_fact_fact_ctx->flatFact());
+        if (result.ok()) {
+            BaseFact basefact = *result;                               //FlatFactFactContext
+            return Fact(basefact);
+        }
```
bgogul, line 114: The following might be a better pattern?
```suggestion
        if (!result.ok())  return result; // Break on error.
        BaseFact basefact = *result;                               //FlatFactFactContext
        return Fact(basefact);
```

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,173 @@
...
+        args_.push_back(std::move(statement->ID()->getText()));
+        index += 1;
+    }
+
+    if (context->NEG() == nullptr) {
+        sign_ = datalog::Sign::kPositive;
+    }
+    else {
+        sign_ = datalog::Sign::kNegated;
+    }
```
bgogul, line 60: It looks like you could declare and initialize the value if you user ternary operator:

```C++
auto sign = context->NEG() == nullptr ? datalog::Sign::kPositive: datalog::Sign::kNegative;
```

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,173 @@
...
+    return datalog::Predicate(name_, args_, sign_);
+
+}
+
+template<typename T, typename = std::enable_if_t<std::is_base_of<AuthLogicParser::VerbphraseContext, T>::value>>
+T* instanceofVerbPhrase(AuthLogicParser::VerbphraseContext* context) {
```
bgogul, line 66: `GetVerbPhraseAs` might be a better name. The usage will be clearer:
```C++
auto* ctx = GetVerbPhraseAs<AuthLogicParser::PredphraseContext>(context)
```

bgogul (commented):
Mostly some style suggestions.

src/ir/auth_logic/BUILD
```diff
@@ -91,3 +91,19 @@ cc_test(
         "//src/common/testing:gtest",
     ],
 )
+
+cc_library(
+    name = "ast_construction",
+    srcs = ["ast_construction.cc"],
+    hdrs = ["ast_construction.h"],
+    copts = ["-fexceptions"],
```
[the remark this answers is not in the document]
harsha-mandadi: Yes. Removing this would cause build to fail with folllowing error.
 (error: exception handling disabled, use '-fexceptions' to enable
   70 |     throw "there is no serialized ATN";)

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicLexer.h" 
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicParser.h"
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicVisitor.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
+
+Principal Parser::ConstructPrincipal (AuthLogicParser::PrincipalContext* context) {
+    std::string name = context->ID()->getText();
+    return Principal(name);
```
markww, line 42: This final use of name should also use `std::move`, as `Principal` takes its `name` parameter by value and `name` is not used after this point in the function.

Alternatively, you could just inline the initializer of `name` into this reference like so:

```C++
return Principal(context->ID()->getText());
```

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+}
+
+datalog::Predicate Parser::ConstructPredicate (AuthLogicParser::PredicateContext* context ) {
```
markww, line 45: See signature suggestions on `constructPrincipal`.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+    std::string name = context->ID()->getText();
+    std::vector<std::string> args;
+    datalog::Sign sign;
```
markww, line 48: In general, C++ locals should be declared as close to their use as possible. This is doubly true if this allows you to not declare an uninitialized variable (as in this case). If you move this past the loop, you can declare `sign` and initialize it all in one line:

```C++
datalog::Sign sign = context->NEG() == nullptr ? datalog::Sign::kPositive : datalog::Sign::kNegated;
```

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+    
+    for (auto& statement : context->pred_arg()) {
+        args.push_back(std::move(statement->ID()->getText()));
+    }
```
markww, line 52: Consider using `utils::MapIter` for this case, might make it a little more concise.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+
+namespace raksha::ir::auth_logic {
+
+Principal Parser::ConstructPrincipal (AuthLogicParser::PrincipalContext* context) {
+    std::string name = context->ID()->getText();
+    return Principal(name);
+}
+
+datalog::Predicate Parser::ConstructPredicate (AuthLogicParser::PredicateContext* context ) {
+    std::string name = context->ID()->getText();
```
markww, line 46: Move the declaration of this local closer to its use at the end of the file (or inline the initializer into the one use and eliminate it altogether).

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+    std::vector<std::string> args;
+    datalog::Sign sign;
+    
+    for (auto& statement : context->pred_arg()) {
+        args.push_back(std::move(statement->ID()->getText()));
+    }
+
+    sign = context->NEG() == nullptr ? datalog::Sign::kPositive : datalog::Sign::kNegated;
+    return datalog::Predicate(name, args, sign);
```
markww, line 55: `name` and `args` should be `std::move`'d into the `Predicate` constructor to prevent them from being copied.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+datalog::Predicate Parser::ConstructPredicate (AuthLogicParser::PredicateContext* context ) {
+    std::string name = context->ID()->getText();
+    std::vector<std::string> args;
+    datalog::Sign sign;
+    
+    for (auto& statement : context->pred_arg()) {
+        args.push_back(std::move(statement->ID()->getText()));
+    }
+
+    sign = context->NEG() == nullptr ? datalog::Sign::kPositive : datalog::Sign::kNegated;
```
markww, line 54: Is it the case that `NEG()` will only return `nullptr` or something else? Can it be explicitly specified as positive?

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+    return datalog::Predicate(name, args, sign);
+
+}
+
+template<typename T, typename = std::enable_if_t<std::is_base_of<AuthLogicParser::VerbphraseContext, T>::value>>
+T* GetVerbPhraseContextAs(AuthLogicParser::VerbphraseContext* context) {
+     return dynamic_cast<T*>(context);
+}
```
markww, line 62: What does this method buy you? Why not just use `dynamic_cast` directly?

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+
+absl::StatusOr<BaseFact> Parser::ConstructVerbphrase(Principal left_principal, AuthLogicParser::VerbphraseContext* context) {
+    if (auto* ctx = GetVerbPhraseContextAs<AuthLogicParser::PredphraseContext>(context)) {
```
markww, line 65: Should probably be `const auto * ctx` unless the `context` is modified by the code below. Similar for other `else if` branch.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+
+        datalog::Predicate predicate = Parser::ConstructPredicate(ctx->predicate());   //PredicateContext
+        return BaseFact(Attribute(left_principal, datalog::Predicate(predicate)));
```
markww, line 68: `predicate` is already of type `Predicate`, why put this constructor call here? Also, you should probably use `std::move` on `predicate` or inline the initializer.

src/ir/auth_logic/ast_construction.h
```diff
@@ -0,0 +1,60 @@
...
+
+absl::StatusOr<Program> ConstructProgram (AuthLogicParser::ProgramContext* context);
+
+absl::StatusOr<Query> ConstructQuery (AuthLogicParser::QueryContext* context);
+
+absl::StatusOr<Fact> ConstructFact(AuthLogicParser::FactContext* context);
+
+absl::StatusOr<BaseFact> ConstructFlatFact(AuthLogicParser::FlatFactContext* context);
+
+absl::StatusOr<BaseFact> ConstructVerbphrase(Principal left_principal, AuthLogicParser::VerbphraseContext* context);
```
markww, line 48: If there is no conceivable way to recover from the errors that would be indicated by the `StatusOr` or if the errors should never come up, you can make these just return the wrapped types and instead use `CHECK` to fail if the checked state doesn't hold. In most of these cases, I think you're just returning `IllegalArgument` if your `dynamic_cast` doesn't work, so I think you can do this for all of these methods.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+    //Creates a parser from the token stream
+    AuthLogicParser parser (&tokens);
+    //program_context points to the root of the parse tree
+    AuthLogicParser::ProgramContext* program_context = parser.program(); 
+    
+    //constructs AST 
+    return Parser::ConstructProgram(program_context);
+}
+
+} //namespace raksha::ir::auth_logic
```
markww, line 164: Newline at end of file, configure your editor to always do this.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+
+        datalog::Predicate predicate = Parser::ConstructPredicate(ctx->predicate());   //PredicateContext
+        return BaseFact(Attribute(left_principal, datalog::Predicate(predicate)));
+    }
+    else if (auto* ctx = GetVerbPhraseContextAs<AuthLogicParser::ActsAsPhraseContext>(context)) {
+        Principal right_principal = Parser::ConstructPrincipal(ctx->principal());    //ActsAsPhraseContext
+        return BaseFact(CanActAs(left_principal, right_principal));
+    }
+
+    return absl::InvalidArgumentError("Invalid context to ConstructVerbphrase");
```
markww, line 75: This should never occur, so it is ok to use a `CHECK` to indicate it rather than using a `StatusOr`.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+    if (auto* ctx = GetFlatFactContextAs<AuthLogicParser::PrinFactContext>(context)) {                                              
+        Principal principal = Parser::ConstructPrincipal(ctx->principal());
+        return Parser::ConstructVerbphrase(principal, ctx->verbphrase());
+    }
+    else if (auto* ctx = GetFlatFactContextAs<AuthLogicParser::PredFactContext>(context)) {
+        datalog::Predicate predicate = Parser::ConstructPredicate(ctx->predicate());   //predFactContext
+        return BaseFact(predicate);
+    }
+
+    return absl::InvalidArgumentError("Invalid context to ConstructFlatFact");
```
markww, line 94: Use a `CHECK`.

src/ir/auth_logic/ast_construction.h
```diff
@@ -0,0 +1,60 @@
...
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicParser.h"
+
+namespace raksha::ir::auth_logic {
+
+class Parser {
+public:
+
+//Parses an input string
+
+absl::StatusOr<Program> ParseProgram (absl::string_view prog_text);
```
markww, line 36: Because this class has no state, you could imagine eliminating the class entirely, exposing `ParseProgram` as a top-level function, and eliminating all of the other files in the header. If no one else needs to know about them, and they don't need to manipulate any shared state, they are likely better off as `static` functions in the `.cc` file.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,74 @@
...
+         actual_name.push_back(each_query.name()); 
+         actual_principal_name.push_back(each_query.principal().name()); 
+         BaseFact basefact(std::move(std::get<BaseFact>(each_query.fact().GetValue())));
+         actual_predicate_name.push_back(std::get<datalog::Predicate>(basefact.GetValue()).name());
+         for (auto const &each_arg : std::get<datalog::Predicate>(basefact.GetValue()).args() ) {
+            actual_predicate_args.push_back(each_arg);
+         }
+        
+     } 
+
```
markww, line 60: What is happening here? Why are you creating a vector and recording some variable number of entries from `prog_out.queries()`, but then only checking the first element?

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,74 @@
...
+    std::string expected_name = "may_notifierA";
+    std::string expected_principal_name = "\"UserA\"";
+    std::string expected_predicate_name = "may";
+    std::string expected_predicate_args = "\"Notifier\"";
+
+    EXPECT_EQ(actual_name[0], expected_name);
+    EXPECT_EQ(actual_principal_name[0], expected_principal_name);
+    EXPECT_EQ(actual_predicate_name[0], expected_predicate_name);
+    EXPECT_EQ(actual_predicate_args[0], expected_predicate_args);
```
markww, line 69: Don't use `operator[]` on `std::vector`s. It performs a non-bounds-checked access and can overflow or underflow. The `at` method does the same thing, but is safe.

markww (commented):
Left comments, but at a high level:

* Prefer references over pointers whenever applicable. C++ provides some (not enough, but some) type system mechanisms to allow you to be confident that a pointer can never be `nullptr`, and they should be used whenever appropriate.

* `StatusOr` is basically only for cases where an error is expected and/or might be recoverable. If the error should never be able to happen and cannot be recovered from (such as in the cases where an unexpected `dynamic_cast` occurs) use `CHECK` instead, which will allow you to have simpler return types and less error handling logic.

* Mark things `const` when you can. Mutability is one of the big things that can really complicate reasoning about a program and everything that does not need to change should not change. In particular, mark the target of pointers and references `const` when possible (`const T &my_var`) and mark the `this` of methods `const` when possible (`void MyMethod(uint64_t x) const {`).

* Declare locals as close to their use as possible. You should almost never declare a local of non-object type without an initializer, as C++ will not stop you from reading uninitialized memory and running with it.

* If a local is only used once, consider whether you need it at all. Sometimes a local helps the code self-document by breaking up hard-to-read complex expressions, but sometimes it just decontextualizes an expression from where it is used. Plus, C++ can do fun optimizations with unnamed temporaries (but this should not be your primary concern).

* Double check whether variable-length objects should be `std::move`d to their destination. Forgetting a `std::move` on a fixed-size value can introduce a copy, but it won't affect the overall asymptotic complexity of the code. For variable sized objects, however (such as `std::string` and almost any container) you can easily multiply your runtime complexity by some extra `N`. `const` reference parameters and `absl::string_view` also can be used to sidestep this, but additionally introduce a lifetime concern.

src/ir/auth_logic/ast_construction.cc
aferr: Right, NEG can only be nullptr or something else. This is the part of the grammar for negating a predicate, so the negation symbol is either in the code that was parsed or it wasn't.

aferr: Maybe it's safer to call `getText()` when `NEG` is not a `nullptr` and set `sign = datalog::Sign::kNegated` only when the text is the right value (`!`) and throw an exception when NEG is some other non-null value. I think this is also fine as it is, though.

src/ir/auth_logic/ast_construction.h
```diff
@@ -0,0 +1,60 @@
...
+
+private:
+
+absl::StatusOr<Program> ConstructProgram (AuthLogicParser::ProgramContext* context);
```
aferr, line 40: A more conventional way to use a visitor to populate the AST from antlr would be to define a subclass of `AuthLogicBaseVisitor` and then override the `visit` methods. In your implementations of the visit methods, when you encounter a child node you would call `child.accept(this)` to cause the visitor to visit the children. Because the accept method is virtual and CPP has single dynamic dispatch for virtual methods, the compiler will call the implementation of the method that depends on which specific subclass the `accept` is being called on. Normally, this would let you cut out the dynamic casts completely, but because the CPP version of the AST no longer follows the grammar structure exactly, I don't think we can fully avoid dynamic casts. 

Here's an example that illustrates how this would normally look. For a moment, imagine the AST class structure follows the gramma structure exactly and we have a VerbPhrase class like in the [rust version](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/ast.rs).
```
class AstConstructionVisitor : public AuthLogicBaseVisitor {

virtual antlrcpp::Any visitPrinFact(AuthLogicParser::PrinFactContext *ctx) override {
    Principal principal = ctx->principal()->accept(this);
    VerbPhrase verpbPhrase = ctx->verbphrase()->accept(this);    // there is no VerbPhrase class in ast.h
    return PrinPhrase(principal, verbPhrase);                // there is no PrinPhrase either
}

}
```

Here calling `ctx->verbphrase()->accept(this)` is important and how you would get rid of those dynamic casts. The return type of `verbphrase()` is `VerbphraseContext` and both `PredphraseContext` and `ActsAsPhraseContext` are subclasses of `VerbphraseContext`. We do not know at compile-time which subclass of `VerbphraseContext` we will get, but because `accept` is virtual, the implementation of that method that gets called will be handled by a vtable lookup and be the one implemented for the subclass. The code that antlr generates for the accept methods is this:

```
antlrcpp::Any AuthLogicParser::PredphraseContext::accept(tree::ParseTreeVisitor *visitor) {
  if (auto parserVisitor = dynamic_cast<AuthLogicVisitor*>(visitor))
    return parserVisitor->visitPredphrase(this);
  else
    return visitor->visitChildren(this);
}
```

The first branch is the one that happens in our case, so all this does is call `visitPredPhrase(this)`, and the implementation for `ActsAsPhraseContext` is similar except it calls `visitActsAsPhrase`. Because the implementations of the visits would be the ones you define, this is enough to get rid of the dynamic casts -- the dynamic dispatch of `accept` handles this for you.

Unfortunately, in our case the structure of the `AST` doesn't follow the structure of the grammar so I think we can't sue this directly. The real implementation of `visitPrinFact` would look like this:
```
virtual antlrcpp::Any visitPrinFact(AuthLogicParser::PrinFactContext *ctx) override {
    Principal principal = ctx->principal()->accept(this);
    // We know that we should return either the Attribute or CanActAs kind of base fact, but we
    // don't know which one because we don't know what `ctx->verbphrase()` is. To construct
   // either one of these we need to give them the principal though.

   // This doesn't quite work:
   return ctx->verbphrase->accep(this, principal);
  // because we can't change accept to take a principal because accept is defined by antlr
  // So we likely need to handle this with a dynamic cast.
}
```

aferr: Another way to fix this other than changing the structure of the AST would be to change the grammar so that it more closely follows our updated AST -- if we are using the same `.g4` file for the rust and CPP versions we would need to make sure the rust tests all still pass, though.

I think AST construction is actually one of the cases where the visitor pattern does not help much because constructing the AST requires a complete description of the AST traversal anway. The visitor pattern mainly helps when the interesting functionality of the visit only happens for a small part of the tree. I think using a listener for AST construction is actually fine because we need to give a complete description of the traversal anyway.

We can also just go on using dynamic casts in some cases for now.

aferr: I didn't check if `antlrcpp::Any` will actually let you return the various type from `ast.h` (like `BaseFact` for example). If it does, then changing the grammar to match the structure of `ast.h` and extending the `AuthLogicBaseVisitor` might be a good way to solve this.

If `antlrcpp::Any` won't let you return those types, then the visitor really does not help with AST construction and it's probably best to keep going with what you have.

bgogul: I think @harsha-mandadi found out the hard way that the visitors won't work well for our use case. `antrlcpp::Any` interferes with our use of `unique_ptr` in the data structures. `unique_ptr` can only be moved, but the visitor interface expects the values to be copyable, if I understand the issue correctly. @harsha-mandadi, please chime in if I am wrong.

aferr: OK sounds good :)

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,164 @@
...
+#include "src/ir/auth_logic/ast_construction.h"
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicLexer.h" 
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicParser.h"
+#include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/AuthLogicVisitor.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
+
+Principal Parser::ConstructPrincipal (AuthLogicParser::PrincipalContext* context) {
+    std::string name = context->ID()->getText();
```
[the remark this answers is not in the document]
harsha-mandadi: I see few problem when using const reference instead of pointers. Please let me know if you have a way to deal with this.
Original code:
```template<typename T, typename = std::enable_if_t<std::is_base_of<AuthLogicParser::FactContext, T>::value>>
T* GetFactContextAs(AuthLogicParser::FactContext* context) {
     return dynamic_cast<T*>(context);
}

absl::StatusOr<Fact> Parser::ConstructFact(AuthLogicParser::FactContext* context) {
    if (auto* flat_fact_fact_ctx = GetFactContextAs<AuthLogicParser::FlatFactFactContext>(context)) {  
        absl::StatusOr<BaseFact> flatfact_or_null = Parser::ConstructFlatFact(flat_fact_fact_ctx->flatFact());
        if (!flatfact_or_null.ok()) return absl::InvalidArgumentError("Invalid context to ConstructFact");
        BaseFact basefact = *flatfact_or_null;                               //FlatFactFactContext
        return Fact(basefact);
    }
    else if (auto* ctx = GetFactContextAs<AuthLogicParser::CanSayFactContext>(context))  {                                                      
        return Parser::ConstructFact(ctx->fact());
    }

    return absl::InvalidArgumentError("Invalid context to ConstructFact");
}```
Modified code with above suggestions:
```template<typename T, typename = std::enable_if_t<std::is_base_of<AuthLogicParser::FactContext, T>::value>>
T GetFactContextAs(AuthLogicParser::FactContext &context) {
     return dynamic_cast<T&>(context);
}

Fact Parser::ConstructFact(AuthLogicParser::FactContext &context) {
    if (auto flat_fact_fact_ctx = GetFactContextAs<AuthLogicParser::FlatFactFactContext>(context)) {  
        BaseFact basefact = Parser::ConstructFlatFact(flat_fact_fact_ctx->flatFact());
        return Fact(basefact);
    }
  else {
    auto ctx = GetFactContextAs<AuthLogicParser::CanSayFactContext>(context))                                                       
    return Parser::ConstructFact(ctx->fact());
}
 
}```

In the original code, dynamic cast on pointer, if the cast fails, it would return a NULL. This lets us know if the object is of that particular type or not( In our case, flatFactFactContext). If it is not, we check the same for another type (CanSayFactContext).
When casting a reference, it is not possible to return a NULL pointer to indicate failure; a dynamic cast of a reference variable will throw the exception std::bad_cast. This causes a problem as we never get to the else part.

File: src/ir/auth_logic/ast_construction_test.cc (hunk 21 above)
Comment (harsha-mandadi) [reply src/ir/auth_logic/ast_construction_test.cc:54]: Intended to test for multi line queries. Hence used std::vector. Modified the tests to handle multiple queries now.

File: src/ir/auth_logic/ast_construction.cc (hunk 14 above)
Comment (harsha-mandadi) [reply src/ir/auth_logic/ast_construction.cc]: `const auto * ctx` would result in compiler errors as object has type qualifiers not compatible with member functions of AuthLogicParser.

File: src/ir/auth_logic/ast_construction.cc [hunk 25]
```diff
@@ -0,0 +1,153 @@
...
+  if (auto* ctx = GetVerbPhraseContextAs<AuthLogicParser::PredphraseContext>(
+          context)) {  // PredicateContext
+    return BaseFact(
+        Attribute(left_principal, ConstructPredicate(ctx->predicate())));
+  }
+  auto* ctx = GetVerbPhraseContextAs<AuthLogicParser::ActsAsPhraseContext>(
+      context);  // ActsAsPhraseContext
+  return BaseFact(
+      CanActAs(left_principal, ConstructPrincipal(ctx->principal())));
+}
```
bgogul, line 74: A few suggestions: 
  - Let us use dynamic_cast directly? 
  -  Also, take a reference in and use `&` to take the address later. That way you can avoid adding a check for context != nullptr`. 
  - Use `CHECK_NOTNULL` at the last use case to make sure we are not dereferencing nullptr.

Other similar functions could use the same pattern.

```suggestion
static BaseFact ConstructVerbphrase(
    Principal left_principal, AuthLogicParser::VerbphraseContext& context) {
  if (auto* ctx = dynamic_cast<AuthLogicParser::PredphraseContext*>(&context)) {
    // PredicateContext
    return BaseFact(
        Attribute(left_principal, ConstructPredicate(ctx->predicate())));
  }
  // ActsAsPhraseContext
  auto* ctx = CHECK_NOTNULL(dynamic_cast<AuthLogicParser::ActsAsPhraseContext*>(&context);
  return BaseFact(
      CanActAs(left_principal, ConstructPrincipal(ctx->principal())));
}
```

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,153 @@
...
+  CHECK(context != nullptr) << "Invalid FlatFactcontext";
+  if (auto* ctx =
+          GetFlatFactContextAs<AuthLogicParser::PrinFactContext>(context)) {
+    return ConstructVerbphrase(ConstructPrincipal(ctx->principal()),
+                               ctx->verbphrase());
+  }
+  auto* ctx = GetFlatFactContextAs<AuthLogicParser::PredFactContext>(
+      context);  // predFactContext
+  return BaseFact(ConstructPredicate(ctx->predicate()));
+}
```
bgogul, line 92: Use suggested pattern from above?

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,153 @@
...
+
+template <typename T, typename = std::enable_if_t<std::is_base_of<
+                          AuthLogicParser::FactContext, T>::value>>
+T* GetFactContextAs(AuthLogicParser::FactContext* context) {
+  return dynamic_cast<T*>(context);
+}
+
```
bgogul, line 99: Get rid of this and use `dynamic_cast` directly.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,153 @@
...
+static Fact ConstructFact(AuthLogicParser::FactContext* context) {
+  CHECK(context != nullptr) << "Invalid Factcontext";
+  if (auto* flat_fact_fact_ctx =
+          GetFactContextAs<AuthLogicParser::FlatFactFactContext>(
+              context)) {  // FlatFactFactContext
+    return Fact(ConstructFlatFact(flat_fact_fact_ctx->flatFact()));
+  }
+  auto* ctx = GetFactContextAs<AuthLogicParser::CanSayFactContext>(context);
+  return ConstructFact(ctx->fact());
+}
```
bgogul, line 109: Use pattern from above?

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,153 @@
...
+
+static Query ConstructQuery(AuthLogicParser::QueryContext* context) {
```
bgogul, line 111: `QueryContext& context`

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  // for ( auto const &statement : context->saysAssertion()) {
+  //     says_assertions_.push_back(std::move(construct_says_assertion(statement)));
+  // }
+
+  for (auto const& statement : context->query()) {
+    CHECK(statement != nullptr) << "Invalid Querycontext";
+    Query query = ConstructQuery(statement);
+    queries.push_back(std::move(query));
+  }
```
bgogul, line 130: What is the return type of `context->query()?

```suggestion
std::vector<Query> queries = utils::MapIter<Query>(context->query(), [](const auto* statement) {
    return ConstructQuery(CHECK_NOTNULL(statement));
}
```

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,78 @@
...
+#include "src/common/testing/gtest.h"
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
+
+TEST(AstConstructionSuite, SimpleQueryTest) {
+  absl::string_view input_text =
+      R"qql(may_notifierA = query "UserA" says mayA("NotifierA")
+                                         may_notifierB = query "UserB" says mayB("NotifierB"))qql";
```
bgogul, line 38: Which is `qql` needed?
```suggestion
      R"(may_notifierA = query "UserA" says mayA("NotifierA")
                                      may_notifierB = query "UserB" says mayB("NotifierB"))";
```

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,78 @@
...
+
+  Program prog_out = ParseProgram(input_text);
+  std::vector<std::string> actual_name;
+  std::vector<std::string> actual_principal_name;
+  std::vector<std::string> actual_predicate_name;
+  std::vector<std::string> actual_predicate_args;
```
bgogul, line 44: Don't you lose the correlation between different parts of the query by testing the components separately? I think you should directly test the list of query and use matchers (https://github.com/google/googletest/blob/main/docs/reference/matchers.md). It will make the tests more readable and easier to debug in case of failures. 

See suggestion below.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,78 @@
...
+  std::string another_expected_predicate_args = R"("NotifierB")";
+
+  EXPECT_EQ(actual_name.at(0), expected_name);
+  EXPECT_EQ(actual_name.at(1), another_expected_name);
+  EXPECT_EQ(actual_principal_name.at(0), expected_principal_name);
+  EXPECT_EQ(actual_principal_name.at(1), another_expected_principal_name);
+  EXPECT_EQ(actual_predicate_name.at(0), expected_predicate_name);
+  EXPECT_EQ(actual_predicate_name.at(1), another_expected_predicate_name);
+  EXPECT_EQ(actual_predicate_args.at(0), expected_predicate_args);
+  EXPECT_EQ(actual_predicate_args.at(1), another_expected_predicate_args);
```
bgogul, line 75: User matchers to make the test more readable:

```suggestion
Predicate mayA("mayA", {"NotiferA"}, datalogg::Sign::kPositive);
... 
EXPECT_THAT(prog_out.queries(), 
  testing::UnorderedElementsAre(
     Query("may_notifierA", Principal("UserA"), Fact(BaseFact(mayA))),
     Query(...)
```

If you do care about the order user `testing::ElementsAre` instead of `testing::UnorderedElementsAre`.

bgogul (commented):
Style and readability suggestions. PTAL.

src/ir/auth_logic/ast_construction.cc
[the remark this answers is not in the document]
markww: Taking a pointer instead of a reference in checking whether the result of the dynamic cast is `NULL` or not is fine. Cases like the above are exempt from the guidance I offered, as we expect `NULL` to be returned by `dynamic_cast` in the normal running of the program. In general, though, if using a reference type can prevent you from having to check whether a pointer is `NULL`, and all you would do if that pointer was `NULL` would be issue an unrecoverable error, it's better to use the reference type.

src/ir/auth_logic/ast_construction.cc
harsha-mandadi: I am going to keep this as it is. antlr parser would throw a RecognitinException error for illegal input text.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,122 @@
...
+namespace raksha::ir::auth_logic {
+
+static Principal ConstructPrincipal(
+    AuthLogicParser::PrincipalContext& context) {
+  return Principal(context.ID()->getText());
+}
+
+static datalog::Predicate ConstructPredicate(
+    AuthLogicParser::PredicateContext& context) {
+  std::vector<std::string> args = utils::MapIter<std::string>(context.pred_arg(), [](auto* statement) { return statement->ID()->getText(); });
```
markww, line 47: Nit: This looks longer than 80 characters. Please split at 80. Note that the clang formatting tool will do this kind of thing for you and Raksha is set up for it to format with Google style. I'm writing this comment on a tiny laptop screen, so disregard if already small enough.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,122 @@
...
+//   for (auto& statement : context.pred_arg()) {
+//     args.push_back(std::move(statement->ID()->getText()));
+//   }
+  // args =  utils::MapIter<std::string>(context->pred_arg(), [] );
```
markww, line 51: Remove commented out code.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,122 @@
...
+    Principal left_principal, AuthLogicParser::VerbphraseContext& context) {
+  if (auto* ctx = dynamic_cast<AuthLogicParser::PredphraseContext*>(
+          &context)) {  // PredicateContext
+    return BaseFact(
+        Attribute(left_principal, ConstructPredicate(*ctx->predicate())));
+  }
+  auto* ctx = CHECK_NOTNULL(dynamic_cast<AuthLogicParser::ActsAsPhraseContext*>(
+      &context));  // ActsAsPhraseContext
+  return BaseFact(
+      CanActAs(left_principal, ConstructPrincipal(*ctx->principal())));
```
markww, line 67: Nit: you could do `*CHECK_NOTNULL` and put the result into a reference type, thus immediately shifting to a type that cannot receive `nullptr`. This, however, is super picky seeing how this is only used in one place directly afterwards, so feel free to ignore.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,122 @@
...
+}
+
+static BaseFact ConstructFlatFact(AuthLogicParser::FlatFactContext& context) {
+  if (auto* ctx = dynamic_cast<AuthLogicParser::PrinFactContext*>(&context)) {
+    return ConstructVerbphrase(ConstructPrincipal(*ctx->principal()),
+                               *ctx->verbphrase());
```
markww, line 73: Should the results of things like `principal()` and `verbphrase()` also go through `CHECK_NOTNULL` before being dereferenced?

This is more important in C++ that it is in languages like Java or Python. While Java and Python have a special `null` value that is checked by the underlying virtual machine, C++'s handling of `nullptr` is relatively crude: `nullptr` is, in most cases, literally the address `0`. On most machines, to support this, the virtual page mapper intentionally leaves the 0th page unmapped so that any reads or writes to it will immediately segfault.

But there are systems where this is not done or is not possible. For instance, the WINE emulator steals this 0th page for its own bookkeeping. Systems that do not have a virtual mapping layer often use this page, as it is some real, physical memory that would otherwise be dead. There exist cases where it is not safe, and in fact a real memory safety issue, to read or write to address 0. So we have to be careful with our handling of pointers that may be `nullptr`, and try to catch the `nullptr`s before they are dereferenced.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,122 @@
...
+  antlr4::ANTLRInputStream input(prog_text);
+  // Creates a lexer from input
+  AuthLogicLexer lexer(&input);
+  // Creates a token stream from lexer
+  antlr4::CommonTokenStream tokens(&lexer);
+  // Creates a parser from the token stream
+  AuthLogicParser parser(&tokens);
+  // program_context points to the root of the parse tree
+  AuthLogicParser::ProgramContext* program_context = parser.program();
+  return ConstructProgram(*program_context);
```
markww, line 119: Need a `CHECK_NOTNULL` in here?

src/ir/auth_logic/ast_construction.h
```diff
@@ -0,0 +1,35 @@
+
```
markww, line 1: Nit: Remove leading blank line.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,90 @@
...
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    AstConstructionTest, AstConstructionTest,
+    testing::Values(
+        QueryTestData(
+            {R"(may_notifierA = query "UserA" says mayA("NotifierA"))",
+             "may_notifierA", R"("UserA")", "mayA",
+             datalog::Predicate("mayA", {R"("NotifierA")"},
+                                datalog::Sign::kPositive)}),
```
markww, line 83: I want to preface this by saying this is entirely optional.

One thing I often like to do with tests like this, wherein the test is ensuring that some string in a parsed language is taken apart into its components correctly, is to define a format string and build the test data from the component pieces using that format string. For instance, in this case, the format string seems to be of the form:

```C++
R"(%s = query %s says %s%s)"
```

By plugging your expected results into this format string, you eliminate repeating them and possibly a source of bugs.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,90 @@
...
+        QueryTestData(
+            {R"(may_notifierB = query "UserB" says mayB("NotifierB"))",
+             "may_notifierB", R"("UserB")", "mayB",
+             datalog::Predicate("mayB", {R"("NotifierB")"},
+                                datalog::Sign::kPositive)})));
```
markww, line 88: I notice both of the tests test only the `kPositive` case. Should we also test the `kNegative` case?

src/ir/auth_logic/ast_construction.cc
bgogul: +1 just make it something like:
```C++
auto& cxt = *CHECK_NOTNULL(...);
return ... ctx.principal() ...
```

src/ir/auth_logic/ast_construction.cc
bgogul: Doesn't the fact that `if (auto *ctx =..)` true guarantee that `ctx` is not nullptr?

src/ir/auth_logic/ast_construction.cc
bgogul: The following might be a good pattern to follow:
```C++
  AuthLogicParser::ProgramContext& program_context = *CHECK_NOTNULL(parser.program());
  return ConstructProgram(program_context);
```

src/ir/auth_logic/ast_construction_test.cc
bgogul: +1, if you are already handing `kNegative` in this PR.

src/ir/auth_logic/ast_construction_test.cc
bgogul: +1 to use this approach wherever possible. However, in these tests, I don't know if there is enough repetition of the same pattern to do it.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -0,0 +1,90 @@
...
+      GetParam();
+  Program prog_out = ParseProgram(input_text);
+  EXPECT_THAT(prog_out.queries(),
+              ElementsAre(Property(&Query::name, query_name)));
+  EXPECT_THAT(prog_out.queries(),
+              ElementsAre((ResultOf(GetQueryPrincipalName, principal_name))));
+  EXPECT_THAT(prog_out.queries(),
+              ElementsAre((ResultOf(GetQueryFactName, fact_name))));
+  EXPECT_THAT(prog_out.queries(),
+              ElementsAre((ResultOf(GetQueryFactArgs, predicate))));
```
bgogul, line 73: Since you only expect the result to be a single query, I think we could test more directly:

```C++
  ASSERT_EQ(prog_out.queries.size() == 1);
  const Query& query = prog_out.queries().front();
  EXPECT_EQ(query.name(), query_name);
  ...
```

bgogul (commented):
Looks mostly good except for minor style issues.

src/ir/auth_logic/ast_construction.cc
markww: It guarantees that `ctx` is not `nullptr`, but I don't think there's any guarantees on the result of `ctx->principal()` which is also being dereferenced via the `*`.

bgogul: aah.. missed that.

src/ir/auth_logic/ast_construction_test.cc
harsha-mandadi: Yeah. I tried to test with multiple queries with a slightly different approach. Have parameterized tests `GetParams()` be our actual testing inputs and compare it against the expected values of prog_out.queries(). Unfortunately this approach was giving out errors. Hence decided to test on single query.

src/ir/auth_logic/ast_construction_test.cc
harsha-mandadi: I am going to take a note of this and use this next time!

src/ir/auth_logic/ast_construction_test.cc
harsha-mandadi: Modified test to include kNegative!

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,123 @@
...
+static Principal ConstructPrincipal(
+    AuthLogicParser::PrincipalContext& context) {
+  return Principal(context.ID()->getText());
+}
+
+static datalog::Predicate ConstructPredicate(
+    AuthLogicParser::PredicateContext& context) {
+  std::vector<std::string> args = utils::MapIter<std::string>(
+      context.pred_arg(),
+      [](auto* statement) { return statement->ID()->getText(); });
```
aferr, line 48: very optional nit: the variable_name `statement` is slightly confusing because you're iterating over predicate arguments rather than statements. A better name might be `arg` or `pred_arg`.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -0,0 +1,123 @@
...
+}
+
+static Query ConstructQuery(AuthLogicParser::QueryContext& context) {
+  return Query(context.ID()->getText(),
+               ConstructPrincipal(*CHECK_NOTNULL(context.principal())),
+               ConstructFact(*CHECK_NOTNULL(context.fact())));
+}
+
+static Program ConstructProgram(AuthLogicParser::ProgramContext& context) {
+  std::vector<SaysAssertion> says_assertions;
```
aferr, line 98: I guess parsing `says_assertions` is still WIP? It looks like these don't get built yet. (I just wanted to make sure this wasn't missed accidentally).

harsha-mandadi: Andrew, says_assertion is still a WIP and will create another PR after this loop is closed

src/policy/auth_logic/BUILD
```diff
@@ -17,6 +17,7 @@
 load("//build_defs:antlr.bzl", "antlr4_cc_combined")
 
 package(
+    default_visibility = ["//visibility:public"],
```
markww, line 20: A `public` default visibility is generally discouraged. Try to enumerate the directories where you need this. If you find whole subtrees of the repo need this, feel free to use the `:__subpackages__` target.

harsha-mandadi: Looks like making this change does not resolve the issue

markww (approved):
LGTM

bgogul (approved):

PR #489, opened 2022-03-24
Change the `TagTransform` preconditions to be a map from name to ID.

markww:
While integrating with the SQL verifier that we wish to integrate with
Raksha, I noticed that it associates each of the preconditions on its
tag transforms with a name in an unordered map. This is an awkward fit
with the way we had originally written the `TagTransform`, which would
associate its preconditions with the preconditions of a policy rule by
an order imposed on the preconditions. For one, it would require me to
repeatedly sort the preconditions on the verifier side when building the
protos, repeatedly performing an O(n * log(n)) operation. For another,
the names would preserve debugging information in Datalog and probably
produce better results if a mechansim like Souffle's `explain` were
used.

For these reasons, this PR modifies the `TagTransform` preconditions to
be a map from precondition name to the ID of the AST node that will be
associated with that precondition. These names are used as the names of
the `input`s to the `TagTransform`, and can be matched by the policy
datalog produced from the SQL verifier's rules.

### DIFF
--- a/src/frontends/sql/decode_tag_transform_operation_test.cc
+++ b/src/frontends/sql/decode_tag_transform_operation_test.cc
@@ -57,9 +57,9 @@ using ::testing::ValuesIn;
 using utils::MapIter;
 
 class DecodeTagTransformTest
-    : public ::testing::TestWithParam<
-          std::tuple<uint64_t, absl::string_view,
-                     std::pair<absl::string_view, std::vector<uint64_t>>>> {};
+    : public ::testing::TestWithParam<std::tuple<
+          uint64_t, absl::string_view, std::vector<absl::string_view>,
+          std::pair<absl::string_view, std::vector<uint64_t>>>> {};
 
 TEST_P(DecodeTagTransformTest, DecodeTagTransformTest) {
   IRContext ir_context;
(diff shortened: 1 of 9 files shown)


src/frontends/sql/decoder_context.cc
```diff
@@ -48,16 +48,12 @@ const ir::Operation &DecoderContext::MakeTagTransformOperation(
...
       {std::string(kTagTransformTransformedValueInputName), transformed_value});
 
-  // Look up each precondition ID to get the corresponding Value.
-  std::vector<ir::Value> precondition_values_vec = utils::MapIter<ir::Value>(
-      preconditions, [&](uint64_t id) { return GetValue(id); });
-
   // Each precondition is given the same input name prefix plus an incrementing
   // number in the same order as in the original ID list.
-  for (uint64_t i = 0; i < precondition_values_vec.size(); ++i) {
+  for (auto &[name, id] : preconditions) {
```
bgogul, line 53: const auto?

src/frontends/sql/testing/operation_view_utils.h
```diff
@@ -25,7 +25,15 @@ std::optional<uint64_t> ExtractIdxAfterPrefix(
...
 // vector. If the density assumption is not met, this method asserts.
 std::vector<ir::Value> GetVecWithPrefix(const ir::NamedValueMap &map,
-                                               absl::string_view prefix);
+                                        absl::string_view prefix);
+
+// This method assumes that, within the given `NamedValueMap`, there are a
+// subset of elements that start with some prefix `prefix` and end with a key
+// that should be associated with that value in the resulting map. It
+// composes that map and returns it.
+absl::flat_hash_map<std::string, ir::Value> GetMapWithPrefix(
```
bgogul, line 34: `GetElementsWithPrefix` or `GetMapEntriesWithPrefix` might be a better name?

src/frontends/sql/decoder_context.cc
```diff
@@ -38,7 +38,7 @@ const Operation &DecoderContext::MakeMergeOperation(
 
 const ir::Operation &DecoderContext::MakeTagTransformOperation(
     ir::Value transformed_value, absl::string_view rule_name,
-    std::vector<uint64_t> preconditions) {
+    absl::flat_hash_map<std::string, uint64_t> preconditions) {
```
bgogul, line 41: Since you are not making a copy of the map, should you change this to `const &`?

src/frontends/sql/decode_tag_transform_operation_test.cc
```diff
@@ -70,22 +70,41 @@ TEST_P(DecodeTagTransformTest, DecodeTagTransformTest) {
...
+                                    }),
+                      value_textproto),
       &expr));
   Value decoded_value = DecodeExpression(expr, decoder_context);
 
-  std::vector<Value> expected_values =
-      MapIter<Value>(value_ids, [&decoder_context](uint64_t current_id) {
-        return decoder_context.GetValue(current_id);
-      });
+  std::vector<std::pair<std::string, Value>> expected_name_to_value_vec =
```
bgogul, line 99: auto `expected_name_to_value_vec` might be more readable?

bgogul (approved):
Mostly style suggestions.

src/frontends/sql/decoder_context.cc
markww: Fixed.

src/frontends/sql/testing/operation_view_utils.h
markww: Went with `GetMapEntriesWithPrefix`.

src/frontends/sql/decoder_context.cc
markww: OK, sounds good.

src/frontends/sql/decode_tag_transform_operation_test.cc
markww: Certainly more concise, and (most of) the type is explicitly written on the RHS. Fixed.

PR #491, opened 2022-03-25
Make inputs to an operation to be simply an ordered list.

bgogul:
Having a named list is quite cumbersome to deal with. If needed, we can attach names to the arguments using an alternate mechanism. e.g., the definition of operator could have a list of names and types.

### DIFF
--- a/src/frontends/sql/decoder_context.cc
+++ b/src/frontends/sql/decoder_context.cc
@@ -33,8 +33,6 @@ const ir::Operation &DecoderContext::MakeTagTransformOperation(
   // We need to reserve one space for the transformed value plus one for each
   // precondition.
   ir::ValueList inputs;
-  // Not adding +1 to avoid potential overflow.
-  inputs.reserve(preconditions.size());
 
   // Insert the value to be transformed.
   inputs.push_back(transformed_value);
--- a/src/frontends/sql/decoder_context.cc
+++ b/src/frontends/sql/decoder_context.cc
@@ -23,8 +23,8 @@ const Operation &DecoderContext::MakeMergeOperation(
                                   static_cast<int64_t>(control_start_index))}});
   // Combine the direct and control inputs.
   absl::c_move(control_inputs, std::back_inserter(direct_inputs));
-  return top_level_block_builder_.AddOperation(merge_operator_, attributes,
-                                               std::move(direct_inputs));
+  return top_level_block_builder_.AddOperation(
+      merge_operator_, std::move(attributes), std::move(direct_inputs));
 }
 
 const ir::Operation &DecoderContext::MakeTagTransformOperation(
--- a/src/frontends/sql/decoder_context.cc
+++ b/src/frontends/sql/decoder_context.cc
@@ -1,6 +1,8 @@
 #include "src/frontends/sql/decoder_context.h"
 
 #include "src/common/utils/map_iter.h"
+#include "src/ir/attributes/attribute.h"
+#include "src/ir/attributes/int_attribute.h"
 
 namespace raksha::frontends::sql {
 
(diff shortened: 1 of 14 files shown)


src/frontends/sql/testing/merge_operation_view.h
```diff
@@ -32,17 +33,30 @@ class MergeOperationView {
   MergeOperationView(const ir::Operation &merge_operation)
       : merge_operation_(&merge_operation) {
     CHECK(merge_operation.op().name() == DecoderContext::kSqlMergeOpName);
-    CHECK(merge_operation.attributes().empty());
+    CHECK(merge_operation.attributes().count(
+              DecoderContext::kMergeOpControlStartIndex) > 0);
```
markww, line 37: Might also want to add a check that it is less than the total number of inputs.

src/ir/operation_test.cc
```diff
@@ -95,17 +97,17 @@ INSTANTIATE_TEST_SUITE_P(
...
              {{"const", Attribute::Create<Int64Attribute>(10)}},
-             {{"a", Value(value::BlockArgument(OperationTest::first_block_,
-                                               "arg0"))},
-              {"b", Value(value::BlockArgument(OperationTest::first_block_,
-                                               "arg1"))}},
-             "%0 = core.minus [const: 10](a: %0.arg0, b: %0.arg1)\n"})));
+             {Value(value::BlockArgument(OperationTest::first_block_, "arg0")),
+              Value(value::BlockArgument(OperationTest::first_block_, "arg1"))},
+             "%0 = core.minus [const: "
+             "10](%0.arg0, %0.arg1)\n"})));
```
markww, line 110: Looks like your string got sheared here. Might want to put it all on one line to improve readability.

src/frontends/sql/decoder_context.cc
```diff
@@ -9,61 +11,45 @@ using ir::NamedValueMap;
...
 
   // Insert the value to be transformed.
-  values_map.insert(
-      {std::string(kTagTransformTransformedValueInputName), transformed_value});
+  inputs.push_back(transformed_value);
 
   // Each precondition is given the same input name prefix plus an incrementing
   // number in the same order as in the original ID list.
+  ir::NamedAttributeMap attributes;
+  unsigned index = 1;
```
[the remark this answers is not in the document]
bgogul: declared int64_t and added a bounds check.

src/frontends/sql/testing/merge_operation_view.h
bgogul: done. needed to jump through too many hoops!

src/ir/operation_test.cc
bgogul: done

src/frontends/sql/decoder_context.cc
```diff
@@ -9,61 +11,45 @@ using ir::NamedValueMap;
...
-          "control_input_", std::move(control_inputs),
-          MakeNamedValueMapFromVector("direct_input_", std::move(direct_inputs),
-                                      {})));
+  auto control_start_index = direct_inputs.size();
+  auto attributes = ir::NamedAttributeMap(
+      {{std::string(kMergeOpControlStartIndex),
+        ir::Attribute::Create<ir::Int64Attribute>(control_start_index)}});
+  // Combine the direct and control inputs.
+  absl::c_move(control_inputs, std::back_inserter(direct_inputs));
+  return top_level_block_builder_.AddOperation(merge_operator_, attributes,
```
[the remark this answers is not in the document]
bgogul: done

bgogul:
Addressed comments. PTAL whenever you get a chance @markww

src/frontends/sql/decoder_context.cc
```diff
@@ -9,61 +13,51 @@ using ir::NamedValueMap;
...
 const ir::Operation &DecoderContext::MakeTagTransformOperation(
     ir::Value transformed_value, absl::string_view rule_name,
     const absl::flat_hash_map<std::string, uint64_t> &preconditions) {
   // We need to reserve one space for the transformed value plus one for each
   // precondition.
-  NamedValueMap values_map;
-  values_map.reserve(preconditions.size() + 1);
+  ir::ValueList inputs;
+  // Not adding +1 to avoid potential overflow.
+  inputs.reserve(preconditions.size());
```
markww, line 37: I think the `+ 1` is fine, actually. If we overflow a `size_t`, it will roll right back to `0`, and `reserve`ing 0 is not a correctness concern, just a performance one. If you don't like the overflow because it is weird to reason about, I think we should just drop it and just lean on `std::vector`'s amortization argument for performance.

src/frontends/sql/decoder_context.cc
```diff
@@ -9,61 +13,51 @@ using ir::NamedValueMap;
...
+
+  attributes.insert({std::string(kTagTransformRuleAttributeName),
+                     ir::Attribute::Create<ir::StringAttribute>(rule_name)});
+  int64_t index = 1;
   for (const auto &[name, id] : preconditions) {
-    auto insert_result = values_map.insert(
-        {absl::StrCat(kTagTransformPreconditionInputPrefix, name),
-         GetValue(id)});
+    CHECK(index != std::numeric_limits<int64_t>::max())
+        << "`int64_t` might overflow. Why would you create so many attributes?";
```
markww, line 51: Lol :laughing:

markww (approved):
Looks good, just a nit about `reserve`.

src/frontends/sql/decoder_context.cc
```diff
@@ -9,61 +13,51 @@ using ir::NamedValueMap;
...
 const Operation &DecoderContext::MakeMergeOperation(
     std::vector<Value> direct_inputs, std::vector<Value> control_inputs) {
-  // Add the direct and control inputs into the merge with predictable
-  // prefixes.
+  size_t control_start_index = direct_inputs.size();
+  CHECK(control_start_index < std::numeric_limits<int64_t>::max());
+  auto attributes =
+      ir::NamedAttributeMap({{std::string(kMergeOpControlStartIndex),
+                              ir::Attribute::Create<ir::Int64Attribute>(
+                                  static_cast<int64_t>(control_start_index))}});
```
markww, line 23: Optionally, it might be helpful to move this logic into a helper factory on `Int64Attribute`, called something like `CreateFromUint64` or something. That way you don't need to worry about it at every callsite.

bgogul: SG. I will send this out as separate PR so that I can add some tests.

bgogul: Good Idea, btw.

src/frontends/sql/decoder_context.cc
bgogul: Will remove reserve altogether.

src/frontends/sql/decoder_context.cc
bgogul: :D

PR #492, opened 2022-03-28
Create sql:MergeOp for modeling merge operations in SQL.

bgogul:


src/frontends/sql/decoder_context.h
```diff
@@ -33,8 +33,7 @@ namespace raksha::frontends::sql {
 class DecoderContext {
  public:
   static constexpr absl::string_view kSqlMergeOpName = "sql.merge";
-  static constexpr absl::string_view kMergeOpControlStartIndex =
-      "control_start_index";
+
```
markww, line 36: Should this blank line be removed?

src/frontends/sql/ops/merge_op.cc
```diff
@@ -0,0 +1,80 @@
...
+}
+
+ir::ValueRange MergeOp::GetControlInputs() const {
+  return utils::make_range(inputs().begin() + GetControlInputStartIndex(),
+                           inputs().end());
+}
+
+size_t MergeOp::GetControlInputStartIndex() const {
+  auto find_result = attributes().find(kControlInputStartIndex);
+  CHECK(find_result != attributes().end());
```
markww, line 72: This is optional, but I try not to call the same function twice and expect that the value is the same in case some kind of side effect gets added. Maybe add a local like so:

```C++
auto attributes = attributes();
auto find_result = attributes.find(kControlInputStartIndex);
CHECK(find_result != attributes.end());
```

src/frontends/sql/ops/merge_op_test.cc
```diff
@@ -0,0 +1,130 @@
...
+  MergeOpTest() { SqlOp::RegisterOperator<MergeOp>(context_); }
+
+ protected:
+  ir::IRContext context_;
+};
+
+using ::testing::ElementsAreArray;
+using ::testing::Eq;
+using ::testing::Pair;
+using ::testing::UnorderedElementsAre;
```
markww, line 42: Should the `using` declarations be up at the top of the `namespace`?

markww (approved):
Looks good other than some nits.

src/frontends/sql/ops/merge_op.cc
bgogul: done.

src/frontends/sql/ops/merge_op_test.cc
bgogul: done.

PR #495, opened 2022-03-29
Refactor auth logic data structures in datalog namespace

harsha-mandadi:
DLIRProgram -> Program
DLIRConditionalAssertion -> Rule.
Remove DLIRAssertion completely. Instead create a Rule with no right hand side.
Change Program to have a vector of Rule instances.

### DIFF
--- a/src/ir/auth_logic/lowering_ast_datalog.cc
+++ b/src/ir/auth_logic/lowering_ast_datalog.cc
@@ -65,7 +65,7 @@ datalog::Predicate CanActAsToDLIR(const CanActAs& can_act_as) {
 }
 }  // namespace
 
-datalog::DLIRAssertion LoweringToDatalogPass::SpokenAttributeToDLIR(
+datalog::Rule LoweringToDatalogPass::SpokenAttributeToDLIR(
     const Principal& speaker, const Attribute& attribute) {
   // Attributes interact with "canActAs" because if "Y canActAs X"
   // then Y also picks up X's attributes. We need to generate
@@ -84,23 +84,17 @@ datalog::DLIRAssertion LoweringToDatalogPass::SpokenAttributeToDLIR(
   // This is `speaker says Y canActAs X`
   datalog::Predicate y_can_act_as_x =
       CanActAsToDLIR(CanActAs(prin_y, attribute.principal()));
-  datalog::Predicate speaker_says_y_can_act_as_x =
-      PushPrincipal("says_", speaker, y_can_act_as_x);
 
   // This is `speaker says X PredX`
   datalog::Predicate main_predicate = AttributeToDLIR(attribute);
-  datalog::Predicate speaker_says_x_pred =
-      PushPrincipal("says_", speaker, main_predicate);
 
   // This is the full generated rule:
   // `speaker says Y PredX :-
   //    speaker says Y canActAs X, speaker says X PredX`
-  return datalog::DLIRAssertion(datalog::DLIRCondAssertion(
-      generated_lhs, {std::move(speaker_says_y_can_act_as_x),
-                      std::move(speaker_says_x_pred)}));
+  return datalog::Rule(generated_lhs, {PushPrincipal("says_", speaker, y_can_act_as_x), PushPrincipal("says_", speaker, main_predicate)});
 }
 
-datalog::DLIRAssertion LoweringToDatalogPass::SpokenCanActAsToDLIR(
+datalog::Rule LoweringToDatalogPass::SpokenCanActAsToDLIR(
     const Principal& speaker, const CanActAs& can_act_as) {
   // "canActAs" facts are passed to principals via other canActAs facts in
   // essentially the same way as attributes. This function adds extra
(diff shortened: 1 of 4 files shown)


src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -197,8 +193,7 @@ LoweringToDatalogPass::FactToDLIR(const Principal& speaker, const Fact& fact) {
                 "says_", speaker,
                 PushPrincipal("canSay_", fresh_principal, inner_fact_dlir));
             auto rhs = {fresh_prin_says_inner, speaker_says_fresh_cansay_inner};
-            datalog::DLIRAssertion generated_rule(
-                datalog::DLIRCondAssertion(lhs, rhs));
+            datalog::Rule generated_rule(lhs, rhs);
             gen_rules.push_back(generated_rule);
```
markww, line 197: Get rid of `generated_rule` and inline like so:

```C++
gen_rules.push_back(datalog::Rule(lhs, rhs));
```

I'd take a look at all of these locals which are used only once and see how many you can eliminate by inlining without sacrificing readability. I suspect many of them should be `std::move`d even if the temporaries should stay.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -214,17 +209,17 @@ LoweringToDatalogPass::FactToDLIR(const Principal& speaker, const Fact& fact) {
...
-std::vector<datalog::DLIRAssertion>
+std::vector<datalog::Rule>
 LoweringToDatalogPass::GenerateDLIRAssertions(const Principal& speaker,
                                               const Fact& fact) {
   auto [fact_predicate, generated_rules] = FactToDLIR(speaker, fact);
-  datalog::DLIRAssertion main_assertion =
-      datalog::DLIRAssertion(PushPrincipal("says_", speaker, fact_predicate));
+  datalog::Rule main_assertion =
+      datalog::Rule(PushPrincipal("says_", speaker, fact_predicate));
   generated_rules.push_back(main_assertion);
```
markww, line 218: Please inline `main_assertion`'s initializer into the `push_back` argument.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -235,13 +230,12 @@ LoweringToDatalogPass::GenerateDLIRAssertions(
       });
   auto [dlir_lhs, gen_rules] = FactToDLIR(speaker, conditional_assertion.lhs());
   auto dlir_lhs_prime = PushPrincipal("says_", speaker, dlir_lhs);
-  datalog::DLIRAssertion dlir_assertion(
-      datalog::DLIRCondAssertion(dlir_lhs_prime, dlir_rhs));
-  gen_rules.push_back(dlir_assertion);
+  datalog::Rule dlir_rule(dlir_lhs_prime, dlir_rhs);
+  gen_rules.push_back(dlir_rule);
```
markww, line 234: Inline initializer.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -269,40 +263,39 @@ std::vector<datalog::DLIRAssertion> LoweringToDatalogPass::SaysAssertionToDLIR(
   return ret;
 }
```
markww, line 265: Could this be turned into a single `MapIter` call?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -269,40 +263,39 @@ std::vector<datalog::DLIRAssertion> LoweringToDatalogPass::SaysAssertionToDLIR(
...
-std::vector<datalog::DLIRAssertion> LoweringToDatalogPass::SaysAssertionsToDLIR(
+std::vector<datalog::Rule> LoweringToDatalogPass::SaysAssertionsToDLIR(
     const std::vector<SaysAssertion>& says_assertions) {
-  std::vector<datalog::DLIRAssertion> ret = {};
+  std::vector<datalog::Rule> ret = {};
   for (const SaysAssertion& says_assertion : says_assertions) {
     absl::c_move(SaysAssertionToDLIR(says_assertion), std::back_inserter(ret));
   }
   return ret;
 }
```
markww, line 273: Could this be turned into a single `MapIter` call?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -269,40 +263,39 @@ std::vector<datalog::DLIRAssertion> LoweringToDatalogPass::SaysAssertionToDLIR(
...
-  datalog::DLIRAssertion dummy_assertion(kDummyPredicate);
+  datalog::Rule dummy_assertion(kDummyPredicate);
   absl::c_move(QueriesToDLIR(program.queries()),
                std::back_inserter(dlir_assertions));
   dlir_assertions.push_back(dummy_assertion);
 
   auto outputs = utils::MapIter<std::string>(
       program.queries(), [](const Query& query) { return query.name(); });
-  return datalog::DLIRProgram(dlir_assertions, outputs);
+  return datalog::Program(dlir_assertions, outputs);
```
markww, line 298: Should probably just inline the initializer of `outputs`.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -74,29 +74,29 @@ class SouffleEmitter {
...
                         absl::StrJoin(rhs_translated, ", "), ".");
   }
 
-  std::string EmitAssertion(const datalog::DLIRAssertion& assertion) {
-    return std::visit([this](auto value) { return EmitAssertionInner(value); },
-                      assertion.GetValue());
+  std::string EmitAssertion(const datalog::Rule& assertion) {
+    if (assertion.rhs().empty()) return EmitAssertionInner(assertion.lhs()); 
+    return EmitAssertionInner(assertion);
   }
```
markww, line 88: I'm not sure that the split between `EmitAssertion` and `EmitAssertionInner` serves readability here. Instead of having the `EmitAssertionInner`, imagine that we had a single `EmitAssertion` function that looked like this:

```C++
  std::string EmitAssertionInner(
      const datalog::Rule& cond_assertion) {
    std::vector rhs_translated = utils::MapIter<std::string>(
        cond_assertion.rhs(),
        [this](const datalog::Predicate& arg) { return EmitPredicate(arg); });
    return absl::StrCat(EmitPredicate(cond_assertion.lhs()), (rhs_translated.size() == 0) ? "" : absl::StrCat(" :- ",
                        absl::StrJoin(rhs_translated, ", ")), ".");
  }

markww (approved):
I left a lot of comments, but the majority of them were present before your changes. You can decide for yourself how much appetite you have to handle them, consider all of them optional. If you did choose to fix one, though, I'd refactor `EmitAssertion` to remove the `EmitAssertionInner` functions. OK to split into a separate PR if you do tackle them too.

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: I think using output variable is better in terms of readability.

src/ir/auth_logic/lowering_ast_datalog.cc
bgogul: MapIter doesn't work well with `move`?

bgogul (approved):

src/ir/auth_logic/lowering_ast_datalog.cc
markww: Good point, forgot about that.

src/ir/auth_logic/lowering_ast_datalog.cc
markww: Yep, you're right, makes sense.

PR #496, opened 2022-03-30
Add a safe static_cast where we compare ints of different signedness

bgogul:


markww (approved):
LGTM

bgogul:
Thanks, @markww The compiler on glinux boxes is strict. As you suggested elsewhere, it might worth adding some unsigned int64_t attribute to avoid these kinds of casts.

PR #497, opened 2022-03-30
Make each particle name have its own index space within a recipe.

bgogul:
This will make the names of particles in the access paths more predictable.


aferr (approved):
LGTM.

bgogul:
@markww FYI also.

markww (approved):
LGTM. Much better, much more predictable.

PR #498, opened 2022-03-30
When generating datalog, use id (if available) of a handle as its name.

bgogul:


src/xform_to_datalog/manifest_datalog_facts.cc
```diff
@@ -55,6 +55,11 @@ ManifestDatalogFacts ManifestDatalogFacts::CreateFromManifestProto(
   uint64_t recipe_num = 0;
   for (const arcs::RecipeProto &recipe_proto : manifest_proto.recipes()) {
     absl::flat_hash_map<std::string, uint64_t> particle_numbers;
+    absl::flat_hash_map<std::string, std::string> handle_ids;
+    for (const arcs::HandleProto &handle_proto : recipe_proto.handles()) {
+      handle_ids[handle_proto.name()] =
+          handle_proto.id().empty() ? handle_proto.name() : handle_proto.id();
```
markww, line 61: The behavior on `operator[]` for C++ maps is to get the mutable reference to the value mapped to the provided key, creating a new entry if the key was not previously present. Is that the behavior we want here? Usually, if we are relying upon the keys being unique in advance (not uniqued by their insertion into the map) I think it's better to use `insert` and `CHECK` that the returned pair's `second` field is `true`, thereby ensuring that a fresh insertion always occurs.

src/xform_to_datalog/manifest_datalog_facts_test.cc
```diff
@@ -370,8 +373,8 @@ static const std::string kExpectedEdgeStrings[] = {
     // Named recipe edges:
     // Edges connecting h1 to NamedR.PS1#0 for fields {field1, field2}.
     R"(// Edges:)",
-    R"(edge("NamedR.h1.field1", "NamedR.PS1.in_handle.field1").)",
-    R"(edge("NamedR.h1.field2", "NamedR.PS1.in_handle.field2").)",
+    R"(edge("NamedR.NamedH1.field1", "NamedR.PS1.in_handle.field1").)",
+    R"(edge("NamedR.NamedH1.field2", "NamedR.PS1.in_handle.field2").)",
```
markww, line 377: Maybe update comment above these test assertions.

markww (approved):
LGTM modulo comment about `operator[]` on maps.

src/xform_to_datalog/manifest_datalog_facts_test.cc
bgogul: done.

src/xform_to_datalog/manifest_datalog_facts.cc
bgogul: Done.

PR #511, opened 2022-04-07
Add package and hdrs to antlr.bzl

harsha-mandadi:


build_defs/antlr.bzl
```diff
@@ -35,12 +35,13 @@ def antlr4_cc_combined(name, src, listener = False, visitor = True):
         language = "Cpp",
         listener = listener,
         visitor = visitor,
-        package = "",
+        package = "auth_logic_cc_generator",
```
bgogul, line 38: You shouldn't hard code it to `auth_logic_cc_generator`. Simply remove `package = ""`. That should work I think.

harsha-mandadi: oh I thought doing this way, keeps the whole AuthLogicParser and AuthLogicLexer under the auth_logic_cc_generator namespace. Otherwise`using auth_logic_cc_generator::AuthLogicParser` in `ast_construction.cc` would fail.

harsha-mandadi: Removing package totally fails build on the above mentioned using statement

bgogul: `package = name,` should work and I think it is compatible with the internal behavior.

bgogul (approved):
Approving modulo `package = name` change.

PR #513, opened 2022-04-12
Arcs frontend: use id (if available) of a handle as a handle's name.

bgogul:
This will help us get predictable names the correspond to arcs handles in datalog.


bgogul:
This is a duplicate of #498

PR #521, opened 2022-04-13
auth_logic AST construction for program assertions

harsha-mandadi:


src/ir/auth_logic/ast_construction.cc
```diff
@@ -78,6 +77,19 @@ static BaseFact ConstructFlatFact(AuthLogicParser::FlatFactContext& context) {
...
+static BaseFact ConstructRvalue(AuthLogicParser::RvalueContext& context) {
+  if (auto* ctx =
+          dynamic_cast<AuthLogicParser::FlatFactRvalueContext*>(&context)) {
+    return ConstructFlatFact(*CHECK_NOTNULL(ctx->flatFact()));
+  }
+  auto& ctx = *CHECK_NOTNULL(
+      dynamic_cast<AuthLogicParser::BinopRvalueContext*>(&context));
+  std::vector<std::string> args = utils::MapIter<std::string>(
+      ctx.pred_arg(), [](auto* pred_arg) { return pred_arg->getText(); });
+  return BaseFact(datalog::Predicate(ctx.binop()->getText(), args,
```
bgogul, line 89: `std::move(args)`

src/ir/auth_logic/ast_construction.cc
```diff
@@ -90,14 +102,53 @@ static Fact ConstructFact(AuthLogicParser::FactContext& context) {
...
+      ConstructFact(*CHECK_NOTNULL(ctx.fact())), std::move(base_facts)));
+}
+
+static SaysAssertion ConstructSaysAssertion(
+    AuthLogicParser::SaysAssertionContext& context) {
+  if (auto* ctx = dynamic_cast<AuthLogicParser::SaysSingleContext*>(&context)) {
+    std::vector<Assertion> assertions;
+    assertions.push_back(ConstructAssertion(*CHECK_NOTNULL(ctx->assertion())));
+    return SaysAssertion(ConstructPrincipal(*CHECK_NOTNULL(ctx->principal())),
+                         std::move(assertions));
```
bgogul, line 127: Optional, but the following would also be OK:
```suggestion
    return SaysAssertion(ConstructPrincipal(*CHECK_NOTNULL(ctx->principal())),
                         {ConstructAssertion(*CHECK_NOTNULL(ctx->assertion()))});
```

src/ir/auth_logic/ast_construction.cc
```diff
@@ -90,14 +102,53 @@ static Fact ConstructFact(AuthLogicParser::FactContext& context) {
...
+  }
+  auto& ctx = *CHECK_NOTNULL(
```
bgogul, line 129: `ctx` -> `says_multi_context`

src/ir/auth_logic/ast_construction.cc
```diff
@@ -90,14 +102,53 @@ static Fact ConstructFact(AuthLogicParser::FactContext& context) {
...
+      ctx.rvalue(), [](AuthLogicParser::RvalueContext* rvalue_context) {
+        return ConstructRvalue(*CHECK_NOTNULL(rvalue_context));
+      });
+  return Assertion(ConditionalAssertion(
+      ConstructFact(*CHECK_NOTNULL(ctx.fact())), std::move(base_facts)));
+}
+
+static SaysAssertion ConstructSaysAssertion(
+    AuthLogicParser::SaysAssertionContext& context) {
+  if (auto* ctx = dynamic_cast<AuthLogicParser::SaysSingleContext*>(&context)) {
```
bgogul, line 123: Don't abbreviate unnecessarily. Meaningful names makes it easier to read code: `ctx` -> `says_single_context`

src/ir/auth_logic/ast_construction.cc
```diff
@@ -90,14 +102,53 @@ static Fact ConstructFact(AuthLogicParser::FactContext& context) {
   return ConstructFact(*CHECK_NOTNULL(ctx.fact()));
 }
 
+static Assertion ConstructAssertion(
+    AuthLogicParser::AssertionContext& context) {
+  if (auto* ctx =
```
bgogul, line 107: `ctx` -> `fact_assertion_context`

src/ir/auth_logic/ast_construction.cc
```diff
@@ -90,14 +102,53 @@ static Fact ConstructFact(AuthLogicParser::FactContext& context) {
...
+          dynamic_cast<AuthLogicParser::FactAssertionContext*>(&context)) {
+    return Assertion(ConstructFact(*CHECK_NOTNULL(ctx->fact())));
+  }
+  auto& ctx = *CHECK_NOTNULL(
```
bgogul, line 111: `ctx` -> `horn_clause_assertion_context`

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -40,9 +40,20 @@ struct QueryTestData {
...
 
-class AstConstructionTest : public testing::TestWithParam<QueryTestData> {};
+struct SaysAssertionTestData {
+  std::string input_text;
+  std::string principal_name;
+  std::string fact_assertion_name;
+  datalog::Predicate cond_assertion_predicate;
+  datalog::Predicate rvalue;
+};
+
```
bgogul, line 50: Move this to after the `INSTANTIATE_TEST_SUITE_P(QueryAstConstructionTest,...)`

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -40,9 +40,20 @@ struct QueryTestData {
...
+class QueryAstConstructionTest : public testing::TestWithParam<QueryTestData> {
+};
+class SaysAssertionAstConstructionTest
+    : public testing::TestWithParam<SaysAssertionTestData> {};
```
bgogul, line 54: Move this to after `after the INSTANTIATE_TEST_SUITE_P(QueryAstConstructionTest, ...)`

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -73,4 +84,62 @@ INSTANTIATE_TEST_SUITE_P(
              datalog::Predicate("mayB", {R"("NotifierB")"},
                                 datalog::Sign::kNegated)})));
 
+TEST_P(SaysAssertionAstConstructionTest, SimpleTestWithSaysAssertionProgram) {
```
bgogul, line 87: I am wondering it is worth splitting this in a separate file `says_assertion_ast_construction_test.cc`?

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -73,4 +84,62 @@ INSTANTIATE_TEST_SUITE_P(
...
+  ASSERT_EQ(prog_out.says_assertions().size(), 1);
+  const SaysAssertion& says_assertion = prog_out.says_assertions().front();
+  EXPECT_EQ(says_assertion.principal().name(), principal_name);
+  EXPECT_EQ(std::get<datalog::Predicate>(
+                std::get<BaseFact>(
+                    std::get<Fact>(says_assertion.assertions().at(0).GetValue())
+                        .GetValue())
+                    .GetValue())
+                .name(),
+            fact_assertion_name);
```
bgogul, line 100: Might be worth adding a comparator for `Predicate`? It is hard to understand what is being verified here.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -73,4 +84,62 @@ INSTANTIATE_TEST_SUITE_P(
...
+                    .at(0)
+                    .GetValue()),
+            rvalue);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    SaysAssertionAstConstructionTest, SaysAssertionAstConstructionTest,
+    testing::Values(
+        SaysAssertionTestData({R"("UserA" says { currTimeIs(2022).
+                               mayA("NotifyA") :- currTimeIs > 2021. })",
```
bgogul, line 122: Shouldn't there be a  `Rule` for this? I don't see it being constructed at all?

src/ir/auth_logic/ast_construction_test.cc
harsha-mandadi: Probably a neat way of testing would be to do end to end testing. From input text input to the parser to DLIR level (The way Andrew implemented in rust version). And have separate files for testing says_assertion, queries, relationDeclarations and various combinations of these programs.Let me know what you think about.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -33,8 +33,8 @@
 #include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/auth_logic_cc_generator/AuthLogicLexer.h"
 #include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/auth_logic_cc_generator/AuthLogicParser.h"
 
-using auth_logic_cc_generator::AuthLogicParser;
 using auth_logic_cc_generator::AuthLogicLexer;
+using auth_logic_cc_generator::AuthLogicParser;
```
markww, line 37: These `using` aliases should be in your `namespace`.

src/ir/auth_logic/ast_construction_test.cc
```diff
@@ -73,4 +84,62 @@ INSTANTIATE_TEST_SUITE_P(
...
+TEST_P(SaysAssertionAstConstructionTest, SimpleTestWithSaysAssertionProgram) {
+  const auto& [input_text, principal_name, fact_assertion_name,
+               cond_assertion_predicate, rvalue] = GetParam();
+  Program prog_out = ParseProgram(input_text);
+  ASSERT_EQ(prog_out.says_assertions().size(), 1);
+  const SaysAssertion& says_assertion = prog_out.says_assertions().front();
+  EXPECT_EQ(says_assertion.principal().name(), principal_name);
+  EXPECT_EQ(std::get<datalog::Predicate>(
+                std::get<BaseFact>(
+                    std::get<Fact>(says_assertion.assertions().at(0).GetValue())
```
markww, line 96: Nit: `at(0)` can probably be replaced with `front`.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -78,6 +77,19 @@ static BaseFact ConstructFlatFact(AuthLogicParser::FlatFactContext& context) {
   return BaseFact(ConstructPredicate(*CHECK_NOTNULL(ctx.predicate())));
 }
 
+static BaseFact ConstructRvalue(AuthLogicParser::RvalueContext& context) {
+  if (auto* ctx =
+          dynamic_cast<AuthLogicParser::FlatFactRvalueContext*>(&context)) {
+    return ConstructFlatFact(*CHECK_NOTNULL(ctx->flatFact()));
+  }
+  auto& ctx = *CHECK_NOTNULL(
+      dynamic_cast<AuthLogicParser::BinopRvalueContext*>(&context));
```
aferr, line 86: It looks like you are representing `#binoprvalue`s as predicates where the binary operator is the name of the predicate instead of extending the structure of the AST to have a separate RValue class. I think in principle this is OK, but it will make translating this into souffle more difficult. In the Rust implementation, the compiler pass for 
generating universe relations also makes use of the separate RValue class to figure out that the arguments in the comparison have the `Number` type (https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/universe_translation.rs).
I think having an explicit RValue class as part of the AST will make a few other things after the AST is built easier.

For the translation into souffle:
Eventually, when an RVALUE gets emitted as souffle code, it should look like:
 - `pred_name(arg, ..., arg)` if the RVALUE was a predicate, or
 - `arg <binop> arg` if the RVALUE was a binop

So these are different. If you keep your current implementation where a binop is represented by just changing the name of the predicate, you would probably implement the translation with an `if/else` that checks if the name of the predicate is a binary operator. I think this is undoing some of the work that should already have been done by the time the parser produces an AST object. If you have a separate RValue class you would instead implement the translation by visiting the variant in the RVALUE (or, if you make the binop/predicate subclasses of RValue you could use a virtual method instead). I think using an RValue class would result in a more idiomatic translation.

For the compiler pass that populates the universes, you can also distinguish between a binop and a predicate rvalue by checking the predicate name, but I think this is another situation where having an RValue class would make that code simpler.

aferr: I am just curious why this is resolved. Did you already discuss this option with @bgogul and/or @markww and decide this is not the way to go?

harsha-mandadi: Andrew, I thought I provided comments earlier. I didn’t realize I started a review and my comments were in pending state. I just published them. Please take a look and let me know.

aferr: OK sg. Yes, it was just that the comment was missing as you said :)

src/ir/auth_logic/ast_construction_test.cc
harsha-mandadi: Done! Separated query and says_assertion testing into 2 files with ToString functions in a common header file.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -40,71 +40,152 @@ using auth_logic_cc_generator::AuthLogicLexer;
...
+      dynamic_cast<AuthLogicParser::BinopRvalueContext*>(&rvalue_context));
+  std::vector<std::string> args = utils::MapIter<std::string>(
+      binop_rvalue_context.pred_arg(),
+      [](auto* pred_arg) { return pred_arg->getText(); });
+  return BaseFact(datalog::Predicate(binop_rvalue_context.binop()->getText(),
+                                     std::move(args),
+                                     datalog::Sign::kPositive));
 }
 
 static Fact ConstructFact(std::forward_list<Principal> delegation_chain,
```
aferr, line 110: It looks like the idea is to populate this list through the recursive calls to `ConstructFact` inside this code, and then whenever a `Fact` is made from some larger piece of syntax like `SaysAssertions` or `Queries`, you pass in an empty list. This makes sense. It might be easier to read though if you either:
 - 1)  Add a comment above this explaining that this list is populated by recursion and that when this is called from outside it should be passed an empty list
- 2) Rename this function to `ConstructFactInner(std::forward_list<Principal> delegation_chain, AuthLogicParser::FactContext& fact_context)` and make a new function with the name and signature `ConstructFact(AuthLogicParser::FactContext& fact_context)`

Or ideally, do both.

harsha-mandadi: Added comments for ConstructFact

src/ir/auth_logic/ast_construction_test.cc
harsha-mandadi: Changed test for assertion to make it simpler for verifying. PTAL

markww (approved):
LGTM

PR #523, opened 2022-04-14
Auth logic relation declaration AST datastructure

harsha-mandadi:
This PR constructs a Relation Declaration AST node. Lowering ast for relation declaration will be handled in another PR keep the PR size small.

### DIFF
--- a/src/ir/auth_logic/ast.h
+++ b/src/ir/auth_logic/ast.h
@@ -139,6 +139,56 @@ class Assertion {
   AssertionVariantType value_;
 };
 
+enum class ConstantType { kNumberType, kPrincipalType };
+
+// Relation declaration types are of 3 forms
+// 1. NumberType
+// 2. PrincipalType
+// 3. User defined CustomType (string to store name of the type)
+class ArgumentType {
+ public:
+  enum class Kind { kNumber, kPrincipal, kCustom };
+  explicit ArgumentType(Kind kind, absl::string_view name)
+      : kind_(kind), name_(name) {}
+  Kind kind() const { return kind_; }
+  absl::string_view name() const { return name_; }
+
+ private:
+  Kind kind_;
+  std::string name_;
+};
+
+class Argument {
+ public:
+  explicit Argument(std::string_view argument_name, ArgumentType argument_type)
+      : argument_name_(argument_name),
+        argument_type_(std::move(argument_type)) {}
+  absl::string_view argument_name() const { return argument_name_; }
+  ArgumentType argument_type() const { return argument_type_; }
+
+ private:
+  std::string argument_name_;
+  ArgumentType argument_type_;
+};
+
+class RelationDeclaration {
+ public:
+  explicit RelationDeclaration(absl::string_view relation_name,
+                               bool is_attribute,
+                               std::vector<Argument> arguments)
+      : relation_name_(relation_name),
+        is_attribute_(is_attribute),
+        arguments_(std::move(arguments)) {}
+  absl::string_view relation_name() const { return relation_name_; }
+  bool is_attribute() const { return is_attribute_; }
+  const std::vector<Argument>& arguments() const { return arguments_; }
+
+ private:
+  std::string relation_name_;
+  bool is_attribute_;
+  std::vector<Argument> arguments_;
+};
+
 // SaysAssertion prepends an assertion with "<principal> says"
 class SaysAssertion {
  public:
@@ -173,16 +223,25 @@ class Query {
 // This is a top-level program.
 class Program {
  public:
-  explicit Program(std::vector<SaysAssertion> says_assertions,
+  explicit Program(std::vector<RelationDeclaration> relation_declarations,
+                   std::vector<SaysAssertion> says_assertions,
                    std::vector<Query> queries)
-      : says_assertions_(std::move(says_assertions)),
+      : relation_declarations_(std::move(relation_declarations)),
+        says_assertions_(std::move(says_assertions)),
         queries_(std::move(queries)) {}
+
+  const std::vector<RelationDeclaration>& relation_declarations() const {
+    return relation_declarations_;
+  }
+
   const std::vector<SaysAssertion>& says_assertions() const {
     return says_assertions_;
   }
+
   const std::vector<Query>& queries() const { return queries_; }
 
  private:
+  std::vector<RelationDeclaration> relation_declarations_;
   std::vector<SaysAssertion> says_assertions_;
   std::vector<Query> queries_;
 };
--- a/src/ir/auth_logic/ast_construction.cc
+++ b/src/ir/auth_logic/ast_construction.cc
@@ -128,6 +128,50 @@ static Fact ConstructFact(std::forward_list<Principal> delegation_chain,
                        *CHECK_NOTNULL(can_say_fact_context.fact()));
 }
 
+static ArgumentType ConstructArgumentType(
+    AuthLogicParser::AuthLogicTypeContext& auth_logic_type_context) {
+  if (auto* number_type_context =
+          dynamic_cast<AuthLogicParser::NumberTypeContext*>(
+              &auth_logic_type_context)) {
+    return ArgumentType(ArgumentType::Kind::kNumber,
+                        number_type_context->NUMBERTYPE()->getText());
+  } else if (auto* principal_type_context =
+                 dynamic_cast<AuthLogicParser::PrincipalTypeContext*>(
+                     &auth_logic_type_context)) {
+    return ArgumentType(ArgumentType::Kind::kPrincipal,
+                        principal_type_context->PRINCIPALTYPE()->getText());
+  }
+  auto& custom_type_context =
+      *CHECK_NOTNULL(dynamic_cast<AuthLogicParser::CustomTypeContext*>(
+          &auth_logic_type_context));
+  return ArgumentType(ArgumentType::Kind::kCustom,
+                      custom_type_context.ID()->getText());
+}
+
+static RelationDeclaration ConstructRelationDeclaration(
+    AuthLogicParser::RelationDeclarationContext& relation_declaration_context) {
+  // grammar rule for relationDeclaration:
+  // '.decl' ATTRIBUTE? ID '(' ID ':' authLogicType (',' ID ':'authLogicType)*
+  // ')' When translated to Antlr generated Parse tree, ID[0] corresponds to
+  // relation_name.
+  // (ID[1] : authLogicType[0]), (ID[2] : authLogicType[1]),.....,(ID[n] :
+  // authLogicType[n-1]) corresponds to
+  //  arguments for the relation.
+  std::vector<Argument> arguments;
+  arguments.reserve(relation_declaration_context.authLogicType().size());
+  for (uint64_t i = 0; i < relation_declaration_context.authLogicType().size();
+       ++i) {
+    arguments.push_back(Argument(
+        relation_declaration_context.ID(i + 1)->getText(),
+        ConstructArgumentType(
+            *CHECK_NOTNULL(relation_declaration_context.authLogicType(i)))));
+  }
+  return RelationDeclaration(
+      relation_declaration_context.ID(0)->getText(),
+      relation_declaration_context.ATTRIBUTE() != nullptr,
+      std::move(arguments));
+}
+
 static Query ConstructQuery(AuthLogicParser::QueryContext& query_context) {
   return Query(query_context.ID()->getText(),
                ConstructPrincipal(*CHECK_NOTNULL(query_context.principal())),
@@ -182,6 +226,13 @@ static SaysAssertion ConstructSaysAssertion(
 
 static Program ConstructProgram(
     AuthLogicParser::ProgramContext& program_context) {
+  auto relation_declarations = utils::MapIter<RelationDeclaration>(
+      program_context.relationDeclaration(),
+      [](AuthLogicParser::RelationDeclarationContext*
+             relation_declaration_context) {
+        return ConstructRelationDeclaration(
+            *CHECK_NOTNULL(relation_declaration_context));
+      });
   std::vector<SaysAssertion> says_assertions = utils::MapIter<SaysAssertion>(
       program_context.saysAssertion(),
       [](AuthLogicParser::SaysAssertionContext* says_assertion_context) {
@@ -192,7 +243,8 @@ static Program ConstructProgram(
       [](AuthLogicParser::QueryContext* query_context) {
         return ConstructQuery(*CHECK_NOTNULL(query_context));
       });
-  return Program(std::move(says_assertions), std::move(queries));
+  return Program(std::move(relation_declarations), std::move(says_assertions),
+                 std::move(queries));
 }
 
 /// This function produces an abstract syntax tree (AST) rooted with a program
--- a/src/ir/auth_logic/BUILD
+++ b/src/ir/auth_logic/BUILD
@@ -153,3 +153,20 @@ cc_test(
         "@absl//absl/strings",
     ],
 )
+
+cc_test(
+    name = "ast_construction_relation_declaration_test",
+    srcs = ["ast_construction_relation_declaration_test.cc"],
+    copts = ["-fexceptions"],
+    features = ["-use_header_modules"],
+    deps = [
+        ":ast",
+        ":ast_construction",
+        ":ast_string",
+        "//src/common/testing:gtest",
+        "//src/common/utils:map_iter",
+        "//src/common/utils:overloaded",
+        "//src/ir/datalog:program",
+        "@absl//absl/strings",
+    ],
+)
--- a/src/ir/auth_logic/ast_construction_relation_declaration_test.cc
+++ b/src/ir/auth_logic/ast_construction_relation_declaration_test.cc
@@ -1,0 +1,87 @@
+//-------------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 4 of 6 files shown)


src/ir/auth_logic/ast.h
```diff
@@ -152,6 +152,43 @@ class Assertion {
...
+  const std::string& predicate_name() const { return predicate_name_; }
+  const bool& is_attribute() const { return is_attribute_; }
+  const std::vector<RelationDeclarationType>& arg_type() const {
+    return arg_type_;
+  }
+
+ private:
+  std::string predicate_name_;
+  bool is_attribute_;
+  std::vector<RelationDeclarationType> arg_type_;
```
markww, line 189: It is odd to call a field containing multiple argument types in a vector `arg_type_` as if it were singular. Perhaps this should be rewritten to `argument_types_`.

src/ir/auth_logic/ast.h
```diff
@@ -186,16 +223,22 @@ class Query {
...
  public:
-  explicit Program(std::vector<SaysAssertion> says_assertions,
+  explicit Program(std::vector<RelationDeclaration> relation_declaration,
+                   std::vector<SaysAssertion> says_assertions,
                    std::vector<Query> queries)
-      : says_assertions_(std::move(says_assertions)),
+      : relation_declaration_(std::move(relation_declaration)),
+        says_assertions_(std::move(says_assertions)),
         queries_(std::move(queries)) {}
+  const std::vector<RelationDeclaration>& relatin_declaration() const {
```
markww, line 232: typo: should be `relation_declaration`.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -33,8 +33,8 @@
 #include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/auth_logic_cc_generator/AuthLogicLexer.h"
 #include "src/policy/auth_logic/auth_logic_cc_generator_grammar.inc/auth_logic_cc_generator/AuthLogicParser.h"
 
-using auth_logic_cc_generator::AuthLogicParser;
 using auth_logic_cc_generator::AuthLogicLexer;
+using auth_logic_cc_generator::AuthLogicParser;
```
markww, line 37: These `using` declarations should be inside the `namespace`.

src/ir/auth_logic/ast.h
```diff
@@ -152,6 +152,43 @@ class Assertion {
...
+// 1. NumberType
+// 2. PrincipalType
+// 3. User defined CustomType (string to store name of the type)
+class RelationDeclarationType {
+ public:
+  using RelationDeclarationTypeVariantType =
+      std::variant<constantType, std::string>;
+  explicit RelationDeclarationType(RelationDeclarationTypeVariantType value)
+      : value_(std::move(value)) {}
+  const RelationDeclarationTypeVariantType& GetValue() const { return value_; }
```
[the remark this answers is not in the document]
bgogul: In terms of naming, the following might also work: `AuthLogicType` or `VariableType` or `ArgumentType`. We also use this for Attribute, so `RelationArgumentType` or `RelationParameterType` might not be quite accurate. 

Anyhow,. assuming `ArgumentType` as the name, you can simplify this datastructure without variants:

```C++
class ArgumentType {
  enum class Kind { kNumber, kPrincipal, kCustom};
private:
  Kind kind_;
  std::string name_;
};
```

bgogul (commented):
It looks like you don't need `variant` at all.

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
harsha-mandadi: Ah, interesting. Nested class is a good idea. Thanks Gogul. I was looking for something simpler and found my answer.

src/ir/auth_logic/ast.h
harsha-mandadi: Oops!

src/ir/auth_logic/ast_construction.cc
```diff
@@ -97,12 +97,14 @@ static Query ConstructQuery(AuthLogicParser::QueryContext& context) {
...
 static Program ConstructProgram(AuthLogicParser::ProgramContext& context) {
+  std::vector<RelationDeclaration> relation_declarations;
   std::vector<SaysAssertion> says_assertions;
   std::vector<Query> queries = utils::MapIter<Query>(
       context.query(), [](AuthLogicParser::QueryContext* query_context) {
         return ConstructQuery(*CHECK_NOTNULL(query_context));
       });
-  return Program(std::move(says_assertions), std::move(queries));
+  return Program(std::move(relation_declarations), std::move(says_assertions),
+                 std::move(queries));
```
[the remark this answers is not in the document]
harsha-mandadi: This is a temporary state and hence empty says_assertion and relation_declaration. This is a WIP PR and will add details for says_assertions and relation_declaration eventually.

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
harsha-mandadi: Question about std::visit for variant types. Our current ast.h has multiple variant types returning through getters. We should probably change them as well?

src/ir/auth_logic/ast.h
```diff
@@ -152,6 +152,43 @@ class Assertion {
   AssertionVariantType value_;
 };
 
+enum class constantType { NumberType, PrincipalType };
```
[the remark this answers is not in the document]
harsha-mandadi: Noted!

src/ir/auth_logic/ast.h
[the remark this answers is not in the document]
markww: Yes, definitely. It's bad encapsulation to return these out of the class that contains them.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,46 @@ class Assertion {
   AssertionVariantType value_;
 };
 
+enum class ConstantType { NumberType, PrincipalType };
```
bgogul, line 142: `kNumberType, kPrincipalType`

See https://google.github.io/styleguide/cppguide.html#Enumerator_Names

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,46 @@ class Assertion {
...
+
+// Relation declaration have types are of 3 forms
+// 1. NumberType
+// 2. PrincipalType
+// 3. User defined CustomType (string to store name of the type)
+class ArguementType {
```
bgogul, line 148: typo?  `ArguementType` -> `ArgumentType`

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,46 @@ class Assertion {
...
+ public:
+  enum class Kind { kNumber, kPrincipal, kCustom };
+  explicit ArguementType(Kind kind, std::string name)
+      : kind_(std::move(kind)), name_(std::move(name)) {}
+  explicit ArguementType(Kind kind) : kind_(std::move(kind)) {}
```
bgogul, line 153: Why is this needed? Don't we always want to have a name? It looks like you expect name only if kind is `CustomType`? An alternative way to do this with static factory methods and simply hide the constructor:

```C++
class ArgumentType {
public:
  // ...
  static ArgumentType MakePrincipalType() { return ArgumentType(Kind::kPrincipal, ""); }
  static ArgumentType MakeNumberType() { return ArgumentType(Kind::kNumber, ""); }
  static ArgumentType MakeCustomType(absl::string_view name) { return ArgumentType(Kind::kCustom, name); }

  ...
private:
     explicit ArgumentType(Kind kind, absl::string_view name) : kind_(std::move(kind)), name_(name) {}
}
```


```

File: src/ir/auth_logic/ast.h [hunk 10]
```diff
@@ -139,6 +139,46 @@ class Assertion {
...
+  const std::string& relation_name() const { return relation_name_; }
+  const bool& is_attribute() const { return is_attribute_; }
+  const std::vector<std::pair<std::string, ArguementType>>& arguements() const {
+    return arguements_;
+  }
+
+ private:
+  std::string relation_name_;
+  bool is_attribute_;
+  std::vector<std::pair<std::string, ArguementType>> arguements_;
```
bgogul, line 179: typo? `arguements_` -> `arguments_`.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -128,6 +128,40 @@ static Fact ConstructFact(std::forward_list<Principal> delegation_chain,
                        *CHECK_NOTNULL(can_say_fact_context.fact()));
 }
 
+static ArguementType ConstructArguementType(
```
bgogul, line 131: typo: `ArugmentType ConstructArugmentType(...`

src/ir/auth_logic/ast_construction.cc
```diff
@@ -182,6 +216,14 @@ static SaysAssertion ConstructSaysAssertion(
 
 static Program ConstructProgram(
     AuthLogicParser::ProgramContext& program_context) {
+  std::vector<RelationDeclaration> relation_declarations =
+      utils::MapIter<RelationDeclaration>(
```
bgogul, line 220: nit: `auto relation_declarations =  utils::MapIter<RelationDeclaration>(...`

bgogul (commented):
Left style-related comments...

src/ir/auth_logic/ast.h
markww: +1. This has the additional benefit of being more type-safe, in the sense that you are disallowing the bad pattern of making a `CustomType` without a name or a `NumberType` with a name.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,46 @@ class Assertion {
...
+ private:
+  Kind kind_;
+  std::string name_;
+};
+
+class RelationDeclaration {
+ public:
+  explicit RelationDeclaration(
+      std::string relation_name, bool is_attribute,
+      std::vector<std::pair<std::string, ArgumentType>> arguments)
```
markww, line 166: This use of `std::pair` is interesting; should there be a simple `Argument` class that contains within it the fields `std::string argument_name_` and `ArgumentType argument_type_`? It's usually the case that `std::pair` and `std::tuple` should only be used for locals or weird template-y situations. Usually for fields of classes and other more formal layout situations a more formal class should be used.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -128,6 +128,40 @@ static Fact ConstructFact(std::forward_list<Principal> delegation_chain,
...
+static RelationDeclaration ConstructRelationDeclaration(
+    AuthLogicParser::RelationDeclarationContext& relation_declaration_context) {
+  std::vector<std::pair<std::string, ArgumentType>> arguments;
+  for (uint64_t i = 0; i < relation_declaration_context.authLogicType().size();
+       ++i) {
+    arguments.push_back(std::pair(
+        relation_declaration_context.ID(i + 1)->getText(),
+        ConstructArgumentType(
+            *CHECK_NOTNULL(relation_declaration_context.authLogicType(i)))));
+  }
```
markww, line 158: This is pretty subtle and looks like it could use a comment explaining it. Why are we looking up `ID(i + 1)`? Is this a 1-indexed collection? It sets off my "maybe there's a bug" alarm bells to see `i+1` and `i` used to index into stuff in `relation_declaration_context` in the same loop, but it could be fine depending.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -128,6 +128,40 @@ static Fact ConstructFact(std::forward_list<Principal> delegation_chain,
...
+  auto& custom_type_context =
+      *CHECK_NOTNULL(dynamic_cast<AuthLogicParser::CustomTypeContext*>(
+          &auth_logic_type_context));
+  return ArgumentType(ArgumentType::Kind::kCustom,
+                      custom_type_context.ID()->getText());
+}
+
+static RelationDeclaration ConstructRelationDeclaration(
+    AuthLogicParser::RelationDeclarationContext& relation_declaration_context) {
+  std::vector<std::pair<std::string, ArgumentType>> arguments;
```
markww, line 151: If you know the size of a `vector` in advance (as you do here), it's a cheap and safe optimization to call `reserve` with the size that you expect the vector to be.

src/ir/auth_logic/ast.h
harsha-mandadi: I made changes to ArgumentType Node to make it look simpler. The reason I left out the name is that, for NumberType and PrincipalType, it is understood that they correspond to 'Number' and 'Principal' (Handle this in ast_construction). Whereas for customType, we need to explicitly provide the name for type. As this does not look like a good pattern to have, I changed to having name for all types. PTAL.

src/ir/auth_logic/ast.h
harsha-mandadi: Added a new node for Argument to remove std::pair

src/ir/auth_logic/ast_construction.cc
harsha-mandadi: Comments provided in the file explaining the need to `i` and `i+1`.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+enum class ConstantType { kNumberType, kPrincipalType };
+
+// Relation declaration types are of 3 forms
+// 1. NumberType
+// 2. PrincipalType
+// 3. User defined CustomType (string to store name of the type)
+class ArgumentType {
+ public:
+  enum class Kind { kNumber, kPrincipal, kCustom };
+  explicit ArgumentType(Kind kind, std::string name)
```
bgogul, line 151: No need for `std::move` for kind. Kind is already cheap to copy.

optional: Instead of taking a string, you could take an absl::string_view:
```C++
explicit ArgumentType(Kind kind, absl::string_view name): kind_(kind), name_(name) {}
...

```
Note that  the field `  std::string name_; ` would still stay as string.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+      : kind_(std::move(kind)), name_(std::move(name)) {}
+  const Kind& kind() const { return kind_; }
```
bgogul, line 153: Simply `Kind` is sufficient here.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+  const std::string name() const { return name_; }
```
bgogul, line 154: `const std::string&` here. Otherwise, you will make an unnecessary copy. You can also use `absl::string_view` above.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+
+ private:
+  Kind kind_;
+  std::string name_;
+};
+
+class Argument {
+ public:
+  explicit Argument(std::string argument_name, ArgumentType argument_type)
```
bgogul, line 163: optional: `std::string argument_name` could be `absl::string_view` as in  `ArgumentType` case.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+      : argument_name_(std::move(argument_name)),
+        argument_type_(std::move(argument_type)) {}
+  const std::string& argument_name() { return argument_name_; }
```
bgogul, line 166: Add const declarations at the end, which marks the members as const functions. Also, return `ArgumentType` instead of `const ArgumentType`
```
  const std::string& argument_name() const  { return argument_name_; } 
  ArgumentType argument_type() const { return argument_type_; }
```

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+
+class RelationDeclaration {
+ public:
+  explicit RelationDeclaration(std::string relation_name, bool is_attribute,
+                               std::vector<Argument> arguments)
+      : relation_name_(std::move(relation_name)),
+        is_attribute_(std::move(is_attribute)),
+        arguments_(std::move(arguments)) {}
+  const std::string& relation_name() const { return relation_name_; }
+  const bool& is_attribute() const { return is_attribute_; }
```
bgogul, line 182: simply `bool` instead of `const bool&`.

src/ir/auth_logic/ast.h
```diff
@@ -139,6 +139,55 @@ class Assertion {
...
+  const ArgumentType argument_type() { return argument_type_; }
+
+ private:
+  std::string argument_name_;
+  ArgumentType argument_type_;
+};
+
+class RelationDeclaration {
+ public:
+  explicit RelationDeclaration(std::string relation_name, bool is_attribute,
```
bgogul, line 176: optional: absl::string_view.

src/ir/auth_logic/ast_construction.cc
```diff
@@ -128,6 +128,50 @@ static Fact ConstructFact(std::forward_list<Principal> delegation_chain,
...
+                      custom_type_context.ID()->getText());
+}
+
+// grammar rule for relationDeclaration:
+// '.decl' ATTRIBUTE? ID '(' ID ':' authLogicType (',' ID ':'authLogicType)* ')'
+// When translated to Antlr generated Parse tree, ID[0] corresponds to
+// relation_name.
+// (ID[1] : authLogicType[0]), (ID[2] : authLogicType[1]),.....,(ID[n] :
+// authLogicType[n-1]) corresponds to
+//  arguements for the relation.
```
bgogul, line 157: nit: Move this into the body of the function?

bgogul (commented):
Noticed a few C++ no nos. Can you take care of those please?

Otherwise, looks good.

bgogul (approved):
Nice. Looks good!

PR #524, opened 2022-04-14
Make `DecodeExpressionArena` have a no-context overload.

markww:
`DecodeExpressionArena`, up til now, has taken a `DecoderContext` as one
of its arguments. You could imagine this being very useful for fancy
scenarios, such as having multiple pieces of a large program containing
a SQL query relying on a common context. However, it is a bit annoying
for the simple case we would like to solve right now, in which we want
to decode the IR for a single SQL query for the SQL verifier
integration. For this case, it is easier to not pass a context, which
should cause a fresh context to be constructed and used for decoding.
This commit adds an overload with that behavior.

### DIFF
--- a/src/frontends/sql/decode.cc
+++ b/src/frontends/sql/decode.cc
@@ -110,8 +110,8 @@ static Value DecodeExpression(const Expression &expr,
 
 // Decode every `Expression` contained within the arena and return the top
 // level `ir::Value` for that arena.
-Value DecodeExpressionArena(
-    const ExpressionArena &expr_arena, DecoderContext &decoder_context) {
+Value DecodeExpressionArena(const ExpressionArena &expr_arena,
+                            DecoderContext &decoder_context) {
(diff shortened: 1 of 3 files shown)


src/frontends/sql/decode.cc
```diff
@@ -123,4 +124,10 @@ Value DecodeExpressionArena(const ExpressionArena &expr_arena,
   return *top_level_value;
 }
 
+Value DecodeExpressionArena(const ExpressionArena &expr_arena) {
+  ir::IRContext ir_context;
+  DecoderContext decoder_context(ir_context);
+  return DecodeExpressionArena(expr_arena, decoder_context);
```
bgogul, line 130: Won't all the operators in `Value` have dangling references when DecoderContext is destroyed?

markww:
Well, that was a silly mistake on my part, thank you for pointing that out @bgogul. Closing.

PR #529, opened 2022-04-19
Add C++ types to express Souffle facts using C++ Souffle value types.

markww:
Adds a `Fact` class that allows expressing a Souffle fact as a C++ type,
with the layout expressed in terms of the C++ Souffle value types
provided in an earlier PR. Using these Souffle value types in templates
to describe the layout allows the C++ compiler to check our work for us.
Provides methods for printing `Fact`s out as datalog.


src/backends/policy_enforcement/souffle/fact.h
```diff
@@ -0,0 +1,47 @@
...
+#define SRC_BACKENDS_POLICY_ENFORCEMENT_SOUFFLE_FACT_H_
+
+#include <utility>
+
+#include "src/backends/policy_enforcement/souffle/value.h"
+
+namespace raksha::backends::policy_enforcement::souffle {
+
+template <const char *kRelationName, class... RelationParameterTypes>
+class Fact {
```
bgogul, line 27: We should reconcile with `src/ir/datalog`. I don't want us to get too fragmented.

bgogul (commented):
Before I take a closer look, can you take a look at my comment about moving this to src/ir/datalog?

src/backends/policy_enforcement/souffle/fact.h
markww: Done.

src/ir/datalog/input_relation_fact.h
```diff
@@ -0,0 +1,54 @@
...
+#include "absl/strings/str_join.h"
+#include "src/ir/datalog/value.h"
+
+namespace raksha::ir::datalog {
+
+// A fact that is built to be fed into an input relation for Souffle.
+// Although the `Rule` class can also express a `Fact` by having no RHS, this
+// is explicitly to prepare input to a Souffle program. You can think of the
+// difference between this `Fact` and a `Rule` fact as a difference between
+// data and code.
```
bgogul, line 32: If this is data, shouldn't this be a `datalog::Value`? It looks like a record might be sufficient for what you are trying to do?

I don't clearly understand  what is the advantage of this over `Rule`. e.g., `ToDatalogString` generates the same representation as a `Rule`? What do we gain by having a separate class?

bgogul: Is the difference that the arguments should be constant terms?

markww: Yes, sort of. For an `InputRelationFact`, we have strong ideas about what kind of static type each of the arguments to that `InputRelationFact` should be. `InputRelationFact` therefore keeps around static type information via templates, which allows the C++ type checker to catch us if we try to do something silly like get our `AttributeList` and `OperationList` backwards. Using full types instead of string info makes certain classes of errors just impossible, such as missing a closing `]` in a relation definition.

Because the authorization logic translator has to deal with custom relations that a user may introduce, it must have a dynamic representation of even `Rule`s that happen to be facts. You could imagine it augmenting its AST with these static type annotations for "core" relations used by the authorization logic though, so it does make sense to place these classes here in hopes of future sharing. But the Souffle backend for the dataflow graph IR cannot use the existing `Rule` class without losing a great deal of valuable type information.

bgogul: I think this is more closely related to `Predicate` than `Rule`. We might want to reconcile these two at some point. I like the type safety, but I expect this to add some friction due to lack of flexibility. Let us run with this see how things go. 

I might be OK calling this simply a `Fact` as well.

src/ir/datalog/input_relation_fact.h
```diff
@@ -0,0 +1,56 @@
...
+template <class... RelationParameterTypes>
+class InputRelationFact {
+ public:
+  InputRelationFact(RelationParameterTypes &&...args)
+      : relation_arguments_(std::forward<RelationParameterTypes>(args)...) {}
+
+  virtual absl::string_view GetRelationName() const = 0;
```
bgogul, line 39: Add a string template argument for the relation name and replace this virtual function with a regular one? That will eliminate a lot of boiler place you see in the tests and can define classes more directly:

```C++
using IsAccessPathFact = InputRelationFact<"IsAccessPath", Symbol>;
```

bgogul (approved):
Approving this, but do consider the comments in the review.

src/ir/datalog/input_relation_fact.h
markww: The approach you suggest is the one I tried first. But I ran into subtle problems. The issue is that only pointers to objects without linkage are allowed as template arguments. That means that, for a derived class with a superclass with a string template argument defined in a header file, the pointer and thus the superclass will be different for every TU that the header gets linked into. That problem is detected by the `-Wsubobject-linkage` warning in GCC and prevented me from getting into very confusing territory.

I think that the way forward for this (and some of the other patterns) may actually be to create macros to hide the low-entropy parts of the definition. You could imagine one for defining `InputRelationFact`s like this:

```C++
#define INPUT_RELATION_FACT_CLASS_DECL(factRelationName) \
  class InputRelationFact_ ## factRelationName
  
#define INPUT_RELATION_FACT_DEFN(factRelationName) \
  INPUT_RELATION_FACT_CLASS_DECL(factRelationName) : public InputRelationFact { \
    using InputRelationFact::InputRelationFact;
    static absl::string_view relation_name() { return #factRelationName ; } \
    absl::string_view GetRelationName() const override { return relation_name(); } \
  }
```

Note that by using the relation name to both generate the class name and produce the string output by `GetRelationName`, we prevent two classes linked into the same translation unit from having different class names but being for the same relation.

src/ir/datalog/input_relation_fact.h
markww: Yes, I see that. Ideally, I'd like `InputRelationFact` to be a static-typed wrapper around the less typed `Predicate`. Unfortunately, `Predicate` is currently too unstructured to be used this way, containing only a `vector` of `std::string`s. If we worked to integrate them, I'd like the arguments of `Predicate` to be represented as a `vector` of `Value`s similar to the ones I've created (or similar type-erased objects).

PR #530, opened 2022-04-19
Add a representation of the `IsOperation` fact in C++.

markww:
Add a C++ representation of the Datalog fact `IsOperation`, used by
Raksha to get the list of `Operation`s that the analysis should concern
itself with. Due to the generic Datalog types already created, creating
this representation requires no new executable code, just new type
definitions. Created some tests showing some plausible SQL types that we
wish to represent being expressed in this system.


src/ir/datalog/operation_test.cc
```diff
@@ -0,0 +1,66 @@
...
+
+class IsOperationFactTest
+    : public TestWithParam<IsOperationFactAndExpectedDatalog> {};
+
+TEST_P(IsOperationFactTest, IsOperationFactTest) {
+  auto [is_operation_fact, expected_datalog] = GetParam();
+  EXPECT_EQ(is_operation_fact->ToDatalogString(), expected_datalog);
+}
+
+static const IsOperationFact kSampleFact1 = IsOperationFact(Operation(
```
bgogul, line 43: Will using `intrusive_ptr` enable us  to create these values directly in `INSTANTIATE_TEST_SUITE_P`? (If so, OK to defer for later.)

src/ir/datalog/attribute.h
```diff
@@ -0,0 +1,61 @@
...
+#ifndef SRC_IR_DATALOG_ATTRIBUTE_H_
+#define SRC_IR_DATALOG_ATTRIBUTE_H_
+
+#include <memory>
+
+#include "src/ir/datalog/value.h"
+
+namespace raksha::ir::datalog {
+
+constexpr char kStringAttributePayloadName[] = "StringAttributePayload";
```
bgogul, line 26: `inline constexpr` here and below. (https://abseil.io/tips/140 for context)

src/ir/datalog/BUILD
```diff
@@ -58,6 +58,21 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "raksha",
```
bgogul, line 62: Use a more specific name?

bgogul (approved):

src/ir/datalog/operation_test.cc
markww: We'd have to use `intrusive_ptr` throughout the Souffle IR data structures, but that would allow us to use them directly, yes.

markww: It would be pretty nice to be able to copy these files, but using a more exotic, bespoke data structure (like `intrusive_ptr`) may affect our ability to export this code to other projects (as we briefly discussed). Deferring for now.

src/ir/datalog/BUILD
markww: Done, now is `raksha_relation_interface`, indicating that this library contains the data structures representing `.input`/`.output` relations that describe the interface to the Raksha analysis.

src/ir/datalog/attribute.h
markww: Done, thanks for the tip.

src/ir/datalog/attribute.h
```diff
@@ -0,0 +1,64 @@
...
+
+inline constexpr absl::string_view kStringAttributePayloadName =
+    "StringAttributePayload";
+inline constexpr absl::string_view kNumberAttributePayloadName =
+    "NumberAttributePayload";
+
+class AttributePayload : public Adt {
+ public:
+  using Adt::Adt;
+};
```
bgogul, line 35: `using AttributePayload = Adt` ?

bgogul (approved):

src/ir/datalog/attribute.h
markww: That would work, but I would lose a type constraint that way: anytime I annotate a `Record` or `Adt`'s type parameters as wanting an `AttributePayload`, I want only ADT branches that belong to the `AttributePayload` ADT. By making `AttributePayload` a subclass of `Adt`, the type system prevents ADT branches belong to other ADTs from being passed in. If I were to make `AttributePayload` a `using` alias of `Adt`, the type system would not protect me in that way.

PR #531, opened 2022-04-21
Interfaces and plumbing for the policy engine using Souffle.

bgogul:


src/backends/policy_engine/souffle/sql_policy.dl
```diff
@@ -0,0 +1,49 @@
...
+//
+
+// Default owner associated with all tags related to SQL queries. 
+#define SQL_OWNER "sql"
+
+// A confidentiality tag used to capture that there restrictions.
+#define RESTRICTED_TAG "restricted"
+
+// Add restrictions on the access path.
+.decl restricted(result: AccessPath, tag: IntegrityTag)
```
markww, line 33: Another thing about `restricted`: unlike in the verifier, we will not have columns marked as implicitly having a tag. Rather, accesses to columns indicated like that in the policy will be wrapped in a `TagTransform` with no preconditions. So we do need to query for the `TagTransform` operations to find the initial tag state.

You can imagine it being something like this:

```
hasAppliedIntegrityTag(result, SQL_OWNER, tag) :- tagTransformResult(tranformNode, result), hasPolicyName(transformNode, "THIS_POLICY_NAME"), tagTransformAddsTag(transformNode, tag).
```

Where the above relations are populated from the attributes on the tag transform operation produced by the IR translator.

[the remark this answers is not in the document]
bgogul: This replicates whatever `restricted` means is in the current implementation, which is the following:

> When the SQL verifier annotates something with a restricted integrity tag, we consider them to annotate that value with both a confidentiality tag and an integrity tag. This is important because an operation may mangle the IntegrityTag away while still having a result with all of its secrecy.

See the inference based on `restricted` below. The name could be changed, but just wanted to start with what already exists.

src/backends/policy_engine/souffle/sql_policy.dl
```diff
@@ -0,0 +1,49 @@
...
+
+// Mark an access path as an output of a SQL query.
+.decl isSqlOutput(path: AccessPath)
+
+// True if there is a violation on this access path.
+.decl violatesSqlPolicy(path: AccessPath)
+.output violatesSqlPolicy
+
+violatesSqlPolicy(path) :- 
+  isSqlOutput(path), mayHaveTag(path, SQL_OWNER, RESTRICTED_TAG).
```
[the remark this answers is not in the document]
bgogul: This is mimicking what the current behavior is.

src/backends/policy_engine/souffle/sql_policy.dl
bgogul: Using `tagTransform` is how I started with, but wasn't sure why we would need it at all. IIUC, `restricted` is added to columns of _known_ tables. If we generate the same access path from `policy` and during the translation, I am not sure why we need the `tagTransform` operator at all.

markww: The reason why I'm using `TagTransform` for even statically-known columns is because I think it's against the spirit of the IR abstraction to have knowledge of how `AccessPath`s are generated from column names present in the frontend. But perhaps it would be a way to sidestep dealing with `TagTransform`s for an initial demo. It would, however, be adding logic for one special case whereas handling the `TagTransform` node would allow linking to policy rules for both columns and inner nodes of the query.

bgogul:
@markww please take a look at this when you get a chance. This just puts together the necessary scaffolding for the SQL integration.

src/backends/policy_engine/souffle/sql_policy.dl
```diff
@@ -0,0 +1,49 @@
...
+
+#include "src/analysis/souffle/taint.dl"
+
+// OK: SELECT 42, 'foobar', null, true
+// OK: SELECT gender FROM people
+// FAIL: SELECT SSN FROM people
+//
+
+// Default owner associated with all tags related to SQL queries. 
+#define SQL_OWNER "sql"
```
markww, line 27: Currently, I'm defining the owning `Principal` as `"UNKNOWN"` because I don't have the context to know I'm in a SQL context in the lowering code: https://github.com/google-research/raksha/pull/534/files#diff-d8cf2a91dae5d2e5a25812afae3b80891cd31b3d5fab5806d169ca539b12fb7dR49

src/backends/policy_engine/souffle/sql_policy.dl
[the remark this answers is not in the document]
markww: I'm not sure if that's the case. The SQL verifier allows arbitrary confidentiality tags and can independently clear those tags by name in policies. `restricted` is a bit of syntactic sugar for if you do not care about the name of the confidentiality tag. It's true that we only need a single confidentiality tag for a first demo, but to capture the full behavior of the SQL verifier, we need to allow it to provide arbitrarily-named confidentiality tags.

markww (commented):
Left some comments. If you are arguing that all we need for a first demo is to show a single confidentiality tag, `"restricted"`, and that we can connect a policy governing that tag on a source column via simple `AccessPath` name correspondence using the obvious algorithm of naming the `AccessPath` after the fully-qualified table name, and it's worth a bit of technical debt to get to that first demo, then I think this looks good. If you are arguing that `restricted` is all you need or that `AccessPath` name correspondence is a good way to link policies on source tables moving forward, I disagree and I think we need to discuss more.

src/backends/policy_engine/souffle/sql_policy.dl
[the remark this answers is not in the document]
bgogul: As far as I understand, `Restricted` allows only configuring with a new integrity label. The internal verifier only adds a default confidentiality label. I tried to replicate it here. 

BTW, none of this is set in stone. I am just adding enough to get us over the hump for the first few examples.  We will know more when we start integrating on an end-to-end example.

src/backends/policy_engine/souffle/sql_policy.dl
bgogul: Either works. I will leave it as "sql" just to help debugging easier? Also, this is a macro that could be changed?

bgogul:
> Left some comments. If you are arguing that all we need for a first demo is to show a single confidentiality tag, `"restricted"`, and that we can connect a policy governing that tag on a source column via simple `AccessPath` name correspondence using the obvious algorithm of naming the `AccessPath` after the fully-qualified table name, and it's worth a bit of technical debt to get to that first demo, then I think this looks good. If you are arguing that `restricted` is all you need or that `AccessPath` name correspondence is a good way to link policies on source tables moving forward, I disagree and I think we need to discuss more.

I am not suggesting that `restricted` is the only confidentiality tag that is needed. As I mentioned elsewhere,  I am just adding enough to get us over the hump for the first few examples. As we get working on the E2E, this may need to be changed.

markww (commented):
LGTM, this overall looks good, the details I'm quibbling over can be iterated on.

bgogul:
Can you approve this PR if you are OK getting this in, @markww ?

markww (approved):
I could have sworn I had selected "Approve" before. Approved.

PR #534, opened 2022-04-21
Add visitor lowering IR into Raksha's version of Souffle C++ objects.

markww:
This PR adds a visitor which lowers Raksha IR into Raksha's objects that mimic the Souffle type system. Note that this does not, by itself, make the IR translator ready for Souffle's analysis. The resulting objects, however, are very analogous to the objects manipulated by Souffle, and so it is relatively easy to turn these representations into strings or perform other future methods of communicating with the Souffle analysis.


src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
...
+        [](std::string *out, const ir::datalog::IsOperationFact &arg) {
+          absl::StrAppend(out, arg.ToDatalogString());
+        });
+  }
+
+ private:
+  std::vector<ir::datalog::IsOperationFact> is_operation_facts_;
+};
+
+ class DatalogLoweringVisitor : public ir::IRTraversingVisitor<DatalogLoweringVisitor> {
```
bgogul, line 36: Move function bodies to `.cc` file?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
+#ifndef SRC_BACKENDS_POLICY_ENGINE_SOUFFLE_DATALOG_LOWERING_VISITOR_H_
```
bgogul, line 1: copyright?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
...
+#include "src/ir/ir_visitor.h"
+#include "src/ir/module.h"
+#include "src/ir/operator.h"
+#include "src/ir/ir_traversing_visitor.h"
+
+namespace raksha::backends::policy_engine::souffle {
+
+// A class containing the Datalog facts produced by the IR translator in
+// structured (ie, C++ objects, not strings) form.
+class RakshaDatalogFacts {
```
bgogul, line 18: separate file?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
...
+ public:
+  // TODO: Dedup this from `DecoderContext` in the SQL frontend.
+  static constexpr absl::string_view kDefaultOutputName = "out";
+  DatalogLoweringVisitor(ir::SsaNames &ssa_names) : ssa_names_(ssa_names) {}
+  virtual ~DatalogLoweringVisitor() {}
+
+  void PreVisit(const ir::Operation &operation) override {
+    // We currently don't have any owner information when outputting IR. We
+    // don't need it yet, really, but we do need to output something.
+    constexpr absl::string_view kUnknownPrincipal = "UNKNOWN";
```
bgogul, line 46: Define this to be class level constant? Use `"sql"` instead of `"UNKNOWN"` for now?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
...
+    datalog_facts_.AddIsOperationFact(
+        ir::datalog::IsOperationFact(std::move(datalog_operation)));
+  }
+
+  const RakshaDatalogFacts &datalog_facts() { return datalog_facts_; }
+
+ private:
+  ir::datalog::AttributePayload GetPayloadForAttribute(
+      ir::Attribute attr) const {
+    if (intrusive_ptr<const ir::Int64Attribute> int_attr =
```
bgogul, line 74: `auto int_attr = ...` here and below?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
...
+  void PreVisit(const ir::Operation &operation) override {
+    // We currently don't have any owner information when outputting IR. We
+    // don't need it yet, really, but we do need to output something.
+    constexpr absl::string_view kUnknownPrincipal = "UNKNOWN";
+    const ir::Operator &op = operation.op();
+    absl::string_view op_name = op.name();
+
+    const ir::ValueList &ir_operand_list = operation.inputs();
+    ir::datalog::OperandList operand_list =
+        RangeToDatalogList<ir::datalog::OperandList>(ir_operand_list.begin(),
```
bgogul, line 52: You could use `absl::c_accumulate` instead of creating a new method? I think that will be more readable?
(https://en.cppreference.com/w/cpp/algorithm/accumulate)

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,118 @@
...
+                                                     ir_operand_list.end());
+    const ir::NamedAttributeMap &ir_attr_map = operation.attributes();
+    ir::datalog::AttributeList attribute_list =
+        RangeToDatalogList<ir::datalog::AttributeList>(ir_attr_map.begin(),
+                                                       ir_attr_map.end());
+
+    ir::datalog::Operation datalog_operation(
```
bgogul, line 59: Why is this not `DatalogOperation`?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+class IrOperationLoweringTest
+    : public TestWithParam<IrAndDatalogOperationPairs> {};
+
+// Check that the IrOperation is lowered to the expected `IsOperationFact`.
+// Rather than introducing a `operator==` for this, we just rely on them
+// producing the same `ToDatalogString` result.
+TEST_P(IrOperationLoweringTest, IrOperationLoweringTest) {
+  const auto [ir_operation, datalog_is_operation] = GetParam();
+
+  ir::SsaNames ssa_names;
```
bgogul, line 45: Isn't this defined in the visitor?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+                   kMergeOpOperator, {}, {ir::Value(ir::value::Any())});
+               block_builder.AddOperation(
+                   kMergeOpOperator, {},
+                   {ir::Value(ir::value::OperationResult(operation, "out"))});
+               ir::Module module;
+               module.AddBlock(block_builder.build());
+               return module;
+             },
+         .facts_fn =
+             []() {
```
bgogul, line 154: Similarly, can we just have a list of `std::vector<InputRelationFact>`instead of lambdas?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+                       ir::datalog::Symbol("sql.MergeOp"),
+                       ir::datalog::Symbol("%0.out"),
+                       ir::datalog::OperandList(ir::datalog::Symbol("<<ANY>>"),
+                                                ir::datalog::OperandList()),
+                       ir::datalog::AttributeList())));
+               return facts;
+             }},
+        // Show two passthroughs chained together.
+        {.module_fn =
+             []() {
```
bgogul, line 142: Instead of lambdas, can we just have a list of `std::vector<Operation>` and build the module in the body of the test?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: `c_accumulate` would definitely be more readable, I hadn't considered that. The one issue is that the documentation for `std::accumulate` indicates that the accumulate operation should have no side effects. If `c_accumulate` has a similar constraint, we can't actually use it, as we alter `ssa_names_` in one of the called functions. I'll check it out.

markww: It looks like `c_accumulate` is just a wrapper around `std::accumulate`, so I don't think I can use it (until C++20).

bgogul: I think it is available before C++20. It is just that some API is changing.

bgogul: Ah, you are saying that you can't use because of side effects?

markww: Right. The guidance from [C++11, C++20) is that the `BinaryOperator` used within `std::accumulate` cannot have side effects. That's unusually strong guidance about side effects for C++ and I am a bit anxious about crossing that specified line.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: No, the visitor takes an `ssa_names` as a parameter. It might not matter for now, but I guess I was thinking that if we had multiple TUs or language frontends that wished to do lowering, it might make sense to use the same `SsaNames` for all of them. I can make it a member if you think that's not worth the extra hassle at the moment.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: My thinking was that there's already a `datalog` in the namespace. Check out how I've updated it with `using` aliases, see what you think.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: That wouldn't be enough to solve our problem. The main issue here is that test parameter values must be copyable. `IsOperation`, the main `InputRelationFact` I am currently concerned with, has `Record` members, which contain `std::unique_ptr`s, which are not copyable.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: I think my initial plan was to put a test exercising `BlockArgument`s, but I realize now I never got around to writing that test.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
```diff
@@ -0,0 +1,103 @@
...
+                                                      ir::SsaNames &ssa_names) {
+  if (auto int_attr = attr.GetIf<ir::Int64Attribute>()) {
+    return DatalogNumberAttributePayload(DatalogNumber(int_attr->value()));
+  } else if (auto string_attr = attr.GetIf<ir::StringAttribute>()) {
+    return DatalogStringAttributePayload(DatalogSymbol(string_attr->value()));
+  } else {
+    LOG(FATAL) << "Unknown attribute kind.";
+  }
+  // Unreachable, just to placate compiler.
+  return DatalogStringAttributePayload(DatalogSymbol(""));
```
bgogul, line 51:
```suggestion
  }
  LOG(FATAL) << "Unknown attribute kind.";
  // Unreachable, just to placate compiler.
  return DatalogStringAttributePayload(DatalogSymbol(""));
```

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
```diff
@@ -0,0 +1,103 @@
...
+// list.
+template <class ListT, class IrNodeIter>
+static ListT RangeToDatalogList(IrNodeIter ir_node_list_begin,
+                                IrNodeIter ir_node_list_end,
+                                ir::SsaNames &ssa_names) {
+  if (ir_node_list_begin == ir_node_list_end) return ListT();
+  auto head_in_dl = TranslateIrNodeToDatalog(*ir_node_list_begin, ssa_names);
+  return ListT(std::move(head_in_dl),
+               RangeToDatalogList<ListT>(++ir_node_list_begin, ir_node_list_end,
+                                         ssa_names));
```
bgogul, line 77: Use a loop instead of recursion?

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
```diff
@@ -0,0 +1,103 @@
...
+}
+
+void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
+  // We currently don't have any owner information when outputting IR. We
+  // don't need it yet, really, but we do need to output something.
```
bgogul, line 82: Comment at the wrong place?

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
```diff
@@ -0,0 +1,103 @@
...
+                                              ir::SsaNames &ssa_names) {
+  return DatalogSymbol(value.ToString(ssa_names));
+}
+
+// Turn a range of IR nodes in a C++ container into a Datalog-style linked
+// list.
+template <class ListT, class IrNodeIter>
+static ListT RangeToDatalogList(IrNodeIter ir_node_list_begin,
+                                IrNodeIter ir_node_list_end,
+                                ir::SsaNames &ssa_names) {
```
bgogul, line 72: Instead of passing in a range, the following provides a more flexible API: 
```suggestion
template <class ListT, class T>
static ListT IrNodeToDatalogList(T ir_nodes, ir::SsaNames &ssa_names) {
  
```

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,52 @@
...
+#include "src/ir/ir_traversing_visitor.h"
+#include "src/ir/module.h"
+#include "src/ir/operator.h"
+
+namespace raksha::backends::policy_engine::souffle {
+
+class DatalogLoweringVisitor
+    : public ir::IRTraversingVisitor<DatalogLoweringVisitor> {
+ public:
+  // TODO: Dedup this from `DecoderContext` in the SQL frontend.
```
bgogul, line 35: Why can't you use the constant defined in DecoderContext itself?

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -0,0 +1,52 @@
...
+  static constexpr absl::string_view kDefaultOutputName = "out";
+  static constexpr absl::string_view kDefaultPrincipal = "sql";
+  DatalogLoweringVisitor(ir::SsaNames &ssa_names) : ssa_names_(ssa_names) {}
+  virtual ~DatalogLoweringVisitor() {}
```
bgogul, line 39: redundant?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: I see your point, but let us make it a member for now? It would be good to avoid unnecessary global state between calls to visitor.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+  const auto [ir_operation, datalog_is_operation] = GetParam();
+
+  ir::SsaNames ssa_names;
+  DatalogLoweringVisitor datalog_lowering_visitor(ssa_names);
+  ir_operation->Accept(datalog_lowering_visitor);
+  const RakshaDatalogFacts &facts = datalog_lowering_visitor.datalog_facts();
+  EXPECT_EQ(facts.ToDatalogString(), datalog_is_operation->ToDatalogString());
+}
+
+static const ir::Operator kLiteralOperator("sql.ReadLiteral");
```
bgogul, line 52: Didn't we decide that we don't need `sql.ReadLiteral`? We can simply make it an access path?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+
+static const ir::Operation kStringLiteralIrOperation(
+    nullptr, kLiteralOperator,
+    {{"literal", ir::Attribute::Create<ir::StringAttribute>("number_5")}}, {});
+
+static const ir::datalog::IsOperationFact kStringLiteralDatalogIsOperation(
+    ir::datalog::Operation(
+        ir::datalog::Symbol("UNKNOWN"), ir::datalog::Symbol("sql.ReadLiteral"),
```
bgogul, line 60: Not using `sql` for principal?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+        ir::datalog::Symbol("%0.out"), ir::datalog::OperandList(),
+        ir::datalog::AttributeList(
+            ir::datalog::Attribute(ir::datalog::Symbol("literal"),
+                                   ir::datalog::StringAttributePayload(
+                                       ir::datalog::Symbol("number_5"))),
+            ir::datalog::AttributeList())));
+
+static const ir::Operator kMergeOpOperator("sql.MergeOp");
```
bgogul, line 68: Why are we not using the factory method in https://github.com/google-research/raksha/blob/main/src/frontends/sql/ops/merge_op.h?

(Given that tests also act as documentation, it is better to use the right API here.)

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: Using vector is more readable for the test?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
```diff
@@ -0,0 +1,243 @@
...
+               ir::Module module;
+               module.AddBlock(block_builder.build());
+               return module;
+             },
+         .facts_fn =
+             []() {
+               RakshaDatalogFacts facts;
+               facts.AddIsOperationFact(
+                   ir::datalog::IsOperationFact(ir::datalog::Operation(
+                       ir::datalog::Symbol("UNKNOWN"),
```
bgogul, line 132: Still using "UNKNOWN" instead of "sql"?

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: Perhaps, it is time to change `unique_ptr` to `intrusive_ptr`? (Not in this PR, but as a follow up.) Nevertheless, is there a way to make this test more readable?

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
markww: I'm using the recursion because the Datalog `ListT` will be constructed by `cons`ing a new head over and over. This is easy to expression using recursion, but awkward to express with a loop.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: Whoops, fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: Done.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Done.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: We do need `ReadLiteral`. What we decided that we do not need (at least for now) is an `Operation` describing source columns.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: The factory method requires an `IRContext` that has been pre-populated with the appropriate `Operator`s. The interface to `IRContext` makes it hard to populate that in a `const` global. So I'm just creating some `Operator`s on the fly, as that is more concise.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
markww: I think this setup would be hard to use with recursive traversal of the list, which is the most natural way to do this because of the structure of a Datalog list type.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Removed.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: I toyed with this for a while. I think it's possible, but I'm not sure that it's a good idea.

Semantically, we want `std::unique_ptr`. These Souffle mirror structures are the owners of their child objects and they should be the only objects holding a strong pointer to those objects. The friction that we are feeling around the use of these objects is entirely the result of non-copyable objects just being tricky to work with.

It is also notable that this friction isn't really a problem in our production code, only in our test code. While I am quite willing to make changes to production code to make it more testable, it feels wrong to make the ownership model more complex (ie, moving from one owner to many owners) just to make tests easier to write. I feel that we should continue to use `std::unique_ptr` for these Souffle mirror objects.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: I can't use a `std::vector<Operation>` because `Operation` has a `std::unique_ptr` field in `impl_module_`, which prevents it from being copyable and thus prevents it from being used directly as test data parameters.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: +1, we should not change production code just for convenience of tests.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: Why can't we just define all of this in a class and use it? I am mainly pushing back because test act as documentation and don't want to leave the wrong usage here. 

See https://github.com/google-research/raksha/blob/eee95e5c52f186b44674f0379bc1c1e397ad29ab/src/frontends/sql/ops/merge_op_test.cc#L37 and https://github.com/google-research/raksha/blob/eee95e5c52f186b44674f0379bc1c1e397ad29ab/src/frontends/sql/ops/tag_transform_op_test.cc#L35

We should probably create a test util class, but we can defer that for later.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: In that case create a some sort of helper to avoid repeating block builder setup ?

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
bgogul: Sounds good. I will defer it to you. BTW, I think it is ok to use std::accumulate as the side-effect restriction is only until C++11. (I think we discussed this on another review thread.)

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: I must have misread this; `std::accumulate` allows side effects from C++11 on. That means I can use it and it might significantly reduce the code needed here.

markww: OK, there is another very annoying reason that `std::accumulate` won't work here. As of C++20, `std::accumulate` uses `std::move` to move the accumulated value from one "iteration" of the fold to another. Before that, it uses a copy. Because `OperandList` and `AttributeList` both contain `std::unique_ptr`s, we can't use `std::accumulate` for this (until C++20).

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
markww: You are correct, I misread the description in the docs earlier, and `std::accumulate` can handle side effects as of C++11. Unfortunately, `std::accumulate` copies its accumulated value until C++20, so we can't use it here.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Added a `fold` in #539 . Depending on the PR's respective timelines, will rewrite in terms of `fold` in this or a followup PR.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: Added a helper to reduce boilerplate.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: So, the issue is that these are used in constructing data parameter objects, which must be fully formed before any test or test fixture code runs. Currently, we have to fill an `IRContext` with the structures that it tracks via `RegisterOperator` and/or `RegisterStorage` function calls from test or test fixture code.

It doesn't necessarily have to be that way. It is easy to imagine adding a constructor onto `IRContext` taking `std::vector`s of `Operator`s and `Storage`s and statically seeding the context with those objects. This would allow a `const` global `IRContext` to be created statically, which would make it possible to construct the `IRContext` before any of these `Operation`s and would solve this problem. But that requires some not hard but not trivial refactoring in `IRContext`.

So I think this is, unfortunately, the best we can do for now, but I think we can do better with an additional constructor in `IRContext`.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
markww: This could also be made more natural by integrating #539 .

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
markww: See #540, which contains the change that I think would be necessary to make using `IRContext` in this fashion easier.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
markww: Fixed.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
markww: Fixed with use of `fold`.

src/backends/policy_engine/souffle/datalog_lowering_visitor_test.cc
bgogul: Actually, I just realized that we should be testing for operations in general (and not for sql operators). As long as we test operators that have  inputs, attributes, and outputs, we are fine. .

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
bgogul: I was saying that `datalog_operation` does not match C++ style.

bgogul (approved):

PR #537, opened 2022-04-29
Move `xform_to_datalog` and arcs code in `src/ir` to `frontends/arcs`

bgogul:
This move is done to unblock the following tasks:
  - Get the old datalog path out of the way to avoid confusion with the new datalog lowering.
  - Cleaning up the ir directory so that we can document it. 
  - Migrate arcs to the new ir.

The data structures related to `AccessPathSelectors` had to be left in `src/ir` because it is still used in types. Once we implement a visitor for types, we can move this out as well.


src/frontends/arcs/handle_connection_spec_test.cc
```diff
@@ -120,25 +118,20 @@ TEST_P(GetAccessPathTest, GetAccessPathTest) {
 }
 
 // Just some different strings that could be used as particle specs.
-std::string sample_particle_spec_names[] = {
-    "ParticleSpec",
-    "PS",
-    "P1",
-    "P2"
-};
+std::string sample_particle_spec_names[] = {"ParticleSpec", "PS", "P1", "P2"};
```
markww, line 121: Since we're touching the code, this should probably be turned into `static constexpr absl::string_view sample_particle_spec_names[]`.

src/frontends/arcs/handle_connection_spec_test.cc
```diff
@@ -120,25 +118,20 @@ TEST_P(GetAccessPathTest, GetAccessPathTest) {
...
 
 ProtoStringAndExpectedAccessPathPieces protos_and_access_path_info[] = {
-    { .textproto =
-        R"(name: "handle_spec" direction: READS type: { primitive: TEXT })",
-      .handle_connection_name = "handle_spec",
-      .access_path_selectors_tostrings = { "" } },
-    { .textproto = R"(
+    {.textproto =
+         R"(name: "handle_spec" direction: READS type: { primitive: TEXT })",
```
markww, line 125: Also this global should be made `static const`.

src/frontends/arcs/particle_spec_test.cc
```diff
@@ -33,46 +32,42 @@ struct ParticleSpecProtoAndExpectedInfo {
...
 
-  types::TypeFactory type_factory_;
+  ir::types::TypeFactory type_factory_;
   std::unique_ptr<ParticleSpec> particle_spec_;
 };
 
 // Locally defined operator== for use in test below.
-bool operator ==(const ir::TagCheck& check1,
-                 const ir::TagCheck* check2) {
+bool operator==(const TagCheck& check1, const TagCheck* check2) {
```
markww, line 56: Perhaps this should be `static` to prevent it being accidentally used elsewhere?

markww (approved):
LGTM. I found some issues, but they existed before your move and aren't causing us problems currently, so I leave it to you to decide whether to address them here, in a follow-up PR, or just ignore them until those files get touched in a more meaningful fahsion.

src/frontends/arcs/handle_connection_spec_test.cc
bgogul: done.

src/frontends/arcs/handle_connection_spec_test.cc
bgogul: done

src/frontends/arcs/particle_spec_test.cc
bgogul: done. Actually, it is not a good idea to define `operator==` outside of the class. Just made this static for now.

PR #539, opened 2022-05-02
Add `fold` and `rfold` functions to our common utils.

markww:
C++ provides `std::accumulate`, which provides the classic higher-order
function `fold`. Unfortunately, `std::accumulate` copies its accumulator
until C++20, making it unsuitable for use with accumulators that have no
copy constructor. To work around this, this change adds a `fold` and an
`rfold` function, which can handle an uncopyable accumulator. If we
advance Raksha to C++20, we can just make these wrappers around
`std::accumulate`.

### DIFF
--- a/src/common/utils/fold.h
+++ b/src/common/utils/fold.h
@@ -32,6 +32,11 @@
 
 namespace raksha::common::utils {
 
+// An implementation of `fold` that takes as inputs an iterator to the beginning
+// of some range and an iterator to the end of that range. This just folds using
+// `fold_fn` from the beginning iterator to the end iterator with an initial
+// value of `accumulated`. Corresponds very directly to `std::accumulate` with a
+// custom `BinaryOperator`.
 template <class I, class A, class F>
 A fold_iter(I begin_iter, I end_iter, A accumulated, F fold_fn) {
   if (begin_iter == end_iter) return accumulated;
@@ -40,12 +45,16 @@ A fold_iter(I begin_iter, I end_iter, A accumulated, F fold_fn) {
                    fold_fn(std::move(accumulated), current_element), fold_fn);
 }
 
+// A left fold over the provided container using `fold_fn` with an initial value
+// of `init`.
 template <class C, class A, class F>
 A fold(const C &container, A &&init, F &&fold_fn) {
   return fold_iter(std::begin(container), std::end(container),
                    std::forward<A>(init), std::forward<F>(fold_fn));
 }
 
+// A right (or reverse) fold over the provided container using `fold_fn` with an
+// initial value of `init`.
 template <class C, class A, class F>
 A rfold(const C &container, A &&init, F &&fold_fn) {
   return fold_iter(std::rbegin(container), std::rend(container),
--- a/src/common/utils/fold.h
+++ b/src/common/utils/fold.h
@@ -1,0 +1,57 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//-----------------------------------------------------------------------------
+
+#ifndef SRC_COMMON_UTILS_FOLD_H_
+#define SRC_COMMON_UTILS_FOLD_H_
+
+// This file contains a C++ implementation of the classic "fold" and
+// "right_fold" higher-order functions. This is somewhat redundant with
+// std::accumulate; unfortunately, std::accumulate requires its accumulator to
+// be copyable until C++20.
+//
+// Note that "right_fold" is named "rfold". This is a bit of a pun: if you come
+// from a functional languages background, the "r" naturally expands to "right",
+// while if you are fluent with C++ iterators, "r" naturally means "reverse".
+// In this context it actually means both: a right_fold achieved by using
+// reverse iterators on the provided container.
+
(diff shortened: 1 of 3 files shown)


src/common/utils/fold.h
```diff
@@ -0,0 +1,57 @@
...
+template <class I, class A, class F>
+A fold_iter(I begin_iter, I end_iter, A accumulated, F fold_fn) {
+  if (begin_iter == end_iter) return accumulated;
+  const auto &current_element = *begin_iter;
+  return fold_iter(++begin_iter, end_iter,
+                   fold_fn(std::move(accumulated), current_element), fold_fn);
+}
+
+template <class C, class A, class F>
+A fold(const C &container, A &&init, F &&fold_fn) {
```
bgogul, line 44: documentation here and below?

src/common/utils/fold.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include <iterator>
+
+namespace raksha::common::utils {
+
+template <class I, class A, class F>
+A fold_iter(I begin_iter, I end_iter, A accumulated, F fold_fn) {
+  if (begin_iter == end_iter) return accumulated;
+  const auto &current_element = *begin_iter;
+  return fold_iter(++begin_iter, end_iter,
```
bgogul, line 39: Can't you convert this to loop, if you change call in `fold` to pass reverse range and `rfold` to pass forward range?

(This suggestion makes sense only if `fold_iter` is private. Otherwise, let us leave this as is.)

src/common/utils/fold_test.cc
```diff
@@ -0,0 +1,96 @@
...
+
+TEST_P(ConcatFoldTest, ConcatFoldTest) {
+  const std::vector<std::string> &string_vec = GetParam();
+  auto concat_fn = [](std::string concat_so_far, const std::string &input) {
+    // Yes, this is bad form in general, but the quadraticness shouldn't matter
+    // for a test.
+    return concat_so_far + input;
+  };
+  EXPECT_EQ(fold(string_vec, std::string(""), concat_fn),
+            absl::StrJoin(string_vec, ""));
```
bgogul, line 58: Instead of computing `StrJoin`, it would be more clear to pass the expected string itself (just like `IntVecSum` case)? 

(I don't feel strongly about this specific test. However, in general, I think it is better to avoid computation of expected values because we could inadvertently introduce bugs in expected values and have tests pass when they shouldn't.)

src/common/utils/fold_test.cc
```diff
@@ -0,0 +1,96 @@
...
+
+TEST_P(BoxFoldTest, BoxFoldTest) {
+  const std::vector<std::string> &string_vec = GetParam();
+  auto box_fn = [](std::vector<std::unique_ptr<std::string>> boxed_so_far,
+                   const std::string &input) {
+    boxed_so_far.push_back(std::make_unique<std::string>(input));
+    return boxed_so_far;
+  };
+  std::vector<std::unique_ptr<std::string>> boxed_strings =
+      fold(string_vec, std::vector<std::unique_ptr<std::string>>(), box_fn);
```
bgogul, line 82: Move this closer to unboxing `fold` call.

src/common/utils/fold_test.cc
```diff
@@ -0,0 +1,96 @@
...
+  auto unbox_fn = [](std::vector<std::string> unboxed_so_far,
+                     const std::unique_ptr<std::string> &to_unbox) {
+    unboxed_so_far.push_back(*to_unbox);
+    return unboxed_so_far;
+  };
+  std::vector<std::string> unboxed_strings =
+      fold(boxed_strings, std::vector<std::string>(), unbox_fn);
```
bgogul, line 89: Doesn't passing `{}` for initial value work?
```suggestion
  std::vector<std::string> unboxed_strings =
      fold(boxed_strings, {}, unbox_fn);
```

bgogul (commented):
Minor clean up/style suggestions...

src/common/utils/fold.h
```diff
@@ -0,0 +1,57 @@
...
+// while if you are fluent with C++ iterators, "r" naturally means "reverse".
+// In this context it actually means both: a right_fold achieved by using
+// reverse iterators on the provided container.
+
+#include <iterator>
+
+namespace raksha::common::utils {
+
+template <class I, class A, class F>
+A fold_iter(I begin_iter, I end_iter, A accumulated, F fold_fn) {
```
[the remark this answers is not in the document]
markww: I was thinking that it might be useful to expose it, as this allows you to `fold` a wider variety of things. For instance, with `fold_iter`, you can `fold` over regular old pointers on a contiguous chunk of memory. It also corresponds very directly with `std::accumulate`, so a transition to C++20 could involve just making the body of this function a call to `std::accumulate`.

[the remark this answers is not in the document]
markww: Added a test and comment for this function, as one should do for fns in the public API.

src/common/utils/fold.h
markww: `fold_iter` is public, leaving as is.

src/common/utils/fold_test.cc
markww: It's inferring the type provided in type parameter `A` from the passed-in `std::vector<std::string>` here. Using a `std::initializer_list` wouldn't provide type information and would make the compiler unhappy unless I explicitly provided type arguments.

src/common/utils/fold.h
markww: Done.

src/common/utils/fold_test.cc
markww: I inlined the lambdas in this test, which should move things closer together and remove name mental-model overhead generally.

src/common/utils/fold_test.cc
markww: Done.

src/common/utils/fold.h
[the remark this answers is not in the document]
bgogul: just call it `fold` then?

bgogul (approved):

src/common/utils/fold.h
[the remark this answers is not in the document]
markww: Done.

PR #540, opened 2022-05-03
Add ctor to `IRContext` allowing creating with initial contents.

markww:
Prior to this commit, all `Operator`s and `Storage`s only got into an
`IRContext` by being passed into their respective `Register` function.
While this makes sense for many circumstances, it makes it harder to
write certain tests, as objects that need an `IRContext` cannot be
constructed in the static scope without global mutable state. In
addition, it could make sense to have an `IRContext` with a known,
initial set of `Operator`s and `Storage`s: if we know an exact set of
allowed `Operation`s in advance, why not just build the `IRContext` with
them in there from the start?

This commit adds a constructor to `IRContext` that takes a `std::vector`
of initial `Operation`s and `Storage`s to seed the `IRContext` with. The
constructor then just calls `Register` for each of those items. This
commit also uses this new constructor in a test where it can make the
test more concise.

### DIFF
--- a/src/ir/block_builder_test.cc
+++ b/src/ir/block_builder_test.cc
@@ -29,12 +29,16 @@ namespace {
 
 class BlockBuilderTest : public testing::Test {
  public:
-  BlockBuilderTest() {
-    context_.RegisterOperator(std::make_unique<Operator>("core.plus"));
-    context_.RegisterOperator(std::make_unique<Operator>("core.minus"));
-    context_.RegisterOperator(std::make_unique<Operator>("core.merge"));
-    context_.RegisterOperator(std::make_unique<Operator>("core.choose"));
-  }
+  BlockBuilderTest()
+      : type_factory_(),
+        context_(
+            {
+                Operator("core.plus"),
+                Operator("core.minus"),
+                Operator("core.merge"),
+                Operator("core.choose"),
+            },
+            {}) {}
 
   types::Type MakeTestEntityType(absl::string_view schema_name) {
     const types::Schema& schema =
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -33,6 +33,19 @@ class IRContext {
   IRContext(const IRContext &) = delete;
   IRContext &operator=(const IRContext &) = delete;
 
+  // Create a context with some number of operators and storages pre-seeded.
+  IRContext(std::vector<ir::Operator> operators,
+            std::vector<ir::Storage> storages) {
+    for (auto iter = std::make_move_iterator(operators.begin());
+         iter != std::make_move_iterator(operators.end()); ++iter) {
+      RegisterOperator(std::make_unique<Operator>(*iter));
+    }
+    for (auto iter = std::make_move_iterator(storages.begin());
+         iter != std::make_move_iterator(storages.end()); ++iter) {
+      RegisterStorage(std::make_unique<Storage>(*iter));
+    }
+  }
+
   types::TypeFactory &type_factory() { return type_factory_; }
 
   // Register an operator and return the stored operator instances.
--- a/src/ir/ir_context.h
+++ b/src/ir/ir_context.h
@@ -36,13 +36,11 @@ class IRContext {
   // Create a context with some number of operators and storages pre-seeded.
   IRContext(std::vector<ir::Operator> operators,
             std::vector<ir::Storage> storages) {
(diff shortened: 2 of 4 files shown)


src/ir/ir_context.h
```diff
@@ -33,6 +33,19 @@ class IRContext {
   IRContext(const IRContext &) = delete;
   IRContext &operator=(const IRContext &) = delete;
 
+  // Create a context with some number of operators and storages pre-seeded.
+  IRContext(std::vector<ir::Operator> operators,
+            std::vector<ir::Storage> storages) {
+    for (auto iter = std::make_move_iterator(operators.begin());
+         iter != std::make_move_iterator(operators.end()); ++iter) {
+      RegisterOperator(std::make_unique<Operator>(*iter));
+    }
```
bgogul, line 42: Doesn't the following work here and below?
```suggestion
    for (const auto& op: operators) {
      RegisterOperator(std::move(op));
    }
```

src/ir/block_builder_test.cc
```diff
@@ -29,12 +29,16 @@ namespace {
 
 class BlockBuilderTest : public testing::Test {
  public:
-  BlockBuilderTest() {
-    context_.RegisterOperator(std::make_unique<Operator>("core.plus"));
-    context_.RegisterOperator(std::make_unique<Operator>("core.minus"));
-    context_.RegisterOperator(std::make_unique<Operator>("core.merge"));
-    context_.RegisterOperator(std::make_unique<Operator>("core.choose"));
-  }
+  BlockBuilderTest()
```
bgogul, line 32: Add test to https://github.com/google-research/raksha/blob/main/src/ir/ir_context_test.cc?

bgogul (commented):
Looks good except for adding a unit test to ir_context_test.cc

bgogul (approved):
Will approve so that you are not blocked on me, but please add a unit test to ir_context_test.cc

src/ir/ir_context.h
markww: It works in the sense that it compiles, but it performs a copy that the current version does not.

While `std::move` will accept a `const` value or a reference, it silently performs a copy in these cases. This is to enable templatized functions to be less annoying if they are instantiated with both `const` and non-`const` types. Because the current code uses a `move` iterator, it can move the values (and maybe copy elide them) into the destination without making a copy.

bgogul: Ah, `const&` was an oversight. Won't removing the `const` ensure it is only moved?
```
for (auto& op: operators) {
      RegisterOperator(std::move(op));
    }
```

markww: This is where things get subtle. I think it still does not, but I am not 100% sure. See this: https://abseil.io/tips/77

My count of the above is that, since `op` is a reference pointing into `operators` which is still considered a live name for all of its contents, there are two names at the time when `std::move` is called on `op`. The name `op` is removed, but the name `operators` is still live, so a copy must be performed.

In the case where a move iterator is used, I think it takes care of the two-name situation for you.

But this is getting to the edge of my ability to reason about C++'s strange semantics.

bgogul: another thought, does this work?
```
 for (auto&& op: operators) {
     RegisterOperator(op);
 }
```

bgogul: Your code with move iterators is totally fine. We can leave it as is if none of the options we discussed here work.  I was just trying to see if there is a shorter code to achieve the same thing and I am also learning as I go. I apologize if I come across as being picky. :( 

Based on  https://abseil.io/tips/77, I think the following should work:
```
for (auto& op: operators) {
  RegisterOperator(std::move(op));
}
```
IIUC, std::move creates an rvalue reference (just like a move iterator). Further, RegisterOperator` takes a `unique_ptr`, it can only be moved?

markww: That makes sense, I think probing the options is potentially valuable. I'll experiment to see if any of these shorter options work.

src/ir/block_builder_test.cc
markww: Done.

src/ir/ir_context.h
markww: I tried the above options with `Storage`, which has a `delete`d copy constructor. It turns out that

```C++
for (Storage &storage : storages) {
  RegisterStorage(std::make_unique_ptr(std::move(storage)));
}
```

works as you'd expect! That was a surprise to me, but a welcome one.

PR #542, opened 2022-05-04
Add constructors for the SQL operations.

markww:
This commit adds constructors to the SQL operations that allow these
operations to be built more easily at the static scope for tests. The
existing `Create` methods are defined in terms of these constructors.

Note for reviewers: see comment in MergeOp, would be interested in
feedback on how to make that better.

### DIFF
--- a/src/frontends/sql/ops/literal_op.cc
+++ b/src/frontends/sql/ops/literal_op.cc
@@ -26,17 +26,15 @@
 
 namespace raksha::frontends::sql {
 
-std::unique_ptr<LiteralOp> LiteralOp::Create(const ir::Block* parent_block,
-                                             const ir::IRContext& context,
-                                             absl::string_view literal) {
-  return std::make_unique<LiteralOp>(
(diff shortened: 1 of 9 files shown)


bgogul:
As I was reviewing the PR, I realized that there might be a simpler way to deal with `unique_ptr` in tests. Why can't we have something like a TestUtil class (either standalone or inherit from `testing::Test`) that has a method like this?

```C++
class TestUtil {
  template <typename T> 
  const Operation& Create<T>(Args&&... a) {
     ops_.push_back(T::create( std::forward<Args>(a)...));
     return *ops_.back();
  }
  ...
private:
   // These could also be static if you want to use it for `testing::Values`
   IRContext context_; 
  std::vector<std::unique_ptr<Operation>> ops_; 
};
```
This will also avoid the awkward code changes for `MergeOp` in this PR. WDYT?

markww:
This has been long abandoned, closing.

PR #543, opened 2022-05-04
lowering AST to datalog

harsha-mandadi:
1. Move the Relation declaration AST to datalog namespace 
2. Relation declarations lowering to datalog


src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+                                       "Principal")),
+       datalog::Argument(
+           "p2", datalog::ArgumentType(datalog::ArgumentType::Kind::kPrincipal,
+                                       "Principal"))}));
+  // Attributes have a principal that the attribute is applied to
+  // as an additional parameter at the beginning of the param list.
+  // Transform the attributes
+  std::vector<datalog::RelationDeclaration> transformed_declarations =
+      utils::MapIter<datalog::RelationDeclaration>(
+          base_declarations, [](datalog::RelationDeclaration declaration) {
```
bgogul, line 308: `const datalog::RelationDeclaration declaration&`

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+            if (declaration.is_attribute()) {
+              std::vector<datalog::Argument> transformed_arguments = {
+                  datalog::Argument("attribute_prin",
+                                    datalog::ArgumentType(
+                                        datalog::ArgumentType::Kind::kPrincipal,
+                                        "Principal"))};
+              transformed_arguments.insert(transformed_arguments.end(),
+                                           declaration.arguments().begin(),
+                                           declaration.arguments().end());
```
bgogul, line 317:
```C++
absl::c_copy(declaration.arguments(), std::back_inserter(transformed_arguments));
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+  std::vector<datalog::RelationDeclaration> base_declarations =
+      relation_declarations;
+  base_declarations.push_back(datalog::RelationDeclaration(
+      "canActAs", false,
+      {datalog::Argument(
+           "p1", datalog::ArgumentType(datalog::ArgumentType::Kind::kPrincipal,
+                                       "Principal")),
+       datalog::Argument(
+           "p2", datalog::ArgumentType(datalog::ArgumentType::Kind::kPrincipal,
+                                       "Principal"))}));
```
bgogul, line 302: Can't you add this directly to the transformed_declarations afterwards? That way you don't have to make a copy of `relation_declarations` at line 293 above?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+              return declaration;
+            }
+          });
+  // Produce a mapping from predicate names to predicate typings
+  // (where a predicate typing is the same as a predicate relation declaration)
+  absl::flat_hash_map<std::string_view, datalog::RelationDeclaration>
+      type_environment;
+  for (datalog::RelationDeclaration declaration : transformed_declarations) {
+    type_environment.insert({declaration.relation_name(), declaration});
+  }
```
bgogul, line 330: `for (const datalog::RelationDeclaration& declaration : transformed_declarations)`

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
     main_pred = PushPrincipal("says_", query.principal(), main_pred);
-    datalog::Predicate lhs(query.name(), {"dummy_var"}, datalog::kPositive);
+    datalog::Predicate lhs(query.name(), {R"("dummy_var")"},
+                           datalog::kPositive);
     return datalog::Rule(lhs, {main_pred, kDummyPredicate});
   });
 }
 
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::RelationDeclarationToDLIR(
```
bgogul, line 283: This function is too long, can you split this function into multiple smaller ones?  _Roughly_, each loop in this class could it be its own function.

src/ir/auth_logic/lowering_ast_datalog.h
```diff
@@ -207,12 +207,32 @@ class LoweringToDatalogPass {
...
+  // like: `says_cansay_..._cansay_foo(A1, A2, ..., An, args)`, so to generate
+  // these declarations, we need to know the number of "canSays" to include for
+  // each base predicate (e.g., foo). This pass does this by tracking the
+  // maximum depth of "canSay"s with which each predicate is used and storing
+  // this in the `cansay_depth` HashMap. When a predicate appears with depth
+  // `n`, it may also need to be used with depths (1..n), so we only maintain
+  // the maximum depth in which it appears and we generate declarations for
+  // (1..n). `cansay_depth` is only updated by `update_cansay_depth` which
+  // tracks this maximum.
+  absl::flat_hash_map<std::string_view, uint64_t> can_say_depth;
```
bgogul, line 235: C++ style: member variables hsould have a `_` at the end. Also `can_say_depths_` (plural) ?

`can_say_depth` ->`can_say_depths_`


Also, what does the key of this map represent?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+  absl::flat_hash_map<std::string_view, datalog::RelationDeclaration>
+      type_environment;
+  for (datalog::RelationDeclaration declaration : transformed_declarations) {
+    type_environment.insert({declaration.relation_name(), declaration});
+  }
+
+  // Generate relation declarations for delegated predicates with
+  // each encountered delegation depth
+  std::vector<datalog::RelationDeclaration> can_say_declarations;
+  for (const auto& can_say_depth_pair : can_say_depth) {
```
bgogul, line 335: You can use [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding):
```C++
for (const auto& [type_name, depth]: can_say_depth) {
...
}
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+    auto type_iterator = type_environment.find(can_say_depth_pair.first);
+
+    if ((can_say_depth_pair.second > 0) &&
+        (type_iterator != type_environment.end())) {
```
bgogul, line 339: Negating the conditions might make it easier to read:
```suggestion
    if (depth == 0) continue;
    auto type_iterator = type_environment.find(type_name);
    if (type_iterator == type_environment.end() continue;
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+                                  "Principal"))};
+        argument.insert(argument.end(), declaration.arguments().begin(),
+                        declaration.arguments().end());
+        can_say_declarations.push_back(datalog::RelationDeclaration(
+            absl::StrCat("canSay_", declaration.relation_name()), false,
+            argument));
+        declaration = datalog::RelationDeclaration(
+            absl::StrCat("canSay_", declaration.relation_name()), false,
+            argument);
+      }
```
bgogul, line 354: This loop is a bit hard to follow. Can you give me an example of what is the result of the transition here is?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,129 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+            arguments));
+  }
+  return says_extended_relational_declarations;
+}
+
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::QueryRelationDeclarationToDLIR(
+    const std::vector<Query>& queries) {
+  std::vector<datalog::RelationDeclaration> query_declarations =
+      utils::MapIter<datalog::RelationDeclaration>(queries, [](Query query) {
```
bgogul, line 382: `const Query&`

Having `Query query` will make a copy, which we do not want?

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -61,6 +67,13 @@ class SouffleEmitter {
   }
 
   std::string EmitPredicate(const datalog::Predicate& predicate) {
+    std::vector<std::string> comparision_operators = {"<",  ">",  "=",
+                                                      "!=", "<=", ">="};
+    if (std::find(comparision_operators.begin(), comparision_operators.end(),
+                  predicate.name()) != comparision_operators.end()) {
+      return absl::StrCat(predicate.args().front(), predicate.name(),
+                          predicate.args().back());
+    }
```
bgogul, line 76: A few chagnes:
 - comparision_operators` -> `comparison_operators`.
 - check the size of `predicate_args`.
 - Define a static array to avoid construction of the vector every time this method is called.

```suggestion
    const absl::string_view comparson_operators[] = {"<", ">", "=", "!=", "<=", ">="};
    if (std::find(std::begin(comparison_operators), std::end(comparison_operators),
                  predicate.name()) != std::end(comparison_operators)) {
      CHECK(predicate_args.size() == 2);
      return absl::StrCat(predicate.args().front(), predicate.name(),
                          predicate.args().back());
    }
```

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +117,47 @@ class SouffleEmitter {
...
+        });
+  }
+
+  std::string EmitRelationDeclarations(
+      const datalog::Program& program,
+      const std::vector<std::string>& skip_declarations) {
+    std::vector<std::string> declaration_string;
+    for (const auto& declaration : program.relation_declarations()) {
+      if (std::find(skip_declarations.begin(), skip_declarations.end(),
+                    declaration.relation_name()) == skip_declarations.end()) {
```
bgogul, line 134: use `absl:c_find`. Also, invert condition with `continue` to make it slightly easier to read.

```C++
if (absl::c_find(skip_declarations, declaration.relation_name() != skip_declarations.end())  continue;
```

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +117,47 @@ class SouffleEmitter {
...
+        arguments, ", ", [](std::string* out, datalog::Argument argument) {
+          return absl::StrAppend(out, argument.argument_name(), " : ",
+                                 argument.argument_type().name());
+        });
+  }
+
+  std::string EmitRelationDeclarations(
+      const datalog::Program& program,
+      const std::vector<std::string>& skip_declarations) {
+    std::vector<std::string> declaration_string;
```
bgogul, line 131: nit: `declaration_strings`

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +117,47 @@ class SouffleEmitter {
...
+  std::string EmitTypeDeclarations(
+      const datalog::Program& program,
+      const std::vector<std::string>& skip_declarations) {
+    std::vector<std::string> type_names;
+    for (const auto& declaration : program.relation_declarations()) {
+      for (const auto& argument : declaration.arguments()) {
+        if ((argument.argument_type().kind() ==
+             datalog::ArgumentType::Kind::kCustom) &&
+            (std::find(skip_declarations.begin(), skip_declarations.end(),
+                       argument.argument_name()) == skip_declarations.end())) {
```
bgogul, line 153: Invert conditions with `continue`
```suggestion
        if ((argument.argument_type().kind() != datalog::ArgumentType::Kind::kCustom)  continue;
        if  (absl::c_find(skip_declarations, argument.argument_name()) != skip_declarations.end()) continue;
        type_names.push_back(...);
```

bgogul (commented):
Looks decent structure wise, but had a few style comments. If you can address them, I will take another look.

src/ir/auth_logic/lowering_ast_datalog.h
harsha-mandadi: Key here is a predicate name. We check if there is any relation name that matches the predicate name and incrementally add new relation declarations based on the depth. Example:` P1 canSay P2 canSay grantAccess(x, y)`. Key here would be `grantAccess` with a depth value of 2

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: Yes. That will makes things simpler. I will do that!

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: You are right. That would be efficient. I will do that.

src/ir/auth_logic/souffle_emitter_test.cc
```diff
@@ -42,220 +42,241 @@ Program BuildPredicateTestProgram() {
...
-  //
-  // .decl theTestQuery(x0: symbol)
-  // .decl says_anything(x0: symbol, x1: symbol)
-  // .decl grounded_dummy(x0: symbol))";
-
-  std::string actual = SouffleEmitter::EmitProgram(
-      LoweringToDatalogPass::Lower(BuildQueryProgram()));
-  EXPECT_EQ(actual, expected);
-}
+// Program BuildAttributeTestProgram() {
```
aferr, line 72: Is there a reason to include this commented out code in the repository? If not, it's probably better to delete this rather than comment it out. If for some reason we need to refer back to this commented out code, it's still in the git revision history. (Just make a note of the commit where this code was last in the repo so that you can find it again).

harsha-mandadi: Removed commented out section!

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: Done!

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: Done!

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +120,48 @@ class SouffleEmitter {
...
 
-  std::string EmitDeclarations() {
-    std::vector elements_vec(std::make_move_iterator(declarations_.begin()),
-                             std::make_move_iterator(declarations_.end()));
-    std::sort(elements_vec.begin(), elements_vec.end());
-    return absl::StrJoin(elements_vec, "\n",
-                         [this](std::string* out, auto decl) {
-                           return absl::StrAppend(out, EmitDeclaration(decl));
-                         });
+  std::string EmitArguments(std::vector<datalog::Argument> arguments) {
```
bgogul, line 123: const reference instead of value here: `const vector...& arguments`?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+      utils::MapIter<datalog::RelationDeclaration>(
+          queries, [](const Query& query) {
+            return datalog::RelationDeclaration(
+                query.name(), false,
+                {datalog::Argument(
+                    "dummy_param",
+                    datalog::ArgumentType(datalog::ArgumentType::Kind::kCustom,
+                                          "DummyType"))});
+          });
+  query_declarations.push_back(datalog::RelationDeclaration(
```
bgogul, line 441: Add a comment on what declaration you are adding.  Something like this: 

```C++
// .decl grounded_dummy(dummy_param: DummyType)
```
(Correct the comment as needed.)

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+      says_extended_relational_declarations =
+          LoweringToDatalogPass::SaysExtendRelationDeclarations(
+              transformed_declarations);
+  return says_extended_relational_declarations;
+}
+
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::QueryRelationDeclarationToDLIR(
+    const std::vector<Query>& queries) {
+  std::vector<datalog::RelationDeclaration> query_declarations =
```
bgogul, line 431: Add a comment like this?

```C++
// For every query add the following declaration:
//   .decl query_name(dummy_param: DummyType)
```

Is my understanding of what is generated correct?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+  absl::flat_hash_map<std::string_view, datalog::RelationDeclaration>
+      type_environment;
+  for (const datalog::RelationDeclaration& declaration :
+       transformed_declarations) {
+    type_environment.insert({declaration.relation_name(), declaration});
+  }
+
+  std::vector<datalog::RelationDeclaration> can_say_declarations =
+      LoweringToDatalogPass::CanSayDeclarations(type_environment);
+  absl::c_copy(can_say_declarations,
```
bgogul, line 398: Should this be `c_move`? If so, you simply do the following?

```C++
absl::c_move(
  LoweringToDatalogPass::CanSayDeclarations(type_environment), 
  std::back_inserter(transformed_declarations));
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+               std::back_inserter(transformed_declarations));
+
+  // The transformed relation declarations are the same as the ones from the
+  // surface program plus another one for "canActAs"
```
bgogul, line 402: Add a comment about what datalog relation or datalog declaration is added.  Is it the following?

```C++
.decl  canActAs(P1: Principal, p2: Principal)
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+      "canActAs", false,
+      {datalog::Argument(
+           "p1", datalog::ArgumentType(datalog::ArgumentType::Kind::kPrincipal,
+                                       "Principal")),
+       datalog::Argument(
+           "p2", datalog::ArgumentType(datalog::ArgumentType::Kind::kPrincipal,
+                                       "Principal"))}));
+  // Another declaration "isNumber" added aa a work around for
+  // universe_translations. To be removed after adding parser for
+  // universe_translations
```
bgogul, line 413: Adding a datalog representation of the decl will be more useful:
```C++
.decl isNumber(x: Number)
```

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+          arguments);
+    }
+  }
+  return can_say_declarations;
+}
+
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::SaysExtendRelationDeclarations(
+    const std::vector<datalog::RelationDeclaration>& transformed_declarations) {
+  // The translated declarations are all extended with "says_" and a speaker
```
bgogul, line 357: Can you add a comment showing what the transformation is? 
e.g.,
```C++
// some_relation(args..) -> says_some_relation(speaker: Principal, args...)
```

You could also something similar to show what the transformations are more clearly in other places.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+PrincipalA cansay PrincipalB grantAccess(secretFile)."
+can_say_depth_(A hashmap) maps "grantAccess" predicate name to depth 2.
+type_environment is a mapping from the relation names to the declaration
+datalog_ir node. In this example, type_environment[grantAccess] maps to
+RelationDeclaration(grantAccess, false, ...).
+The method adds 2 new declarations
+.decl canSay_grantAccess(delegatee1 : Principal, x0 : symbol, x1 : symbol) and
+.decl canSay_canSay_grantAccess(delegatee2 : Principal, delegatee1: Principal,
+x0 : symbol, x1 : symbol)
+*/
```
bgogul, line 322: `//` comment style?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+        type_environment) {
+  // Generate relation declarations for delegated predicates with
+  // each encountered delegation depth
+  std::vector<datalog::RelationDeclaration> can_say_declarations;
+  for (const auto& [predicate_name, depth] : can_say_depth_) {
+    if (depth == 0) continue;
+    // find relation declaration corresponding to the predicate name
+    auto type_iterator = type_environment.find(predicate_name);
+    if (type_iterator != type_environment.end()) continue;
+    datalog::RelationDeclaration declaration = type_iterator->second;
```
bgogul, line 335: `const & declaration`?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+      declaration = datalog::RelationDeclaration(
+          absl::StrCat("canSay_", declaration.relation_name()), false,
+          arguments);
+    }
+  }
+  return can_say_declarations;
+}
+
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::SaysExtendRelationDeclarations(
```
bgogul, line 355: `SaysExtendRelationDeclarations` -> `GetSaysExtendedRelationDeclarations`?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+type_environment is a mapping from the relation names to the declaration
+datalog_ir node. In this example, type_environment[grantAccess] maps to
+RelationDeclaration(grantAccess, false, ...).
+The method adds 2 new declarations
+.decl canSay_grantAccess(delegatee1 : Principal, x0 : symbol, x1 : symbol) and
+.decl canSay_canSay_grantAccess(delegatee2 : Principal, delegatee1: Principal,
+x0 : symbol, x1 : symbol)
+*/
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::CanSayDeclarations(
```
bgogul, line 324: `CanSayDeclarations` -> `GetCanSayDeclarations`?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
     main_pred = PushPrincipal("says_", query.principal(), main_pred);
-    datalog::Predicate lhs(query.name(), {"dummy_var"}, datalog::kPositive);
+    datalog::Predicate lhs(query.name(), {R"("dummy_var")"},
+                           datalog::kPositive);
     return datalog::Rule(lhs, {main_pred, kDummyPredicate});
   });
 }
 
+std::vector<datalog::RelationDeclaration>
+LoweringToDatalogPass::TransformDeclarations(
```
bgogul, line 283: More specific name?
`TransformDeclarations` -> `TransformAttributeDeclarations`

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+    const std::vector<datalog::RelationDeclaration>& relation_declarations) {
+  // Attributes have a principal that the attribute is applied to
+  // as an additional parameter at the beginning of the param list.
+  // Example below shows transforming the attributes.
+  // Ex: .decl ATTRIBUTE hasProperty(x0 : Symbol) -> .decl
+  // hasProperty(attribute_prin : Prinipal, x0 : Symbol)
```
bgogul, line 289: nit: Move this comment block to just above the function?

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+  return utils::MapIter<datalog::RelationDeclaration>(
+      relation_declarations,
+      [](const datalog::RelationDeclaration& declaration) {
+        if (declaration.is_attribute()) {
```
bgogul, line 293: Inverting the condition will make it a tad more readable:

```C++
if (!declaration.is_attribute()) return declarations;
....
return dataog::RelationDeclaration(...)
```

bgogul (commented):
Looks quite good. My main question is about why some tests have been deleted. 

My other feedback is mostly centered around documentation and naming nits.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -67,6 +67,15 @@ datalog::Predicate CanActAsToDLIR(const CanActAs& can_act_as) {
...
 }  // namespace
 
+void LoweringToDatalogPass::UpdateCanSayDepth(std::string_view predicate_name,
+                                              uint64_t depth) {
+  if (can_say_depth_.find(predicate_name) == can_say_depth_.end()) {
+    can_say_depth_.insert({predicate_name, depth});
+  } else if (depth > can_say_depth_.find(predicate_name)->second) {
+    can_say_depth_.find(predicate_name)->second = depth;
+  }
+}
```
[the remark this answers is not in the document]
markww: `can_say_depth` is a map, so this can be made even simpler. On C++ maps, `std::insert` returns a result indicating if the result successfully occurred or whether there was a key collision on insertion. If there is a key collision, it does not perform the insertion over the key, instead returning an iterator to the existing entry.

This means that you can rewrite this like so:

```C++
auto insert_result = can_say_depth.insert({predicate_name, depth});
if (!insert_result.second) {
  insert_result.first->second = std::max(depth, insert_result.first->second);
}
```

(although some temporaries to

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +273,180 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+    std::vector<datalog::Argument> arguments = {datalog::Argument(
+        "speaker", datalog::ArgumentType(
+                       datalog::ArgumentType::Kind::kPrincipal, "Principal"))};
+    absl::c_copy(declaration.arguments(), std::back_inserter(arguments));
+    says_extended_relational_declarations.push_back(
+        datalog::RelationDeclaration(
+            absl::StrCat("says_", declaration.relation_name()), false,
+            arguments));
+  }
+  return says_extended_relational_declarations;
```
markww, line 372: Would this be more concise as a `MapIter` call, the result of which can be directly `return`ed from this function?

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +120,48 @@ class SouffleEmitter {
...
+        });
+  }
+
+  std::string EmitRelationDeclarations(
+      const datalog::Program& program,
+      const std::vector<std::string>& skip_declarations) {
+    std::vector<std::string> declaration_strings;
+    for (const auto& declaration : program.relation_declarations()) {
+      if (absl::c_find(skip_declarations, declaration.relation_name()) !=
+          skip_declarations.end())
```
markww, line 137: This `c_find` is `O(N)`, where `N` is the number of `skip_declarations`. It could instead be `O(1)` if `skip_declarations` were instead an `absl::flat_hash_set` of `std::string`.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +120,48 @@ class SouffleEmitter {
...
+      const std::vector<std::string>& skip_declarations) {
+    std::vector<std::string> type_names;
+    for (const auto& declaration : program.relation_declarations()) {
+      for (const auto& argument : declaration.arguments()) {
+        if (argument.argument_type().kind() !=
+            datalog::ArgumentType::Kind::kCustom)
+          continue;
+        if (absl::c_find(skip_declarations, argument.argument_name()) !=
+            skip_declarations.end())
+          continue;
```
markww, line 158: Again, unnecessarily linear. Change `skip_declarations` to be an `absl::flat_hash_set` to cut down the asymptotic time.

src/ir/auth_logic/souffle_emitter.h
bgogul: Ah, I thought `skip_declarations` was a `set`

src/ir/auth_logic/souffle_emitter_test.cc
```diff
@@ -42,219 +42,32 @@ Program BuildPredicateTestProgram() {
                                                      datalog::kPositive))))));
 }
 
-TEST(EmitterTestSuite, SimpleTest) {
```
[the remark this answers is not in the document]
harsha-mandadi: Removed most of the tests written earlier as they would fail with current changes. Before this PR, relation declarations were generated from predicate name(As a word around I guess). For this reason, tests written earlier did not have correct relation declarations and were missing type declarations, canActAs declaration etc. I wrote new tests with couple of base cases (empty auth logic and simple predicate fact). More tests needs to be added moving forward. I will take care of that in another PR.

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: Yes that is correct.

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: `declaration` here is being modified n the for loop below. We assign a new `RelationDeclaration` in each loop.

src/ir/auth_logic/lowering_ast_datalog.cc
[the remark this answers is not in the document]
harsha-mandadi: That is a very good optimization I was not aware of. Thanks Mark!

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: Agree. Made these changes accordingly.

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: Ahh I missed this optimization. Thanks for the suggestion!

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: Oops. Missed this!

harsha-mandadi:
> Looks quite good. My main question is about why some tests have been deleted.
> 
> My other feedback is mostly centered around documentation and naming nits.

Made most of the changes suggested. Left comments for the ones I did not make change changes and left comments. Please take a look again.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +272,179 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+          datalog::ArgumentType(datalog::ArgumentType::Kind::kPrincipal,
+                                "Principal"))};
+      absl::c_copy(declaration.arguments(), std::back_inserter(arguments));
+      can_say_declarations.push_back(datalog::RelationDeclaration(
+          absl::StrCat("canSay_", declaration.relation_name()), false,
+          arguments));
+      declaration = datalog::RelationDeclaration(
+          absl::StrCat("canSay_", declaration.relation_name()), false,
+          arguments);
+    }
```
markww, line 342: These two lines build the same `RelationDeclaration` twice.

I think overwriting `declaration` in this fashion is a bad pattern. In general, when possible, we should strive to not change the values of variables, treating them as if they were declared `const`.  The reason why is that mutation makes it harder to follow the logic. I don't know what you are trying to do by overwriting `declaration` on each iteration in this fashion.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -195,7 +203,9 @@ LoweringToDatalogPass::FactToDLIR(const Principal& speaker, const Fact& fact) {
     datalog::Predicate prin_cansay_pred =
         PushPrincipal("canSay_", delegatees, inner_fact_dlir);
     inner_fact_dlir = prin_cansay_pred;
+    can_say_depth_inner += 1;
```
markww, line 206: Doesn't this logic just make `can_say_depth_inner` have the same value as `fact.delecation_chain().size()`? Would it be more concise to just pass that into `UpdateCanSayDepth`?

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -102,14 +122,48 @@ class SouffleEmitter {
...
+        });
+  }
+
+  std::string EmitRelationDeclarations(
+      const datalog::Program& program,
+      const absl::flat_hash_set<std::string>& skip_declarations) {
+    std::vector<std::string> declaration_strings;
+    for (const auto& declaration : program.relation_declarations()) {
+      if (absl::c_find(skip_declarations, declaration.relation_name()) !=
+          skip_declarations.end())
```
markww, line 139: Why `c_find` rather than just calling `skip_declarations.find(declaration.relation_name())`?

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: Ahh yes. That was silly of me. I shouldn't have coded it that way.

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: Just curious. Is `skip_declarations.find(..)` better in terms of performance than `c_find`?

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: You are correct. That would eliminate introducing a new variable. I will add some more tests around can_say_depth just to be safe here.

src/ir/auth_logic/souffle_emitter.h
bgogul: Yes, I think it just `c_find` does a linear search. `skip_declarations.find` will be amortized constant time as `skip_declarations` is hash set.

src/ir/auth_logic/lowering_ast_datalog.cc
```diff
@@ -262,11 +272,179 @@ std::vector<datalog::Rule> LoweringToDatalogPass::QueriesToDLIR(
...
+  // delegated predicates with each encountered delegation depth
+  std::vector<datalog::RelationDeclaration> can_say_declarations;
+  for (const auto& [predicate_name, depth] : can_say_depth_) {
+    if (depth == 0) continue;
+    // find relation declaration corresponding to the predicate name
+    auto type_iterator = type_environment.find(predicate_name);
+    if (type_iterator != type_environment.end()) continue;
+    datalog::RelationDeclaration declaration = type_iterator->second;
+    // Adds new relation declaration for each can say delegation
+    for (uint64_t i = 1; i < depth + 1; ++i) {
```
bgogul, line 330: Is `i <= depth` the same as `i < depth + 1`?  `depth + 1` has a possibility of overflowing. (Overflow situation is rare because depth has to be `UINT64_MAX`, but good to avoid if we can.)

src/ir/auth_logic/lowering_ast_datalog.cc
bgogul: Are you essentially arguments of all predicates in the `canSay` chain? Is this referring to the relation corresponding to the `canSay` chain that has been build so far? If so, how about you just hold onto the address as follows: 

```C++
const datalog::RelationDeclaration* can_say_chain_suffix = &type_iterator->second;
```

You are making an unncessary copy with `datalog::RelationDeclaration declaration = type_iterator->second;`.

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -61,6 +68,17 @@ class SouffleEmitter {
   }
 
   std::string EmitPredicate(const datalog::Predicate& predicate) {
+    const absl::string_view comparison_operators[] = {"<",  ">",  "=",
```
bgogul, line 71: Make this a `flat_hash_set`. 

```
static const absl::flat_hash_set<std::string> comparison_operators = {"<", ">", "=", "!=", "<=", ">="};
```

src/ir/auth_logic/souffle_emitter.h
```diff
@@ -61,6 +68,17 @@ class SouffleEmitter {
...
+                                                      "!=", "<=", ">="};
+    // Rvalue rules which are represented as datalog::Predicates in AST, we use
+    // infix notation while translating to datalog.
+    // Ex: "<"(number1, number2) -> number1 < number2.
+    if (absl::c_find(comparison_operators, predicate.name()) !=
```
bgogul, line 76: When you change `comparison_operators` to a set, 
```C++
if (comparison_operator.count(predicate.name()) > 0) {
```

This avoids a linear search as in the other cases.

src/ir/auth_logic/souffle_emitter.h
bgogul: Yes, I think `c_find` does a linear search. `skip_declarations.find` is like a hash table lookup, which should be amortized constant time.

src/ir/auth_logic/souffle_emitter.h
bgogul: skip_declarations is already a flat_hash_set, change to `skip_declarations.find` ?

src/ir/auth_logic/souffle_emitter_test.cc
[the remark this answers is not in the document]
bgogul: I see. Sounds good.

bgogul (approved):
Looks good except for a few minor things. I will approve assuming you will address these before merging it in.

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: I actually modified code to remove the need for this declaration. As per these changes, no new copies or local variables are needed. Left comments to make sure these changes are easy to understand. Please let

harsha-mandadi: I actually modified code to remove the need for this declaration. As per these changes, no new copies or local variables are needed. Left comments to make sure these changes are easy to understand. Please let

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: I see. Thanks! Made code changes to use `skip_declaration.find(..)`

harsha-mandadi: I see. Thanks! Made code changes to use `skip_declaration.find(..)`

src/ir/auth_logic/lowering_ast_datalog.cc
harsha-mandadi: ok. Done!

harsha-mandadi: ok. Done!

src/ir/auth_logic/souffle_emitter.h
harsha-mandadi: ok. Done!

harsha-mandadi: ok. Done!

PR #544, opened 2022-05-05
Dot generator for the IR

bgogul:
This  PR implements a class to return a dot representation of the given IR module. This will be useful for debugging/demos/etc.


harsha-mandadi (approved):
LGTM

PR #547, opened 2022-05-07
Fix syntax in the Adt::ToDatalogString().

bgogul:


markww (approved):
Whoops, thanks for fixing this!

PR #548, opened 2022-05-09
Change names for relations that lookup operand and attribute of an operation.

bgogul:

PR #549, opened 2022-05-09
Calculate positional indices of operands to operations.

markww:
Because lists in Souffle are just recursive linked-list structs, it is
not very natural to look up elements by index. For the tag transform
operation from the SQL verifier, however, we need to do exactly that.
This commit introduces `operationHasOperandAtIndex`, which enriches the
old `operationToOperands` relation by adding the position the operand
holds in the operation's `operandList`.

### DIFF
--- a/src/analysis/souffle/operations.dl
+++ b/src/analysis/souffle/operations.dl
@@ -48,11 +48,32 @@ operationToPartialOperandLists([owner, op, result, operandList, attrs], operandL
 operationToPartialOperandLists(operation, tail) :-
   operationToPartialOperandLists(operation, [_, tail]).
 
-// Map full operations to operands.
-.decl operationToOperands(operation:Operation, operand: AccessPath)
+// `listLength` is the length of the `operandList`.
+.decl operandListLength(operandList: OperandList, listLength: number)
 
-operationToOperands(operation, operand) :-
-  isOperation(operation), operationToPartialOperandLists(operation, [operand, _]).
+operandListLength(nil, 0).
+operandListLength([head, tail], tailLen + 1) :-
+  operationToPartialOperandLists(_, [head, tail]), operandListLength(tail, tailLen).
+
+// `operandListLength` is the length of the `operandList` on the given
+// `operation`.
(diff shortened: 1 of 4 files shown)


src/analysis/souffle/operations.dl
```diff
@@ -48,11 +48,32 @@ operationToPartialOperandLists([owner, op, result, operandList, attrs], operandL
 operationToPartialOperandLists(operation, tail) :-
   operationToPartialOperandLists(operation, [_, tail]).
 
-// Map full operations to operands.
-.decl operationToOperands(operation:Operation, operand: AccessPath)
+// `listLength` is the length of the `operandList`.
+.decl operandListLength(operandList: OperandList, listLength: number)
```
bgogul, line 52: Should I just drop #548 and you can fix the names in this PR? Or, should I try to merge it in?

src/analysis/souffle/operations.dl
```diff
@@ -48,11 +48,32 @@ operationToPartialOperandLists([owner, op, result, operandList, attrs], operandL
...
 
-operationToOperands(operation, operand) :-
-  isOperation(operation), operationToPartialOperandLists(operation, [operand, _]).
+operandListLength(nil, 0).
+operandListLength([head, tail], tailLen + 1) :-
+  operationToPartialOperandLists(_, [head, tail]), operandListLength(tail, tailLen).
```
bgogul, line 56: If you decide to do the renaming in #548, please change this to `operationHasOperandListSuffix`. Similarly, for `operationToPartialAttributeLists`

bgogul (approved):

src/analysis/souffle/operations.dl
markww: Maybe that would be best. If we both submitted, there would be some significant conflicts that would just duplicate the work anyway.

bgogul: Sounds good. I will close the other PR.

src/analysis/souffle/operations.dl
markww: Done.

PR #552, opened 2022-05-10
Separate `.input`/`.output` directives from regular datalog.

markww:
Souffle allows indicating that certain relations are `.input` or
`.output`. The analysis will expect facts to be provided at startup to
`.input` relations and shall output facts to the `.output` relations.

Before this commit, the `.input` and `.output` directives were scattered
through the regular Datalog files. This is a bad move, as the same
analysis datalog may input and output different facts when run in
different ways. For instance, the `.input` and `.output` relations
previously specified in the datalog files were geared towards unit
tests, but were inappropriate for regular runs of the analysis.

This commit collects those `.input` and `.output` relations into a new
`tag_check_and_expectation_interface.dl` file, which indicates the
interface to the analysis for this kind of unit test. This is a pattern
I'd like to establish going forward, where each binary or library that we
produce based upon our Souffle datalog shall correspond to a separate
interface file, indicating the input and output realtions used by that
binary.

### DIFF
--- a/build_defs/test_helpers.bzl
+++ b/build_defs/test_helpers.bzl
@@ -130,11 +130,12 @@ def run_taint_exec_compare_check_results(
         outs = ["checkAndResult", "expectedCheckAndResult"],
         srcs = input_files,
         testonly = True,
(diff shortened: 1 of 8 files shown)


bgogul (approved):
Nice!

PR #554, opened 2022-05-11
Add a utility to dump facts as files in a directory.

bgogul:
This is needed to provide inputs for a souffle program run using a C++ interface.

### DIFF
--- a/src/ir/datalog/input_relation_fact.h
+++ b/src/ir/datalog/input_relation_fact.h
@@ -50,7 +50,7 @@ class InputRelationFact {
   // Returns string representation of a fact that can be used in an input
   // relations file.
   std::string ToDatalogFactsFileString() const {
-    return absl::StrJoin(relation_arguments_, ", ",
+    return absl::StrJoin(relation_arguments_, "; ",
                          [](std::string *str, const auto &arg) {
                            absl::StrAppend(str, arg.ToDatalogString());
                          });
--- a/src/ir/datalog/input_relation_fact.h
+++ b/src/ir/datalog/input_relation_fact.h
@@ -47,6 +47,15 @@ class InputRelationFact {
                                          }));
   }
 
+  // Returns string representation of a fact that can be used in an input
+  // relations file.
+  std::string ToDatalogFactsFileString() const {
+    return absl::StrJoin(relation_arguments_, ", ",
+                         [](std::string *str, const auto &arg) {
+                           absl::StrAppend(str, arg.ToDatalogString());
+                         });
+  }
+
  private:
   std::tuple<RelationParameterTypes...> relation_arguments_;
 };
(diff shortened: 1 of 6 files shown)


src/ir/datalog/input_relation_fact.h
```diff
@@ -47,6 +47,15 @@ class InputRelationFact {
...
   }
 
+  // Returns string representation of a fact that can be used in an input
+  // relations file.
+  std::string ToDatalogFactsFileString() const {
+    return absl::StrJoin(relation_arguments_, ", ",
+                         [](std::string *str, const auto &arg) {
+                           absl::StrAppend(str, arg.ToDatalogString());
+                         });
+  }
```
markww, line 57: Talked to you previously about this: I don't think `,` is a good choice of separator. For ADTs that have multiple arguments, the parser will get confused. I've been using `;`, which I'd suggest using here as well.

src/ir/datalog/input_relation_fact_test.cc
```diff
@@ -70,11 +76,13 @@ class OneOfEachFact
...
-                          Record<Number, Symbol>(Number(3), Symbol("bar")),
-                          SimpleAdt(UnitAdtBranch()))
-                .ToDatalogString(),
+  auto fact = OneOfEachFact(Number(5), Symbol("foo"),
+                            Record<Number, Symbol>(Number(3), Symbol("bar")),
+                            SimpleAdt(UnitAdtBranch()));
+  EXPECT_EQ(fact.ToDatalogString(),
             R"(oneOfEach(5, "foo", [3, "bar"], $Unit()).)");
+  EXPECT_EQ(fact.ToDatalogFactsFileString(),
+            R"(5, "foo", [3, "bar"], $Unit())");
```
markww, line 85: Ah, records will also pose a problem with `,` being your separator. Note here that `,` is used to both separate the fields in the record and delimit the boundary between the column containing the record and the next and prior column. Souffle will get very confused if it tries to read this line from an input file.

bgogul: I replicated whatever was in `ToDatalogString` without the relation name. What is the right format?

markww: You can choose whatever delimiter you want to separate rows in Souffle. I tend to choose `;`, which would cause the expected output to be as follows:

```C++
 R"(5; "foo"; [3, "bar"]; $Unit())"
 ```

File: src/ir/datalog/input_relation_fact.h (hunk 1 above)
Comment (bgogul) [reply src/ir/datalog/input_relation_fact.h]: Done.

File: src/ir/datalog/input_relation_fact_test.cc [hunk 3]
```diff
@@ -70,11 +76,13 @@ class OneOfEachFact
 };
 
 TEST(OneOfEachFactTest, OneOfEachFactTest) {
```
bgogul, line 78: @markww do we have enough test coverage?

bgogul:
@markww Addressed your comments, PTAL.

src/ir/datalog/input_relation_fact_test.cc
markww: I think we have enough unit test coverage, but I think we are missing integration test coverage. That is, these strings aren't consumed by Souffle at this point, so we can't notice issues with Souffle misparsing them. I think some integration tests having Souffle parse the results may be in order soon.

markww (approved):
LGTM

src/ir/datalog/input_relation_fact_test.cc
markww: Yes, I think that sounds right.

PR #555, opened 2022-05-11
Connect SQL verifier tag transforms to policy rules in datalog.

markww:
This change connects the SQL verifier tag transform nodes to the policy
rules that control their behavior, and shows this behavior control
occurring in unit tests.


markww:
Sorry the diff is so messy, I don't know why GitHub didn't realize that many of these commits are already on main. In any case, the only change with anything new is the one with the title that matches the PR title, so maybe you can review that commit only? Let me see if I can clean this up.

markww:
Managed to make a cleaner diff in #556 , so closing this one.

PR #558, opened 2022-05-13
Wrap the top `Value` from an arena with a new `SqlOutputOp`.

markww:
In a previous commit, we introduced the concept of a "sql.sql_output"
operator in datalog, which would cause a failure if any confidentiality
tag made it to the access path wrapped by that operation. This
introduces the code to use that concept in the IR translator by wrapping
the top level value with an operation having operator `SqlOutputOp`.


bgogul (approved):
datalog lowering needs to be updated, right? Will that be in a follow up PR?

markww:
It doesn't actually need to be updated because this is expressed as a new `Operator`/`Operation`. As such, the lowering can just handle it as it would handle any other `Operation`/`Operator`.

PR #560, opened 2022-05-14
Add a top-level driver to tie the Raksha SQL analysis together.

markww:
This commit adds a top level driver that ties together the pieces of the
SQL analysis, causing a dataflow graph proto and a string containing
datalog `isSqlPolicyRule` facts to be turned into a boolean, indicating
whether the policy was violated or not.

This change also refactors the creation of temporary directories by
creating a function in `utils` for doing it the right way that can be
called in multiple locations.


src/backends/policy_engine/souffle/datalog_fact_policy.h
```diff
@@ -0,0 +1,54 @@
...
+    return WriteFactsStringToFactsFile(output_path, relation_name_,
+                                       facts_string_);
+  }
+
+ private:
+  std::string relation_name_;
+  std::string facts_string_;
+};
+
+class SqlPolicyRuleFactPolicy : public DatalogFactPolicy {
```
bgogul, line 46: Even though the current implementation calls this a policy, it is technically not a policy. Rather,  it is part of the semantics of the operation. In the long run, I don't think we should have this at all.

See my comment in `IsModuleCompliantInner`. We should just get rid of this for now and use `WriteFactsStringToFactsFile` to dump the rules directly.

src/backends/policy_engine/souffle/souffle_policy_checker.cc
```diff
@@ -18,25 +18,59 @@
...
+using DatalogLoweringVisitor = souffle::DatalogLoweringVisitor;
+
 static constexpr char kDatalogPolicyEngine[] = "datalog_policy_verifier_cxx";
 static constexpr char kViolatesPolicyRelation[] = "violatesPolicy";
 
-bool SoufflePolicyChecker::IsModulePolicyCompliant(const ir::Module& module,
-                                                   const Policy& policy) const {
-  std::unique_ptr<souffle::SouffleProgram> program(CHECK_NOTNULL(
-      souffle::ProgramFactory::newInstance(std::string(kDatalogPolicyEngine))));
+static bool IsModulePolicyCompliantInner(
```
bgogul, line 40: nit: `IsModulePolicyCompliantInner` -> `IsModulePolicyCompliantHelper`

src/backends/policy_engine/souffle/souffle_policy_checker.cc
```diff
@@ -18,25 +18,59 @@
...
+    const ir::Module& module, const Policy& policy,
+    std::filesystem::path facts_directory) {
```
bgogul, line 42: `const std::filesystem::path& facts_directory`?

src/backends/policy_engine/souffle/datalog_fact_policy.h
```diff
@@ -0,0 +1,54 @@
...
+  absl::Status OutputAsDlFactsToDirectory(
+      const std::filesystem::path &output_path) const override {
+    return WriteFactsStringToFactsFile(output_path, relation_name_,
+                                       facts_string_);
+  }
+
+ private:
+  std::string relation_name_;
+  std::string facts_string_;
+};
```
bgogul, line 44: Haven't thought this through, but I don't know if this is right abstraction of policy for the long run. Let us get rid of this because we don't need this right now?

If you want to keep this class, then make it simple and move the logic in OutputAsDlFactsToDirectory back to souffle_policy_checker?

src/backends/policy_engine/souffle/raksha_datalog_facts_test.cc
```diff
@@ -52,6 +54,18 @@ class DumpFactsToDirectoryTest : public testing::Test {
...
+
+  // Usually, the test fixture's destructor is fine. But I'm pretty sure that
+  // `remove_all` can throw an exception in general, which would cause undefined
+  // behavior if we put this in the destructor. This probably isn't possible
+  // with the noexception flags that we're using, but when it comes to undefined
+  // behavior in dtors, better safe than sorry.
+  void TearDown() override {
+    if (!directory_.empty()) {
+      std::filesystem::remove_all(directory_);
+    }
```
bgogul, line 68: If it is temporary directory, do we really need to clean it up explicitly?

src/backends/policy_engine/souffle/utils.cc
```diff
@@ -0,0 +1,47 @@
...
+
+#include "src/backends/policy_engine/souffle/utils.h"
+
+#include <fstream>
+
+#include "absl/strings/str_format.h"
+
+namespace raksha::backends::policy_engine::souffle {
+
+absl::Status WriteFactsStringToFactsFile(
```
bgogul, line 25: If you add an API, we should add tests too. This could probably stay as a static method of RakshaDatalogFacts?

src/backends/policy_engine/souffle/souffle_policy_checker.cc
```diff
@@ -18,25 +18,59 @@
...
+  absl::Status output_policy_status =
+      policy.OutputAsDlFactsToDirectory(facts_directory);
+  CHECK(output_policy_status.ok())
+      << "Unexpected error while outputting policy: " << output_policy_status;
```
bgogul, line 46: Why don't we just use `WriteFactsStringToFactsFile` to dump the `isSqlPolicyRule` string passed in through the swig interface?

src/frontends/sql/driver.cc
```diff
@@ -0,0 +1,36 @@
...
+#include "src/frontends/sql/driver.h"
+
+#include "src/frontends/sql/decode.h"
+#include "src/frontends/sql/sql_ir.pb.h"
+#include "src/backends/policy_engine/souffle/souffle_policy_checker.h"
+
+namespace raksha::frontends::sql {
+
+bool verify(const ExpressionArena &arena,
+            backends::policy_engine::souffle::DatalogFactPolicy policy) {
```
bgogul, line 26: Aren't you passing in a `string` instead of policy from Swig?

src/common/utils/filesystem.cc
```diff
@@ -0,0 +1,43 @@
...
+  std::error_code error_code;
+  std::filesystem::path temp_path =
+      std::filesystem::temp_directory_path(error_code);
+  CHECK(!error_code) << absl::StrFormat(
+      "Unable to access temporary file directory: `%s`", error_code.message());
+
+  absl::string_view temp_file = std::tmpnam(nullptr);
+  std::filesystem::path new_temp_path = temp_path / temp_file;
+  if (!std::filesystem::create_directory(new_temp_path, error_code)) {
+    LOG(FATAL) << absl::StrFormat("Unable to create directory `%s`: `%s`",
```
bgogul, line 37: Why not return `absl::StatusOr` as it used to be?  Given that this could run on an android app, we should gracefully fail. We cannot have `FATAL` failures if we can help it.

That is also why I tried to use non-exception methods from std::filesystem.

bgogul (commented):
Left some comments. Changes like having a `LOG(FATAL)` in the code path of an Android app is a no-no. I was thinking that if there was a failure like this, we should return false for policy compliance back, which is safer than crashing the app.

src/backends/policy_engine/souffle/raksha_datalog_facts_test.cc
markww: Yes, because it isn't temporary in the sense of "will be cleaned up when all references to it expire", it's more likely temporary in the sense of "will be cleaned up when the machine is powered off". Garbage can accumulate in such directories surprisingly quickly, especially for devices that people don't think to turn off frequently (like phones).

src/backends/policy_engine/souffle/datalog_fact_policy.h
markww: Got rid of this class.

src/backends/policy_engine/souffle/datalog_fact_policy.h
markww: Got rid of this, put the string directly in `Policy` for now.

src/frontends/sql/driver.cc
markww: Yes, but it is likely that the long-term direction will be to instead pass a `Policy` in. I was thinking it would feel most correct to construct a `Policy` in the bridge from the policy string for now.

src/backends/policy_engine/souffle/souffle_policy_checker.cc
markww: Done.

src/backends/policy_engine/souffle/souffle_policy_checker.cc
markww: Fixed.

src/common/utils/filesystem.cc
markww: Fixed `CreateTemporaryDirectory` to return an `absl::StatusOr` instead of failing in a `FATAL` fashion.

src/backends/policy_engine/souffle/utils.cc
markww: The policy rules aren't part of `RakshaDatalogFacts`, so in order to share code like this I have to make this file public. So I guess I'll write some tests for it.

markww: Wrote tests for this function.

src/backends/policy_engine/souffle/souffle_policy_checker.cc
markww: I don't 100% understand this comment. I've refactored things a bit so that the `isSqlPolicyRule` facts are a field on the `Policy` object and we are using `WriteFactsStringToFactsFile` to put them into a `.facts` file in the facts directory. Does that satisfy this comment?

bgogul (approved):

PR #562, opened 2022-05-23
Add a proto representation of the Raksha IR.

markww:
This PR adds a representation that directly captures the Raksha IR.
Unlike our previous attempt at creating a SQL-specific IR proto, this
proto maps directly to the Raksha IR structures, and does not decide on
the set of allowed operations in advance. Instead, it defers to the
frontend to provide the operations that that frontend is interested in.

In an attempt to future-proof the IR, this introduces one structure for
which there is no analog in the current IR: an `IrTranslationUnit`
message. The `IrTranslationUnit` contains a single top-level module
plus some additional fields to annotate that module (currently, just the
name of the frontend from which the IR originated). While not extremely
useful at the moment, this top-level structure gives us a place to put
things relevant to the analysis on the whole TU rather than any
individual structure. These things may include analysis tuning or
checker-selecting options, the commit hash or version string of the
frontend, the person or entity requesting the analysis, etc.

### DIFF
--- a/src/ir/proto/raksha_ir.proto
+++ b/src/ir/proto/raksha_ir.proto
@@ -51,9 +51,9 @@ message NamedValueMap {
 
 message Operation {
   string operator = 1;
-  NamedValueMap results = 2;
+  repeated Value inputs = 2;
   NamedAttributeMap attributes = 3;
-  Module impl_module = 4;
+  NamedValueMap results = 4;
 }
 
 message AnyValue {}
--- a/src/ir/proto/raksha_ir.proto
+++ b/src/ir/proto/raksha_ir.proto
@@ -1,0 +1,101 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+// This proto file contains the proto representation of the Raksha IR. This
+// allows frontends to construct the Raksha IR in any language and send the IR
+// over the wire or across an FFI layer for analysis.
+
+syntax = "proto3";
+
+package raksha.ir.proto;
+
+message Module {
+  map<uint64, Block> blocks = 1;
+}
+
+message Block {
+  // A mapping from the ID for an operation in the block to that operation.
+  // This allows values produced by this `Operation` to "point" to the
+  // `Operation` that produced them within the block.
+  map<uint64, Operation> operations = 1;
+  NamedValueMap results = 2;
+}
+
+message AttributePayload {
+  oneof AttributePayloadKind {
+    int64 int64_payload = 1;
+    string string_payload = 2;
+  }
+}
+
+message NamedAttributeMap {
+  map<string, AttributePayload> attributes = 1;
+}
+
+message NamedValueMap {
+  map<string, Value> values = 1;
+}
+
+message Operation {
+  string operator = 1;
+  NamedValueMap results = 2;
+  NamedAttributeMap attributes = 3;
+  Module impl_module = 4;
+}
+
+message AnyValue {}
+
+message OperationResultValue {
+  uint64 operation_id = 1;
+  string output_name = 2;
+}
+
+message BlockArgumentValue {
+  uint64 block_id = 1;
+  string block_argument_name = 2;
+}
+
+message Value {
+  oneof ValueKind {
+    AnyValue any_value = 1;
+    OperationResultValue operation_result_value = 2;
+    BlockArgumentValue block_argument_value = 3;
+  }
+}
+
+// The top level IR message. Contains the top level module, plus some
+// bookkeeping that may help with diagnosing issues.
+message IrTranslationUnit {
+  // The top level module to be analyzed. This is the only field that's strictly
+  // necessary for an analysis to occur.
+  Module top_level_module = 1;
+  // The frontend that produced these IR messages.
+  string frontend = 2;
+}
+
+// A status indicating whether the analysis succeeded, failed, or encountered an
+// internal error.
+enum AnalysisResultStatus {
+  AR_UNKNOWN = 0;
+  AR_SUCCESS = 1;
+  AR_FAILURE = 2;
+  AR_INTERNAL_ERROR = 3;
+}
+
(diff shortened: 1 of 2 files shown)


src/ir/proto/raksha_ir.proto
```diff
@@ -0,0 +1,101 @@
...
+
+message NamedValueMap {
+  map<string, Value> values = 1;
+}
+
+message Operation {
+  string operator = 1;
+  NamedValueMap results = 2;
+  NamedAttributeMap attributes = 3;
+  Module impl_module = 4;
```
bgogul, line 56: I have been debating about whether we should support more than one module. We could cover our use cases (like `if`) by having multiple blocks.  However, I am wondering if we should make this a repeated field in the proto.

src/ir/proto/raksha_ir.proto
```diff
@@ -0,0 +1,101 @@
...
+
+message NamedAttributeMap {
+  map<string, AttributePayload> attributes = 1;
+}
+
+message NamedValueMap {
+  map<string, Value> values = 1;
+}
+
+message Operation {
```
bgogul, line 52: Why is there no field for arguments of an operation?

src/ir/proto/raksha_ir.proto
```diff
@@ -0,0 +1,101 @@
...
+  string operator = 1;
```
bgogul, line 53: Create an `Operator` message and make this a field of `Operator` type. I expect we will add some additional info (like number of inputs, is_variadic, etc.)

src/ir/proto/raksha_ir.proto
```diff
@@ -0,0 +1,101 @@
...
+  string output_name = 2;
+}
+
+message BlockArgumentValue {
+  uint64 block_id = 1;
+  string block_argument_name = 2;
+}
+
+message Value {
+  oneof ValueKind {
```
bgogul, line 72: I see that you have not added storage. Let us keep it that way. It is easier to add a new kind later right?

src/ir/proto/raksha_ir.proto
```diff
@@ -0,0 +1,101 @@
...
+// This proto file contains the proto representation of the Raksha IR. This
+// allows frontends to construct the Raksha IR in any language and send the IR
+// over the wire or across an FFI layer for analysis.
+
+syntax = "proto3";
+
+package raksha.ir.proto;
+
+message Module {
+  map<uint64, Block> blocks = 1;
```
bgogul, line 26: Why not just a repeated field?

src/ir/proto/raksha_ir.proto
```diff
@@ -0,0 +1,101 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+// This proto file contains the proto representation of the Raksha IR. This
+// allows frontends to construct the Raksha IR in any language and send the IR
+// over the wire or across an FFI layer for analysis.
+
+syntax = "proto3";
```
bgogul, line 21: Do you plan to write the encoding and decoding in a separate PR?

src/ir/proto/raksha_ir.proto
markww: We aren't using `impl_module` at all right now, so what I might do is just delete this altogether. That will allow us to add it when we have a clearer idea of what form we want it to take.

bgogul: Even better.

src/ir/proto/raksha_ir.proto
markww: I was thinking of going a different way with it: if we were to add more fields to `Operator`, it is highly likely that we would want those additional fields to be the same across all uses of the `Operator` (ie, every `+` operator has 2 inputs, is not variadic, etc). As such, I was thinking that we could, in the TU-level structure, have a list of `FullOperatorInfo` (or maybe just `Operator`) providing this additional info in a map keyed on the operator name.

I suppose that if we want to do that, the best thing to do would be to rename `operator` here to `operator_name` to make it clear that this isn't the whole `Operator` concept, but just the name piece of it.

src/ir/proto/raksha_ir.proto
markww: Yes, intentionally omitted `Storage` because it's probably going away. And yes, it's easy to add things to `proto`s, harder to (fully) remove them.

src/ir/proto/raksha_ir.proto
markww: For the `Value` `BlockArgument`, we need to be able to specify the `Block` that the arguments are from. However, this `Value` shouldn't own the `Block`; the module that the `Block` resides in should own the `Block`. Therefore, we need some way to point at the `Block`. For that reason, I'm assigning an ID to each `Block`.

Of course, a `repeated` field can be thought of as a map from `index` to value, so we could provide the "map" that way. But I think it might be wise to not impose a density requirement unnecessarily, as that provides frontends the freedom to drop blocks. This might be useful, for instance, in the case of a TU containing errors where the frontend was still able to correctly parse some portions of the TU.

src/ir/proto/raksha_ir.proto
markww: Yes. Just wanted to get this up to get some feedback on the structure of the protos before I started writing some C++. Also, protos tend to be harder to import into Google3, so wanted to round trip this file first.

src/ir/proto/raksha_ir.proto
markww: Oversight, I thought I had added it. Let me put that in now.

src/ir/proto/raksha_ir.proto
markww: Done.

src/ir/proto/raksha_ir.proto
markww: Done.

src/ir/proto/raksha_ir.proto
bgogul: Having an id sounds good. Is this going to be global id across all the blocks in the ir? Or, it is just scoped to the module? I assume the former? Add a brief comment like the one for `operations` in `Block`.

bgogul (approved):

src/ir/proto/raksha_ir.proto
bgogul: `operator` -> `operator_name` with a brief comment sounds good.

src/ir/proto/raksha_ir.proto
markww: To be honest, I wasn't 100% sure, but `uint64` should be enough for either. If you think an ID unique across the TU is the right call, that probably is the way to go. I'll add a comment explaining this.

src/ir/proto/raksha_ir.proto
markww: Done.

src/ir/proto/raksha_ir.proto
markww: Done.

PR #563, opened 2022-05-24
Allow additional annotations to be added for operations in dot output.

bgogul:
The `DotGeneratorConfig` is narrowly focussed so that we have something to build on. We might have to rethink this when we have more experience with what we want.

### DIFF
--- a/src/ir/dot/dot_generator.cc
+++ b/src/ir/dot/dot_generator.cc
@@ -124,6 +124,17 @@ void DotGeneratorHelper::AddResult(const Operation& op,
   }
 }
 
+namespace {
+
+// Safe initialization of static empty btree set.
+const absl::btree_set<std::string>& GetEmptyStringBtreeSet() {
+  static const absl::btree_set<std::string>* empty_set =
+      new absl::btree_set<std::string>();
+  return *empty_set;
+}
+
+}  // namespace
+
 std::string DotGeneratorHelper::GetDotNode(const Operation& op) {
   // Generates a dot representation of an operation as a node. Suppose that the
   // operation has `m` inputs and `n`results, the generated node will be
@@ -139,7 +150,6 @@ std::string DotGeneratorHelper::GetDotNode(const Operation& op) {
   //
   // The `Ix` and `Rx` are ports that we can connect edges to.
   // (cf. https://www.graphviz.org/doc/info/shapes.html#record)
-  static absl::btree_set<std::string> empty_results_set;
   std::string node_name = GetNodeName(op);
   std::string input_ports = absl::StrJoin(
       op.inputs(), "|", [i = 0](std::string* out, const Value& v) mutable {
@@ -149,7 +159,7 @@ std::string DotGeneratorHelper::GetDotNode(const Operation& op) {
       });
   auto find_result = operation_results_.find(node_name);
   const absl::btree_set<std::string>& results =
-      (find_result == operation_results_.end()) ? empty_results_set
+      (find_result == operation_results_.end()) ? GetEmptyStringBtreeSet()
                                                 : find_result->second;
   std::string output_ports =
       results.empty()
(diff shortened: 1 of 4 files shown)


src/ir/dot/dot_generator_test.cc
```diff
@@ -235,6 +237,38 @@ TEST_F(DotGeneratorTest, AddsEdgeFromOperationResultToInputs) {
...
+
+  EXPECT_THAT(
+      GetNodes(dot_graph_lines),
+      testing::UnorderedElementsAre(
+          // block b0
+          "B0",
+          // %0 = core.pair []()
+          R"(B0_0[label="{{}|core.pair|core.pair-2|{<first>first|<second>second}}"])",
+          // %1 = core.plus [](%0.first, %1.second)
+          R"(B0_1[label="{{<I0>I0|<I1>I1}|core.plus|core.plus-0|{<out>out}}"])"));
```
harsha-mandadi, line 269: If the result of an operation is a valueList, the usage of `%0.first` will be replaced with `%0`. Specifically, there will no longer be a value `ID.ID`. Is my understanding correct??

bgogul: That is correct.

harsha-mandadi: ok. Is the DotGeneratorConfig in this case provides additional label information of size of the operation result corresponding to the operator name? It was not obvious to me by reading the code. Does it make sense to add comments?

bgogul: Yes, it adds `core.plus-0` this text to the label. I will add a comment.

bgogul: Done, I also changed the format to `numResults(<op-name>): <num-results>`. It hopefully makes the additional annotation more obvious.

bgogul:
@harsha-mandadi PTAL. I addressed your comments.

harsha-mandadi (approved):

src/ir/dot/dot_generator.cc
```diff
@@ -136,6 +139,7 @@ std::string DotGeneratorHelper::GetDotNode(const Operation& op) {
   //
   // The `Ix` and `Rx` are ports that we can connect edges to.
   // (cf. https://www.graphviz.org/doc/info/shapes.html#record)
+  static absl::btree_set<std::string> empty_results_set;
```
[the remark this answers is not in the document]
bgogul: Good point. I will make it const. I don't know if we can make it constexpr. This is mainly needed to support the `operation_labeler` interface. Once we have the results directly in operation, this will go away. However, it is needed for now.

src/ir/dot/dot_generator.cc
```diff
@@ -144,14 +148,20 @@ std::string DotGeneratorHelper::GetDotNode(const Operation& op) {
                         absl::StrFormat(R"(<I%d>I%d)", node_num, node_num));
       });
   auto find_result = operation_results_.find(node_name);
+  const absl::btree_set<std::string>& results =
+      (find_result == operation_results_.end()) ? empty_results_set
+                                                : find_result->second;
```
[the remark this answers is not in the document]
bgogul: I am using a btree_set to get a predicable output. The alternative was to keep a `flat_hash_set` or `vector` and later sort it. However, shouldn't the asymptotic complexity be the same for both uses?

src/ir/dot/dot_generator.cc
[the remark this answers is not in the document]
bgogul: Fixed this using a recommended idiom.

bgogul:
@markww PTAL. I addressed your comments.

src/ir/dot/dot_generator.cc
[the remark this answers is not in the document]
markww: The asymptotic complexity is the same, but often in C++, the performance penalty for using an online-sorting datastructure like a btree is worse than you might expect from complexity alone due to the less dense layout in memory (and thus additional cache misses) than something like a `std::vector` or an `absl::flat_hash_set` would provide. I usually try to nudge towards the dense structures at the time when a sorting structure is introduced. But since it's already in, and this isn't exactly a performance-critical piece of code, I think this is probably fine.

markww (approved):
LGTM

PR #568, opened 2022-05-31
Add changes to make the IR parser build pass on winterrowd@'s machine.

markww:
My build was failing because the IR parser test did not have
`-fexceptions`, but the ANTLR-generated code wanted to throw exceptions
in some cases. I don't think I did anything to expose this. This may
just be another GCC/Clang difference showing up for me.


harsha-mandadi (approved):
LGTM

PR #570, opened 2022-06-01
parser for block and module in ir

harsha-mandadi:


src/ir/value.h
```diff
@@ -50,8 +50,7 @@ class NamedValue {
   const T& element() const { return *element_; }
   absl::string_view name() const { return name_; }
   std::string ToString(SsaNames& ssa_names) const {
-    return absl::StrFormat("%%%d.%s", ssa_names.GetOrCreateID(*element_),
-                           name_);
+    return absl::StrFormat("%s", name_);
```
bgogul, line 53: Note that we are losing some critical information about the block or operation, which is needed. 

Why change this now? Can't we leave it as is? If you leave this unchanged, you don't have to change the tests in this PR also.

harsha-mandadi: `absl::StrFormat("%%%d.%s", ssa_names.GetOrCreateID(*element_), name_` this would cause parser to print out `%0.0 = core.minus [const: 10](arg0, arg1)\n"}))` instead of `%0 = core.minus [const: 10](arg0, arg1)\n"}))`. Would it make more sense to make this to keep ssa_names and remove name_. Like this `return absl::StrFormat("%%%d.%s", ssa_names.GetOrCreateID(*element_)`

src/parser/ir/ir_parser.cc
```diff
@@ -98,34 +100,51 @@ const Operation& IrProgramParser::ConstructOperation(
...
-                  CHECK(find_value_id_operation == operation_map_.end());
-                  if (find_value_id_operation != operation_map_.end()) {
-                    return Value(value::OperationResult(
-                        *find_value_id_operation->second, value_id));
+                  auto find_value_id_operation = value_map_.find(value_id);
+                  if (find_value_id_operation != value_map_.end()) {
+                    return find_value_id_operation->second;
                   }
                   // Storage Value
+                  std::cout << "Storage" << std::endl;
```
bgogul, line 108: stale `cout` statements?

src/parser/ir/ir_parser.cc
```diff
@@ -98,34 +100,51 @@ const Operation& IrProgramParser::ConstructOperation(
...
                       CHECK_NOTNULL(named_value_context)->VALUE_ID()->getText();
                   // Operation Result Value
-                  auto find_value_id_operation = operation_map_.find(value_id);
-                  // check in place as we do not support operation lists yet.
-                  CHECK(find_value_id_operation == operation_map_.end());
-                  if (find_value_id_operation != operation_map_.end()) {
-                    return Value(value::OperationResult(
-                        *find_value_id_operation->second, value_id));
+                  auto find_value_id_operation = value_map_.find(value_id);
+                  if (find_value_id_operation != value_map_.end()) {
```
bgogul, line 104: Actually change this to a check and get rid of the storage case altogether. We plan to deprecate storage anyway. 

(Do the tests have coverage for storage?)

Also, add a death test for the case where a value will not be found.

src/parser/ir/ir_parser_test.cc
```diff
@@ -30,10 +30,33 @@ TEST_P(IrParserTest, SimpleTestOperation) {
...
+    %1 = core.plus [access: private, transform: no](<<ANY>>, <<ANY>>)
+  }  // block b1
+}  // module m0
+)",
+R"(module m0 {
+  block b0 {
+    %0 = core.plus []()
+  }  // block b0
+  block b1 {
+    %1 = core.plus [access: private, transform: no](%0.out, <<ANY>>)
```
bgogul, line 57: Just a clarification. Once you fix ssa names, `%0.out` will simply be `%0`, right?

src/parser/ir/ir_parser.h
```diff
@@ -37,16 +38,21 @@ using ir_parser_generator::IrParser;
...
+                          BlockBuilder& block_builder);
+  void ConstructBlock(IrParser::BlockContext& block_context);
+  void ConstructModule(IrParser::ModuleContext& module_context);
+  // As it parses through the program, value_map inserts new entity for
+  // each operation Ex: %0 = core.minus []() inserts a new entity {"%0",
   // Operation(nullptr, Operator(core.minus), ..., nullptr )}
-  absl::flat_hash_map<std::string, std::unique_ptr<Operation>> operation_map_;
+  absl::flat_hash_map<std::string, Value> value_map_;
+  absl::flat_hash_map<std::string, const Block*> block_map_;
+  SsaNames ssa_names_;
```
bgogul, line 53: It is not used right now. Delete it?

src/parser/ir/ir_parser_test.cc
harsha-mandadi: As per my understanding, fixing ssa_names is not enough. `%0.out` here is WA. https://github.com/google-research/raksha/blob/63c0edce07cd17a23f39881911a5e2279ab96397/src/ir/value.h#L53. This part of ToString needs to be changed to print only `%0`.

src/parser/ir/ir_parser.cc
```diff
@@ -98,34 +100,44 @@ const Operation& IrProgramParser::ConstructOperation(
...
-                  if (find_value_id_operation != operation_map_.end()) {
-                    return Value(value::OperationResult(
-                        *find_value_id_operation->second, value_id));
-                  }
-                  // Storage Value
-                  types::TypeFactory type_factory;
-                  return Value(value::StoredValue(
-                      Storage(value_id, type_factory.MakePrimitiveType())));
+                  auto find_value_id_operation = value_map_.find(value_id);
+                  CHECK(find_value_id_operation != value_map_.end()) << "Value must be found";
```
bgogul, line 104: Worth adding the missing value_id to the error message:
`"Value must be found: " << value_id`;

(You won't have to change the death test, I think.)

Also nit: "Value not found" might be a better message?

src/parser/ir/ir_parser_test.cc
bgogul: Got it.

src/parser/ir/ir_parser_test.cc
```diff
@@ -30,10 +30,47 @@ TEST_P(IrParserTest, SimpleTestOperation) {
             input_program_text);
 }
 
-INSTANTIATE_TEST_SUITE_P(
-    IrParserTest, IrParserTest,
-    ::testing::Values(
-        "%0 = core.plus []()\n",
-        "%0 = core.minus [access: private, send: yes](<<ANY>>, <<ANY>>)\n"));
+INSTANTIATE_TEST_SUITE_P(IrParserTest, IrParserTest,
```
bgogul, line 33: Add a test case with multiple operations in a block?

bgogul (approved):
Looks good modulo a missing test.

PR #573, opened 2022-06-07
Add "magic" tag xforms to change itag propagation behavior on merges.

markww:
While integrating Raksha into the SQL verifier that we are targeting, we
discovered that the SQL protos that we created are factored wrongly, and
in the future we would like frontends to target protos that mirror the
structure of the IR directly. However, to complete the integration of
Raksha and the SQL frontend, we discovered that we need to support merge
operations that have different ways of propagating integrity tags. To do
this the right way would involve modifying the SQL IR protos, which
would be a lot of work on something we now know we want to throw away.

This commit introduces a hack to get us the rest of the way to
supporting the SQL verifier using the current SQL IR protos without
incurring the code changes needed to update the SQL IR protos. We
introduce some specially-named TagTransform operations that modify the
integrity tag propagation behavior of the merge operations that they
wrap. While not ideal, this allows us to get the behavior we want with
the current SQL IR structure.


src/analysis/souffle/tag_transforms.dl
```diff
@@ -153,6 +155,84 @@ mustHaveIntegrityTag(result, DEFAULT_SQL_OWNER, iTag) :-
...
 
+// In the SQL verifier, we have some merge operations that we would like to
+// propagate integrity tags in a non-default fashion (ie, rather than dropping
+// the integrity tags, we'd like them to intersect or union across all direct
+// inputs). To do that, we introduce a tag transform with special behavior:
+// rather than linking to a `SqlPolicyRule` fact, they shall instead link to
+// built-in rules here that modify how the integrity tags propagate across the
+// wrapped merge operation.
+
+#define RAKSHA_INTERNAL_INTERSECT_ITAG_RULE_NAME "##raksha_internal#intersect_itag_rule"
```
bgogul, line 166: Are the `##` used to avoid collisions?

src/analysis/souffle/tag_transforms.dl
```diff
@@ -153,6 +155,84 @@ mustHaveIntegrityTag(result, DEFAULT_SQL_OWNER, iTag) :-
...
+
+.decl isMergeOperationDirectInput(mergeOp: Operation, ap: AccessPath)
+isMergeOperationDirectInput(mergeOp, inputPath) :-
+  isMergeOp(mergeOp), isAccessPath(inputPath), operationHasOperandAtIndex(mergeOp, inputPath, idx),
+  mergeOperationHasControlInputStartIndex(mergeOp, controlStartIdx),
+  idx < controlStartIdx.
+
+// For the union rule, any itag on a direct input should propagate to the
+// output.
+mustHaveIntegrityTag(result, DEFAULT_SQL_OWNER, iTag) :-
```
bgogul, line 199: Just an observation and question. Result would have certain integrity either because of the default propagation rule or because of this special rule.

I guess if we want to have a special rule for `mayHaveTag`, we will need to do something such as excluding the default propagation rule from triggering?

bgogul (approved):

src/analysis/souffle/tag_transforms.dl
markww: Yes, I just added them to make collisions less likely.

src/analysis/souffle/tag_transforms.dl
markww: The current default behavior is that we draw edges from all inputs to all outputs, but also draw an edge from a fake access path with no integrity tags. This will propagate all confidentiality tags, but block the default propagation of all integrity tags. As such, I think we're doing OK here. These rules act as a sort of generic tag propagation rule overriding the strict default propagation behavior.

PR #575, opened 2022-06-07
Refactor relations on `Operation`s to allow more concise matching.

markww:
This PR adds relations used to get particular fields of an `Operation`
so `Operation`s don't have to be fully deconstructed when a small number
of fields are required. Also updates some rules to use this mechanism
where it would make them more concise and eliminate some duplicate
rules.

### DIFF
--- a/src/analysis/souffle/operations.dl
+++ b/src/analysis/souffle/operations.dl
@@ -35,7 +35,34 @@
 // Operation universe.
 .decl isOperation(operation: Operation)
 
-isPrincipal(owner), isAccessPath(result) :- isOperation([owner, _, result, _, _]).
+// OperandList universe.
+.decl isOperandList(operandList: OperandList)
+
+// Operator universe.
+.decl isOperator(operator: Operator)
+
+// Attribute list universe.
+.decl isAttributeList(attributeList: AttributeList)
+
+// Fill the appropriate universes with the pieces of the operation.
+isPrincipal(owner), isOperator(operator), isAccessPath(result),
+isOperandList(operandList), isAttributeList(attributeList) :-
+isOperation([owner, operator, result, operandList, attributeList]).
+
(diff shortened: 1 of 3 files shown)


src/analysis/souffle/operations.dl
```diff
@@ -35,7 +35,34 @@
...
+.decl operationHasOperator(operation: Operation, operator: Operator)
+.decl operationHasOperandList(operation: Operation, operandList: OperandList)
+.decl operationHasAttributeList(operation: Operation, attributeList: AttributeList)
+
+// Populate field association relations.
+operationHasResult(operation, result), operationHasOwner(operation, owner),
+operationHasOperator(operation, operator), operationHasOperandList(operation, operands),
+operationHasAttributeList(operation, attributes) :-
+  isPrincipal(owner), isOperator(operator), isAccessPath(result), isOperandList(operands),
+  isAttributeList(attributes), isOperation(operation),
```
bgogul, line 64: Is this needed?

bgogul (approved):
Do we have tests? Approving modulo tests.

src/analysis/souffle/operations.dl
markww: Yes, this is a multi-head rule that deconstructs all of the fields of the `operation` to populate the various field-access relations.

markww: I realized your question might be "are all of these universe groundings needed?" The answer to that is also yes:  the line `operation = [owner, operator, result, operands, attributes]` is not deconstructing the `operation` the way you might expect a pattern matching operation to do so; it is instead literally saying that there exists an `operation` that is equal to the record `[owner, operator, result, operands, attributes]`. It has the *effect* of pattern matching, but is not handled in a nice way, so we have to ground the members like this.

markww:
Added a unit test.

PR #581, opened 2022-06-15
Add policy compliance check binary.

bgogul:


bgogul:
@asejfia FYI.

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  std::ifstream file_stream(file_path);
+  if (!file_stream) {
+    return absl::FailedPreconditionError(absl::StrCat(
+        "Unable to read file '", file_path.string(), "': ", strerror(errno)));
+  }
+
+  std::ostringstream string_stream;
+  string_stream << file_stream.rdbuf();
```
markww, line 55: I don't think that this is what you want. `rdbuf` just returns the underlying buffer for the file stream; I don't think this is guaranteed to contain the whole file (or, indeed, anything at all).

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -0,0 +1,100 @@
...
+  SoufflePolicyChecker checker;
+  IrProgramParser ir_parser;
+  const Module& module = ir_parser.ParseProgram(*ir_string);
+  Policy policy(*sql_policy_rules);
+  if (checker.IsModulePolicyCompliant(module, policy)) {
+    std::cerr << "Policy check succeeded!";
+    return 0;
+  } else {
+    std::cerr << "Policy check failed!";
+    return 1;
```
markww, line 98: Shouldn't this be `LOG(ERROR)` as above?

src/backends/policy_engine/souffle/check_policy_compliance_test.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
```
markww, line 2: Should we have a copyright header here? Also maybe a quick comment describing what this file does?

src/backends/policy_engine/souffle/check_policy_compliance_test.sh
```diff
@@ -0,0 +1,49 @@
...
+IR_FILE=${ROOT_DIR}/${IR_FILE_ARG}
+SQL_POLICY_RULES_FILE=${ROOT_DIR}/${SQL_POLICY_RULES_FILE_ARG}
+
+$CMD --ir=$IR_FILE --sql_policy_rules=$SQL_POLICY_RULES_FILE
+if [ $? -eq 0 ]; then
+  ACTUAL_RESULT="${PASS_RESULT}"
+else 
+  ACTUAL_RESULT="${FAIL_RESULT}"
+fi
+if [ "${ACTUAL_RESULT}" != ${EXPECTATION_ARG} ]; then
```
markww, line 43: Should `${EXPECTATION_ARG}` be surrounded in quotes here?

src/backends/policy_engine/souffle/check_policy_compliance.cc
bgogul: the .str() below converts the entire stream as string?

src/backends/policy_engine/souffle/check_policy_compliance_test.sh
bgogul: done

bgogul:
@markww PTAL.

src/backends/policy_engine/souffle/check_policy_compliance.cc
markww: Yes, the `.str()` is correctly converting the contents of the `std::ostringstream string_stream` to a `std::string`, and is capturing the entire contents of `string_stream. My concern, however, is about the line:

```C++
string_stream << file_stream.rdbuf();
```

`file_stream.rdbuf()` just exposes the current contents of the file stream's underlying read buffer. This is just a view into the portion of the file that the stream has currently in memory. It need not be the entire file. Depending on how the filestream is implemented or the conditions on the filesystem when it is opened, I don't think it is even guaranteed to be non-empty for a non-empty file.

bgogul: See version 10 in https://en.cppreference.com/w/cpp/io/basic_ostream/operator_ltlt

Seems like it will do the right thing?

markww: Oh, wow, TIL. Thanks for the link, this looks good then.

markww (approved):
LGTM

src/backends/policy_engine/souffle/check_policy_compliance.cc
bgogul: I added a small comment to capture the discussion here.

PR #590, opened 2022-06-23
modify ssa_names and update ir_parser tests

harsha-mandadi:


src/ir/dot/dot_generator.cc
```diff
@@ -163,43 +163,45 @@ std::string DotGeneratorHelper::GetDotNode(const Operation& op) {
...
           ? absl::StrFormat(R"(<TD COLSPAN="%d">&nbsp;</TD>)", total_colspan)
           : absl::StrJoin(
-              op.inputs(), "", [i = 0, input_colspan](std::string* out, const Value& v) mutable {
-                int node_num = i++;
-                absl::StrAppend(out,
-                    absl::StrFormat(
-                        R"(<TD COLSPAN="%d" PORT="I%d">I%d</TD>)", 
-                        input_colspan, node_num, node_num));
-              });
+                op.inputs(), "",
```
harsha-mandadi, line 170: Rest of the changes below in this file are format changes on save!

src/ir/block_builder_test.cc
```diff
@@ -190,10 +191,13 @@ TEST_F(BlockBuilderTest, AddImplementationMakingMultipleUpdates) {
   const NamedValueMap& results = block->results();
   SsaNames ssa_names;
   ASSERT_EQ(results.count("entity_output"), 1);
-  EXPECT_EQ(results.at("entity_output").ToString(ssa_names), "%0.entity_value");
+  //%0 is the entity_value OperationResult Value :
```
bgogul, line 194: It is not guaranteed that  `Value(value::OperationResult(op, "entity_value"))` will always be `%0`. There is a similar concern with the other value. 

Why don't you populate `ssa_names` so thaValue(value::OperationResult(op, "entity_value"))` maps to `"entity_value"` and Value(value::OperationResult(op, "primitive_value"))` maps to  
 `"primitive_value"`. 
 
Then the test will not be fragile.

src/ir/dot/dot_generator.cc
```diff
@@ -57,14 +57,14 @@ class DotGeneratorHelper : public IRTraversingVisitor<DotGeneratorHelper> {
...
-    return absl::StrFormat("%s_%d",
+    return absl::StrFormat("%s_%s",
                            block == nullptr ? "g" : GetNodeName(*block), id);
   }
 
   // Returns the dot node name for the given `block`.
   std::string GetNodeName(const Block& block) {
     auto id = ssa_names_.GetOrCreateID(block);
-    return absl::StrFormat("B%d", id);
+    return absl::StrFormat("B%s", id);
```
bgogul, line 67: We only added the prefix because the ids were numbers earlier. Also, to avoid issues in dot rendering, we will need to enclose it in quotes.

```suggestion
    return absl::StrFormat("\"%s\"", id);
```

The tests need to be updated as well.

src/ir/dot/dot_generator_test.cc
```diff
@@ -140,8 +142,8 @@ TEST_F(DotGeneratorTest, CreatesAttributesInNodesForOperation) {
   EXPECT_THAT(
       GetNodes(dot_graph_lines),
       testing::UnorderedElementsAre(
-          "B0 [shape=Mrecord]",
-          R"(B0_0 [label=<<TABLE BORDER="0" CELLBORDER="1" CELLSPACING="0"><TR><TD COLSPAN="1">&nbsp;</TD></TR><TR><TD COLSPAN="1">core.plus [string: something, ttl: 10] </TD></TR><TR><TD COLSPAN="1" PORT="out">out</TD></TR></TABLE>>])"));
+          "Bb0 [shape=Mrecord]",
+          R"(Bb0_%0 [label=<<TABLE BORDER="0" CELLBORDER="1" CELLSPACING="0"><TR><TD COLSPAN="1">&nbsp;</TD></TR><TR><TD COLSPAN="1">core.plus [string: something, ttl: 10] </TD></TR><TR><TD COLSPAN="1" PORT="out">out</TD></TR></TABLE>>])"));
```
bgogul, line 146: Unfortunately, the `%0` in  `Bb0_%0` will cause issues rendering the generated dot file. Can you enclose the name with quotes (`"`)?  (I have added suggestions at the relevant places.)

src/ir/dot/dot_generator.cc
```diff
@@ -57,14 +57,14 @@ class DotGeneratorHelper : public IRTraversingVisitor<DotGeneratorHelper> {
   std::string GetNodeName(const Operation& operation) {
     const Block* block = operation.parent();
     auto id = ssa_names_.GetOrCreateID(operation);
-    return absl::StrFormat("%s_%d",
+    return absl::StrFormat("%s_%s",
```
bgogul, line 60: Enclose the node name in quotes (`"`):
```suggestion
    return absl::StrFormat("\"%s_%s\"",
```

src/ir/ssa_names.h
```diff
@@ -17,51 +17,89 @@
...
+
+    ID GetOrCreateID(T entity) {
+      // Note that if `entity` is already in the map, the `insert` call below
+      // will return the existing entity. Otherwise, a new entity is added.
+      auto insert_result =
+          item_ids_.insert({entity, absl::StrCat(prefix_, item_ids_.size())});
+      return insert_result.first->second;
+    }
+
+    ID GetOrCreateID(T entity, ID id) {
```
bgogul, line 87: Can you change this to `void ReplaceID` (or `void UpdateID`) and make sure it is changed even if the entity exists? You can use `insert_or_assign` to update the value. e.g.,

```
void ReplaceID(T entity, ID id) { item_ids_.insert_or_assign(entity, id);  }
```


Also, we will need to add tests.

src/ir/value.cc
```diff
@@ -0,0 +1,43 @@
...
+#include <vector>
+
+#include "absl/strings/str_format.h"
+#include "src/common/utils/overloaded.h"
+#include "src/ir/ssa_names.h"
+
+namespace raksha::ir {
+
+std::string Value::ToString(SsaNames& ssa_names) const {
+  return std::visit(
```
bgogul, line 27: Didn't we decide that this change is not needed?

src/ir/value.h
```diff
@@ -107,6 +116,11 @@ class Any {
  public:
   std::string ToString(const SsaNames& ssa_names) const { return "<<ANY>>"; }
 
+  template <typename H>
+  friend H AbslHashValue(H h, const Any& v) {
```
bgogul, line 120: We will need to add tests for the hash values.  See https://abseil.io/docs/cpp/guides/hash for how to add tests.

src/ir/value_test.cc
```diff
@@ -118,14 +119,14 @@ TEST_P(ValueToStringTest, ToStringReturnsExpectedFormat) {
...
     testing::Values(
-        std::make_pair(Value(test_data.first_block_first_arg), "%0.one"),
-        std::make_pair(Value(test_data.first_block_second_arg), "%0.two"),
-        std::make_pair(Value(test_data.second_block_first_arg), "%1.one"),
-        std::make_pair(Value(test_data.second_block_second_arg), "%1.two"),
-        std::make_pair(Value(test_data.plus_operation_result1), "%0.res1"),
-        std::make_pair(Value(test_data.plus_operation_result2), "%0.res2"),
-        std::make_pair(Value(test_data.minus_operation_result1), "%1.res1"),
-        std::make_pair(Value(test_data.minus_operation_result2), "%1.res2"),
+        std::make_pair(Value(test_data.first_block_first_arg), "%0"),
```
bgogul, line 122: You should populate SSA names so that we don't rely on the order in which names were assigne.d 

Also, this is somewhat related to #596 . Will need to figure out how to refer to block arguments and results in the parser.

bgogul (commented):
Looks to be in good shape. I have suggested changes in a few places and also some tests are needed. 

Also, I wonder if  we should split the changes to SsaNames and Value to  separate CLs?

src/ir/dot/dot_generator.cc
markww: If you do this, you might want to make it a raw string to avoid the escaping if that makes things more readable.

src/parser/ir/ir_parser.h
```diff
@@ -50,8 +63,9 @@ class IrProgramParser {
   // Operation(nullptr, Operator(core.minus), ..., nullptr )}
   absl::flat_hash_map<std::string, Value> value_map_;
   absl::flat_hash_map<std::string, const Block*> block_map_;
-  IRContext context_;
-  Module module_;
+  std::unique_ptr<IRContext> context_;
+  std::unique_ptr<SsaNames> ssa_names_;
+  std::unique_ptr<Module> module_;
```
markww, line 68: Remove these fields, as mentioned above.

src/parser/ir/ir_parser.h
```diff
@@ -37,10 +38,22 @@ using ir_parser_generator::IrParser;
...
+    std::unique_ptr<Module> module;
+    std::unique_ptr<SsaNames> ssa_names;
+  };
+
+  static Result ParseProgram(absl::string_view prog_text) {
+    IrProgramParser parser(prog_text);
+    return {.context = std::move(parser.context_),
+            .module = std::move(parser.module_),
+            .ssa_names = std::move(parser.ssa_names_)};
+  }
```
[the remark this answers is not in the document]
bgogul: You will still need std::move even if you don't use unique_ptr?

In any case, we needed to return `unique_ptr` for pointer stability. Note that `ssa_names_` is keeping a map from `const Module*` to `std::string`. When we do `std::move(module)` and `std::move(ssa_names_)`, the entry for `module` in the `ssa_names_` becomes invalid. As a quick hack, I suggested using `std::unique_ptr`.

src/parser/ir/ir_parser_test.cc
```diff
@@ -25,8 +25,9 @@ class IrParserTest : public testing::TestWithParam<absl::string_view> {};
 
 TEST_P(IrParserTest, SimpleTestOperation) {
   auto input_program_text = GetParam();
-  IrProgramParser ir_parser;
-  EXPECT_EQ(IRPrinter::ToString(ir_parser.ParseProgram(input_program_text)),
+  auto [context, module, ssa_names] =
+      IrProgramParser::ParseProgram(input_program_text);
+  EXPECT_EQ(IRPrinter::ToString(*module, std::move(*ssa_names)),
```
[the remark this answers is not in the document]
bgogul: See other replies about why `unique_ptr` was  used. 

I agree `std::move` is probably not a clean solution here, but we are limited by the way `SsaNames` are used in the `IrPrinter`. I think we should really move away from treating `SsaNames` as free standing and  make it part of something like `IRContext` or a top-level module.  In the interim, we will need to have a way of passing `ssa_names` to IrPrinter. Perhaps, a _slightly_ cleaner solution is the following:

```
  EXPECT_EQ(IRPrinter::ToString(*module, std::move(*(ssa_names.release()))),
```

Another option is to make `IrPrinter::ToString` to take `std::unique_ptr<SsaNames>`. Even then you will still need to std::move.

I would suggest that we get this in as we need to unblock @asejfia to make progress on the DP project. After getting this in, let us take a step back and rethink where `SsaNames` should be kept.

src/parser/ir/ir_parser.h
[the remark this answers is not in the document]
markww: I meant that comment in the context of my suggestion to eliminate the mutable result fields. As you pointed out, that's less straightforward than I initially thought, so this comment makes less sense than it once did.

I still think it might make sense to have one `std::unique_ptr` to a `Result` rather than three `std::unique_ptr`s to these different objects though, depending on how these are used later.

src/parser/ir/ir_parser.h
markww: As Gogul pointed out, this suggestion may not make sense, but consider making one `std::unique_ptr` to a `Result`.

src/parser/ir/ir_parser_test.cc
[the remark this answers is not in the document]
markww: OK, that is fine for me for now. But rather than having a pattern that looks so dangerous, I think it might make sense to eliminate the `std::move` and just make this a copy. It's one copy in a test, so it's not like it will cause performance issues, but it will require less reasoning to understand that it's safe.

src/ir/dot/dot_generator.cc
harsha-mandadi: Done!

src/ir/dot/dot_generator_test.cc
harsha-mandadi: Done!

src/ir/dot/dot_generator.cc
harsha-mandadi: Done!

src/ir/value_test.cc
harsha-mandadi: The reason ssa names reply on the order in which names are assigned is because of the way we populate ssa names based on the size of the hash map. I think we do this to make sure that ids are unique. To avoid relying on order, specifically for Values, we can do some combination of hash map size and value name. (this also preserves uniqueness of ids)

src/parser/ir/ir_parser.cc
```diff
@@ -147,8 +150,7 @@ const Module& IrProgramParser::ParseProgram(absl::string_view prog_text) {
   // program_context points to the root of the parse tree
   IrParser::IrProgramContext& program_context =
       *CHECK_NOTNULL(parser.irProgram());
-  IrProgramParser::ConstructModule(*CHECK_NOTNULL(program_context.module()));
-  return module_;
+  ConstructModule(*CHECK_NOTNULL(program_context.module()));
```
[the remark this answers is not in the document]
harsha-mandadi: I took your suggestion and made following changes:
1. Embed Result with 3 objects as Values.
2. Remove unique_ptrs on three fields and have unique_ptr only on Result, keeping std::move(module) and std::move(ssa_names) as is.
This resulted in tests failing as it resulted in mismatched module names( m0 expected but m1 actual). My guess, not using unique_ptr on module and ssa_names, makes original module address invalid. 
I am keeping 3 unique ptr as of now. Please let me know if you have serious concerns with this change.

src/parser/ir/ir_parser.h
harsha-mandadi: See my comments above on why `std::unique_ptr` on result did not work

src/ir/value.cc
harsha-mandadi: I realized later that the change we discussed is not possible. I think what we agreed to do was something like this.   
`
std::string ToString(SsaNames& ssa_names) const {
    return absl::StrFormat("%s", ssa_names.GetOrCreateID(Value(*element_));
  }
`
In place of https://github.com/google-research/raksha/blob/c8c4593d6febcbea31ef864ca3e2267bac5d38e1/src/ir/value.h#L52

We would need to know the variant of value which is not something we can do in NameValue class. Let me know if you had other thoughts on this.

src/parser/ir/ir_parser.h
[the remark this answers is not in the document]
harsha-mandadi: Re:Take the work out of the constructor for IrProgramParser and put it here, making this method non-static.
@bgogul : We will be losing the static nature of the function. I remember you mentioning it would be good to have this static. Do we all agree on this?

src/parser/ir/ir_parser_test.cc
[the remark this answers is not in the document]
harsha-mandadi: Using this solution here `EXPECT_EQ(IRPrinter::ToString(*module, std::move(*(ssa_names.release()))),`. Cannot make a copy as SsaNames do not have a copy constructor

src/ir/value_test.cc
harsha-mandadi: Moving Value changes and discussions to another PR

src/ir/value.h
harsha-mandadi: Moving Value changes and discussions to another PR

src/ir/value.cc
harsha-mandadi: Moving Value changes and discussions to another PR.

src/ir/block_builder_test.cc
harsha-mandadi: Moving Value changes and discussions to another PR.

harsha-mandadi:
@bgogul As per the suggestion above, this PR handles SsaNames. Any further changes related to Value will be done in another PR. PTAL the changes made based on review comments.

src/ir/value.h
harsha-mandadi: Because I had time before the another round of review, added test for hash values.

src/ir/value_test.cc
harsha-mandadi: Because I had time before the another round of review, added changes to value test.

harsha-mandadi:
Note: No new PR is created for Values. Had some time before another round of review to fix value changes in this PR. Sorry for the confusion.

src/ir/ssa_names.h
```diff
@@ -17,51 +17,89 @@
...
+  // Disable copy semantics.
   SsaNames(const SsaNames &) = delete;
   SsaNames &operator=(const SsaNames &) = delete;
+  SsaNames(SsaNames &&) = default;
 
   ID GetOrCreateID(const Operation &operation) {
     return operation_ids_.GetOrCreateID(&operation);
   }
 
+  ID UpdateID(const Operation &operation, ID id) {
```
markww, line 47: As we talked about at lunch, let's make this `AddID` and have it `CHECK`-fail if there is already an entry for the item being added.

bgogul (approved):
LGTM modulo the following:
 - AddID
 - SsaNames ssa_names_ --> std::unique_ptr<SsaNames> ssa_names_ in IRPrinter.

src/parser/ir/ir_parser_test.cc
```diff
@@ -26,7 +26,8 @@ class IrParserTest : public testing::TestWithParam<absl::string_view> {};
 TEST_P(IrParserTest, SimpleTestOperation) {
   auto input_program_text = GetParam();
   IrProgramParser ir_parser;
-  EXPECT_EQ(IRPrinter::ToString(ir_parser.ParseProgram(input_program_text)),
+  auto result = ir_parser.ParseProgram(input_program_text);
+  EXPECT_EQ(IRPrinter::ToString(*result.module, std::move((*result.ssa_names.release()))),
```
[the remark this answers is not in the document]
harsha-mandadi: Done!. IRPrinter now holds on to a unique ptr of ssa names.!

src/parser/ir/ir_parser_test.cc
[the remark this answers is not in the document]
markww: OK, it's fine that we need to keep the move. But we should not be using `release`. Doing so will cause the memory managed by `std::unique_ptr` to not be freed and will cause a memory leak. This can be safely replaced with just:

```C++
EXPECT_EQ(IRPrinter::ToString(*module, std::move(*(ssa_names)))
```

That will move the value out of `*ssa_names`, but leave the pointer to `ssa_names` in `std::unique_ptr`, allowing the allocated memory to be cleaned up.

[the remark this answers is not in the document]
harsha-mandadi: Mark, I made changes to IRPrinter to hold on to ssa_names as a unique_ptr. Hence, we do not need to have `EXPECT_EQ(IRPrinter::ToString(*module, std::move(*(ssa_names)))` instead it become `EXPECT_EQ(IRPrinter::ToString(*module, std::move((ssa_names)))`

src/parser/ir/ir_parser_test.cc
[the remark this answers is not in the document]
markww: I'm still seeing the `release` here. If we `release` the pointer from `std::unique_ptr`, the `std::unique_ptr` will not `delete` it when it goes out of scope. I don't see how the allocated memory would be cleaned up; this looks like a leak to me.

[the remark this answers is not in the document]
harsha-mandadi: Mark, Can you refresh the PR. We no longer have `release`

src/parser/ir/ir_parser_test.cc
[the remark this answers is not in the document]
markww: ah, ok, sounds good. I was refreshing, but I still saw release. I guess we can chalk that up to a stale view or something. Sorry about that.

markww (approved):
LGTM

src/ir/ssa_names_test.cc
```diff
@@ -54,5 +54,15 @@ TYPED_TEST(SsaNamesTest, GetOrCreateIDReturnsUniqueIDs) {
   EXPECT_EQ(names.GetOrCreateID(this->second_entity_), second_id);
 }
 
+TYPED_TEST(SsaNamesTest, UpdateIDReturnsCorrectIDs) {
```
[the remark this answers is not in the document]
harsha-mandadi: Done!

PR #592, opened 2022-06-28
IR/Proto converters

Cypher1:
There's still a few things to finish wiring up, and the tests are stubs based on other tests but I hope this gives some indication of the direction I'm taking and the progress made so far.


src/ir/ir_traversing_visitor_with_result.h
```diff
@@ -0,0 +1,80 @@
...
+#include "src/ir/ir_visitor.h"
+#include "src/ir/module.h"
+
+namespace raksha::ir {
+
+// A visitor that also traverses the children of a node and allows performing
+// different actions before (PreVisit) and after (PostVisit) the children are
+// visited. Override any of the `PreVisit` and `PostVisit` methods as needed.
+template <typename Derived, typename Result=std::monostate>
+class IRTraversingVisitorWithResult : public IRVisitor<Derived, Result> {
```
bgogul, line 29: Why do we have a separate class? Can't we just add an additional `Result` parameter to `IrTraversingVisitor`?

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,92 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#include "src/ir/proto_to_ir.h"
+
+#include <memory>   // For std::pair.
+#include <variant>  // For std::visit.
```
bgogul, line 20: We don't need the comments at the end.

src/ir/ir_traversing_visitor_with_result.h
```diff
@@ -0,0 +1,80 @@
...
+  // Invoked before all the children of `block` is visited.
+  virtual Result PreVisit(const Block& block) { return Result(); }
+  // Invoked after all the children of `block` is visited.
+  virtual void PostVisit(const Block& block, Result& result) {}
+  virtual void FoldResult(const Block& block, Result child_result, Result& result) {}
+  // Invoked before all the children of `operation` is visited.
+  virtual Result PreVisit(const Operation& operation) { return Result(); }
+  // Invoked after all the children of `operation` is visited.
+  virtual void PostVisit(const Operation& operation, Result& result) {}
+  virtual void FoldResult(const Operation& operation, Result child_result, Result& result) {}
```
bgogul, line 47: The parent `Block` as an argument?

src/ir/ir_traversing_visitor_with_result.h
```diff
@@ -0,0 +1,80 @@
...
+  // Invoked before all the children of `module` is visited.
+  virtual Result PreVisit(const Module& module) { return Result(); }
+  // Invoked after all the children of `module` is visited.
+  virtual void PostVisit(const Module& module, Result& result) {}
+  virtual void FoldResult(const Module& module, Result child_result, Result& result) {}
+  // Invoked before all the children of `block` is visited.
+  virtual Result PreVisit(const Block& block) { return Result(); }
+  // Invoked after all the children of `block` is visited.
+  virtual void PostVisit(const Block& block, Result& result) {}
+  virtual void FoldResult(const Block& block, Result child_result, Result& result) {}
```
bgogul, line 42: parent `Module` as an argument?

src/ir/ir_traversing_visitor_with_result.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+    PostVisit(block, result);
+    return result;
+  }
+
+  Result Visit(const Operation& operation) final override {
+    Result result = PreVisit(operation);
+    const Module* module = operation.impl_module();
+    if (module != nullptr) {
+      FoldResult(operation, module->Accept<Derived, Result>(*this), result);
```
bgogul, line 71: Should this be `FoldResult(*module, module->Accept<Derived, Result>(*this), result)`?

src/ir/ir_traversing_visitor_with_result.h
```diff
@@ -0,0 +1,80 @@
...
+template <typename Derived, typename Result=std::monostate>
+class IRTraversingVisitorWithResult : public IRVisitor<Derived, Result> {
+ public:
+  virtual ~IRTraversingVisitorWithResult() {}
+
+  // Invoked before all the children of `module` is visited.
+  virtual Result PreVisit(const Module& module) { return Result(); }
+  // Invoked after all the children of `module` is visited.
+  virtual void PostVisit(const Module& module, Result& result) {}
+  virtual void FoldResult(const Module& module, Result child_result, Result& result) {}
```
bgogul, line 37: Don't we need the parent `Operation` as argument as well?

src/ir/ir_traversing_visitor_with_result.h
```diff
@@ -0,0 +1,80 @@
...
+      FoldResult(operation, module->Accept<Derived, Result>(*this), result);
+    }
+    PostVisit(operation, result);
+    return result;
+  }
+};
+
+}  // namespace raksha::ir
+
+#endif  // SRC_IR_IR_TRAVERSING_VISITOR_WITH_RESULT_H_
```
bgogul, line 80: add new line

src/ir/ir_traversing_visitor_with_result_test.cc
```diff
@@ -0,0 +1,169 @@
...
+  CollectingVisitor all_order_visitor(CollectingVisitor::TraversalType::BOTH);
+  ResultType nodes3 = block->Accept(all_order_visitor);
+  EXPECT_THAT(
+      nodes3,
+      testing::ElementsAre(block.get(), plus_op_instance, plus_op_instance,
+                           minus_op_instance, minus_op_instance, block.get()));
+}
+
+TEST(IRTraversingVisitorTest, TraversesOperationAsExpected) {
+  auto plus_op = std::make_unique<Operator>("core.plus");
```
bgogul, line 140: You can use the IR parser to create the input module: https://github.com/google-research/raksha/blob/main/src/parser/ir/ir_parser_test.cc

bgogul (commented):
I only looked at the IrTraversingVisitorWithResult implementation and had some suggestions. 

Let us iterate and converge on that first? (Also, once we converge on it, we should split it off to a separate PR.)

src/ir/ir_traversing_visitor_with_result.h
Cypher1: Aferr suggested that I implement beside the IRTraversingVisitor (at least until incompatibilities can be resolved).

src/ir/proto_to_ir.cc
Cypher1: Got it. Will remove

src/ir/ir_traversing_visitor_with_result.h
Cypher1: The argument of FoldResult *is* the parent in this case. The child_result holds the result from visiting the child.

src/ir/ir_traversing_visitor_with_result.h
Cypher1: Similarly to above, `operation` is the parent, passed to ensure that, if different FoldResult functions are needed, they can be specialized based on the parent type. Passing the module in once via the visit and once via the fold doesn't seem useful to me.

src/ir/ir_traversing_visitor_with_result.h
Cypher1: Similar to above.

src/ir/ir_traversing_visitor_with_result.h
Cypher1: Done

src/ir/ir_traversing_visitor_with_result_test.cc
Cypher1: That's a handy approach, thanks. Mark had previously suggested using the proto text format, but thanks for the options there.

Cypher1:
Thanks for the comments Gogul.

I included both Proto<->IR changes and IRTraversal changes in the same draft so that we could look at how compatible they are, should be very separable.
I'll look forward to your follow up comments on the traversal changes and will continue work on the Proto<->IR changes without trying to force them into the Visitor pattern (especially for the Proto->IR work it's poorly suited to a visitor over the IR).

src/ir/ir_traversing_visitor_with_result.h
bgogul: In that case, do you want to have the child, which is an Operation in this case?

bgogul: There is more detailed discussion in another thread. We can continue there. I will resolve this comment.

src/ir/ir_traversing_visitor_with_result.h
bgogul: > Passing the module in once via the visit and once via the fold doesn't seem useful to me. 

It could be useful if you want to combine the result from a children in specific way depending upon what the child is? 

In any case, it was hard to grasp what `Fold` is happening. A simple fix might be to add a `parent_` prefix to argument name as in `parent_operation`.

src/ir/ir_traversing_visitor_with_result.h
bgogul: Will continue discussion in the other thread.

src/ir/ir_traversing_visitor_with_result_test.cc
bgogul: > Mark had previously suggested using the proto text format, 

To use the proto text format, won't you need the proto deserializer, which is not yet available?  I was suggesting the IR parser for testing the IRTraversingVisitor in particular.

src/ir/ir_traversing_visitor_with_result.h
bgogul: After seeing this draft, I don't see any major incompatibilities. Let us go ahead and build it into IRTraversingVisitor itself?

src/ir/ir_traversing_visitor_with_result.h
bgogul: Will continue the discussion in the other thread.

src/ir/ir_traversing_visitor_with_result.h
bgogul: I also wonder if it makes sense to add a `Context` argument for the `Visit` methods.

src/ir/ir_traversing_visitor_with_result.h
Cypher1: Ah, I hadn't sent this out before our chat today.

For completeness:
Sounds good to me :) I'll upload the split off PR as you suggested and see how easy the change is to merge in.
The main difference is relying on the default construction of the `Return` type to have a decent default for the `PreVisit` functions and defaulting `Return` to `std::monostate`.

R.E. Our chat, I'll upload the updated form soon.

src/ir/block_builder.h
```diff
@@ -104,6 +109,22 @@ class BlockBuilder {
...
+    // The following check makes sure that we don't attempt to call
+    // `unsafe_get_block_ptr` on an empty `BlockBuilder` by checking that
+    // `block_` is not `nullptr`.
+    CHECK(block_) << "Attempt to get block pointer from a `BlockBuilder` that "
+                     "has already been"
+                     " built.";
+    return block_.get();
+  }
+
+  bool is_built() const {
```
bgogul, line 124: IsBuilt()

src/ir/block_builder.h
```diff
@@ -104,6 +109,22 @@ class BlockBuilder {
     return std::move(block_);
   }
 
+  // A way to get non-owning access to the block (for reference by operations)
+  // while the block is still under construction.
+  Block* get_block_ptr() {
```
bgogul, line 114: GetBlockPtr()

src/ir/ir_traversing_visitor_with_result.h
Cypher1: Is the context (in the class) not enough for this purpose already?

It would make it more functional and simplify testing (i.e. no need to inject values into the IRVisitor) but are there other benefits?

src/ir/block_builder.h
Cypher1: Done!
Though I find Java style method names less readable, thanks for the consistency notes.
Should we change other methods too or is this context dependent?

Cypher1:
@harsha-mandadi hey Harsha, I've started seeing my tests fail since your #588 PR merged.

Specifically
```
module m0 {
  block b0 {
    %0 = core.plus []()
  }  // block b0
  block b1 {
    %1 = core.plus [access: "private", transform: "no"](%0.out, <<ANY>>)
  }  // block b1
}  // module m0
",
```
Fails to parse with `value not found%0.out".

I'm wondering if you have some insight into that?

Cypher1:
@harsha-mandadi I discovered that the printer would incorrectly print lower ids for `OperationResults` if the SsaNames weren't told about Operation's during ProtoToIR conversion. This seems undesirable and I'm wondering if we can find a way to avoid this need.

Thanks

harsha-mandadi:
J, Can you provide me with expected result and actual result? That might help me in understanding issue better!

Cypher1:
Thanks Harsha, sorry I should have done that already.
Without my patch, the actual result is:
```
module m0 {
  block b0 {
    %0 = core.select []()
    %1 = core.merge [](<<ANY>>, <<ANY>>)
  }  // block b0
  block b1 {
    %2 = core.plus [access: "private", transform: "no"](%0, <<ANY>>)
    %3 = core.mult [access: "private", transform: "no"](%0, %1)
  }  // block b1
}  // module m0
```
Note the `%1` in the arguments of `%3`.

The input (and the expected result is however:
```
module m0 {
  block b0 {
    %0 = core.select []()
    %1 = core.merge [](<<ANY>>, <<ANY>>)
  }  // block b0
  block b1 {
    %2 = core.plus [access: "private", transform: "no"](%0, <<ANY>>)
    %3 = core.mult [access: "private", transform: "no"](%0, %2)
  }  // block b1
}  // module m0
```
Note the `%2` in the arguments of `%3`.

This is because (by default) the Printer looks up the operation result in `SsaNames` and doesn't find it, and so creates a new one, which happens to be `%1`.
Inserting the operation results in the `SsaNames` while converting from proto avoids this as the values are already in the maps.
I'm not sure what the best solution here is, perhaps keeping the IRContext, SsaNames and IR datastructures synchronized using builders would be better than this manual construction?

Keen to hear your thoughts.

harsha-mandadi:
J, Yes, your code would not behave as expected if SsaNames are not pre populated during Proto to IR conversion and handed it over to the IRPrinter. Even during IR parsing, we have to manually updated ssa_names for every new operation, block, module. We are planning on changing this kind of behavior in the near future. Can you please populate SsaNames during conversion as a temporary fix and we can clean it up later when we decide on the best way to synchronize SsaNames?

src/ir/proto_to_ir.cc
```diff
@@ -117,6 +120,8 @@ void ProtoToIR::PostVisit(BlockBuilder& block,
   for (const auto& result_it : block_proto.results().values()) {
     PostVisit(result_it.second);
   }
+  // Register the module with 'SsaNames'.
+  ssa_names_.GetOrCreateID(block.GetBlockPtr());
 }
 
 void ProtoToIR::PostVisit(Operation& op,
```
harsha-mandadi, line 127: We should populate ssa_names for Operation as well. Tests are not failing currently because we have written tests in sequential order for Operations.  So, IRPrinter looks at new operation and populate ssa_names based on the size of the ssa_names operation map so far. Suppose we have a test as mentioned below, it would fail. 
`
R"(module m0 {
  block b0 {
    %1 = core.select []()
    %0 = core.merge [](<<ANY>>, <<ANY>>)
  }  // block b0
  block b1 {
    %3 = core.plus [access: "private", transform: "no"](%0, <<ANY>>)
    %2 = core.mult [access: "private", transform: "no"](%0, %1)
  }  // block b1
}  // module m0
)"
`

src/ir/ir_to_proto_test.cc
```diff
@@ -0,0 +1,131 @@
...
+}  // module m0
+)",
+*/
+R"(module m0 {
+  block b0 {
+    %0 = core.select []()
+    %1 = core.merge [](<<ANY>>, <<ANY>>)
+  }  // block b0
+  block b1 {
+    %2 = core.plus [access: "private", transform: "no"](%0, <<ANY>>)
```
harsha-mandadi, line 115: Heads up. Moving forward we are planning on not supporting value access between blocks, instead have blocks return values to be used in other blocks (Exact details need to be figured out). In that case, this test would fail as `%0` in block b1 would be initialized in block b0. Please let us know your thoughts about this!

bgogul: Thanks, Harsha. Just to add a little more context. We should have the inputs and outputs of a block to be more constrained. Allowing operations in one block to directly access operations in another block will make it hard to reason about behavior of blocks. 

Harsha and I were chatting about this and we decided that the names should be unique within the block alone. If we want to support a block from accessing the result of another block, we might want to have a different syntax with a block prefix like `b0.%1`.

src/ir/ir_to_proto_test.cc
```diff
@@ -0,0 +1,131 @@
...
+  ASSERT_TRUE(differencer.Compare(orig_translation_unit_proto,
+                                  recovered_translation_unit_proto))
+      << "RECOVERED DATA DID NOT MATCH!\n"
+      << difference;
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    IrToProtoTest, IrToProtoTest,
+    ::testing::Values(
+        R"(module m0 {
```
bgogul, line 68: Since you are testing `proto -> IR -> proto` roundtrip, I wonder if you should just use the text representation of the proto itself, instead of the text representation of the IR. 

Won't that  should side-step the issues that you encounter with SsaNames? Specifically, you will not have to worry about populating `SsaNames` at all?

Cypher1: It'd be one way around it for this test, but it would produce IR that cannot be printed without misleading information.

I.e. in the example shown, the IR would be printed with the wrong arguments in some operations, making debugging very difficult.

I'd prefer to do this and protect against this case.

src/ir/proto_to_ir.cc
Cypher1: Operations are being registered in the PreVisit so that the ordering is consistent with their definitions rather than usages.

src/ir/ir_to_proto_test.cc
Cypher1: Gotcha. Thanks for the heads up. Can we land this and iterate?

src/ir/proto_to_ir.cc
harsha-mandadi: Ahh I see. Sounds good!!

bgogul:
@Cypher1, before you import it, can I take a quick look?

src/ir/ir_to_proto.h
```diff
@@ -0,0 +1,92 @@
...
+#include "src/ir/module.h"
+#include "src/ir/proto/raksha_ir.pb.h"
+#include "src/ir/value.h"
+
+namespace raksha::ir {
+
+class IRToProto {
+ public:
+  using ID = uint64_t;
+  IRToProto(ID& id)
```
bgogul, line 32: Why passing ID by reference? Just pass them by value?

I presume we will run  multiple `IRToProto` operations in sequence? If so, we can easily initialize the second one by exposing the `last_used_id` of the first one?

src/ir/ir_to_proto.h
```diff
@@ -0,0 +1,92 @@
...
+    if (used_ids_it != used_ids_.end()) {
+      return used_ids_it->second;
+    }
+    const ID id = ++last_used_id_;
+    used_ids_.insert({ptr, id});
+    return id;
+  }
+
+  absl::flat_hash_map<const void*, ID> used_ids_;
+  // This is a reference so that a two translations can more easily avoid having
```
bgogul, line 85: See comment above about passing ID by value.

src/ir/ir_to_proto.h
```diff
@@ -0,0 +1,92 @@
...
+  template <typename T>
+  const ID GetID(const T& thing_to_get_value_for) {
+    // Ensure that each ptr has a single unique ID.
+    const void* ptr = &thing_to_get_value_for;
+    const auto used_ids_it = used_ids_.find(ptr);
+    if (used_ids_it != used_ids_.end()) {
+      return used_ids_it->second;
+    }
+    const ID id = ++last_used_id_;
+    used_ids_.insert({ptr, id});
```
bgogul, line 80: You can use the result of an insert to avoid looking up twice: https://en.cppreference.com/w/cpp/container/unordered_map/insert

```suggestion
   auto insert_result = used_ids_.insert({ptr, last_used_id_ + 1});
   if (!insert_result.second) { return *insert_result.first; }
   return ++last_used_id;
```

src/ir/proto_to_ir.h
```diff
@@ -0,0 +1,122 @@
...
+  void PostVisit(const proto::Value& value);
+  void PostVisit(const proto::AnyValue& any_value);
+  void PostVisit(const proto::BlockArgumentValue& block_argument);
+  void PostVisit(const proto::OperationResultValue&);
+  void PostVisit(const proto::AttributePayload&);
+
+ private:
+  const IRContext& context_;
+  SsaNames& ssa_names_;
+  // TODO: Consider an 'OperationBuilder'
```
bgogul, line 115: File an issue?

src/ir/proto_to_ir.h
```diff
@@ -0,0 +1,122 @@
...
+    auto ssa_names = std::make_unique<SsaNames>();
+    ProtoToIR state(context, *ssa_names);
+    // Construct objects to point to during construction.
+    auto module = state.PreVisit(root.top_level_module());
+    // Perform construction.
+    state.ToIR(root.top_level_module());
+    // Move objects into their 'owning' objects.
+    state.PostVisit(*module, root.top_level_module());
+
+    // TODO: Check that the construction correctly handled ownership of all
```
bgogul, line 71: File an issue?

Cypher1:
> @Cypher1, before you import it, can I take a quick look?

Sure

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,260 @@
...
+}
+
+void ProtoToIR::PreVisit(const proto::Operation& operation_proto) {
+  for (const auto& input_proto : operation_proto.inputs()) {
+    PreVisit(input_proto);
+  }
+  for (const auto& attribute_it : operation_proto.attributes().attributes()) {
+    PreVisit(attribute_it.second);
+  }
+  // TODO: PreVisit(impl_module);
```
bgogul, line 64: File an issue?

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,260 @@
...
+  for (const auto& block_it : module_proto.blocks()) {
+    auto find_block_ptr = blocks_.find(block_it.id());
+    CHECK(find_block_ptr != blocks_.end())
+        << "This block builder should be pre-constructed.";
+    auto& builder = find_block_ptr->second;
+    PostVisit(builder, block_it.block());
+    module.AddBlock(builder.build());
+  }
+  // Register the module with 'SsaNames'.
+  ssa_names_.GetOrCreateID(module);
```
bgogul, line 107: Is there chance that this will be called multiple times for the same block? If not, you should be able to use `AddID`?

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,260 @@
...
+    ToIR(block_it.id(), block_it.block());
+  }
+}
+
+void ProtoToIR::ToIR(ProtoToIR::ID block_id, const proto::Block& block_proto) {
+  auto find_block_ptr = blocks_.find(block_id);
+  CHECK(find_block_ptr != blocks_.end())
+      << "This block should be pre-constructed.";
+  auto& block_builder = find_block_ptr->second;
+  // TODO: Consider preserving inputs/outputs.
```
bgogul, line 179: File an issue?

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,260 @@
...
+
+    Operation* operation = operation_ir_it->second.get();
+    CHECK(operation != nullptr) << "Operation ptr is empty!";
+    ToIR(*operation, operation_proto);
+  }
+}
+
+void ProtoToIR::ToIR(Operation& operation,
+                     const proto::Operation& operation_proto) {
+  // TODO: Consider preserving results and impl_module.
```
bgogul, line 200: File an issue?

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,260 @@
...
+
+void ProtoToIR::PostVisit(Operation& op,
+                          const proto::Operation& operation_proto) {
+  for (const auto& input_proto : operation_proto.inputs()) {
+    PostVisit(input_proto);
+  }
+  for (const auto& attribute_it : operation_proto.attributes().attributes()) {
+    PostVisit(attribute_it.second);
+  }
+  // TODO: impl_module
```
bgogul, line 135: file an issue?

src/ir/proto_to_ir.cc
```diff
@@ -0,0 +1,260 @@
...
+  auto block_it = blocks_.find(id);
+  CHECK(block_it != blocks_.end()) << "Could not find block with id " << id;
+  Block* block = block_it->second.GetBlockPtr();
+  return value::BlockArgument(*block,
+                              block_argument_proto.block_argument_name());
+}
+
+value::OperationResult ProtoToIR::ToIR(
+    const proto::OperationResultValue& operation_argument_proto) {
+  const auto id = operation_argument_proto.operation_id();
```
bgogul, line 243: why `const`?

src/ir/proto_to_ir.h
```diff
@@ -0,0 +1,122 @@
...
+#include "src/ir/block_builder.h"
+#include "src/ir/ir_context.h"
+#include "src/ir/module.h"
+#include "src/ir/proto/raksha_ir.pb.h"
+#include "src/ir/value.h"
+#include "src/ir/ssa_names.h"
+
+namespace raksha::ir {
+
+using ResultType =
```
bgogul, line 29: Too common a name to be in `raksha::ir` namespace. Move it inside `ProtoToIR` class?

src/ir/proto_to_ir.h
```diff
@@ -0,0 +1,122 @@
...
+  Attribute ToIR(const proto::AttributePayload&);
+
+  void PostVisit(Module& module, const proto::Module& module_proto);
+  void PostVisit(BlockBuilder& block, const proto::Block& block_proto);
+  void PostVisit(Operation& op, const proto::Operation& operation_proto);
+  void PostVisit(const proto::Value& value);
+  void PostVisit(const proto::AnyValue& any_value);
+  void PostVisit(const proto::BlockArgumentValue& block_argument);
+  void PostVisit(const proto::OperationResultValue&);
+  void PostVisit(const proto::AttributePayload&);
```
bgogul, line 110: Shoudln't all of these methods be private?

src/ir/module.h
```diff
@@ -62,7 +68,12 @@ class Operation {
     return visitor.Visit(*this);
   }
 
+  friend class ProtoToIR;
+
  private:
+  // Set the block to which this operation belongs. This is private so that only
+  // the Block can set it via its `AddOperation` function.
+  void set_parent_block(const Block& block) { parent_ = &block; }
```
bgogul, line 76: OK for now as this is private. However, @harsha-mandadi is adding a similar public method in one of her other PRs.

src/ir/module.h
```diff
@@ -85,6 +96,11 @@ class Block {
 
   Block() : parent_module_(nullptr) {}
 
+  Block(OperationList operations, NamedValueMap results)
+      : parent_module_(nullptr),
+        operations_{std::move(operations)},
+        results_{results} {}
```
bgogul, line 102: `results_{std::move(results)}`?

bgogul (approved):
Overall, looks good. Left a few nits and clean up suggestions.

src/ir/proto_to_ir.h
Cypher1: Ah, that's unused. Removed

src/ir/proto_to_ir.cc
Cypher1: There's no valid mutation to do to the id. The other ids are also const as they shouldn't be changed.

src/ir/proto_to_ir.h
Cypher1: Actually probably! Moved as much as I could to the `private:` section.

src/ir/module.h
Cypher1: I thought this was one of those cases where the name can't be used after this assignment so there's no need for a move, but I can put it in :)

src/ir/module.h
Cypher1: Okay, depending on the merge order I can remove it :)

src/ir/ir_to_proto.cc
```diff
@@ -0,0 +1,125 @@
...
+    proto::BlockWithId* block_proto = module_proto.mutable_blocks()->Add();
+    (*block_proto->mutable_block()) = BlockToProto(*block);
+    block_proto->set_id(GetID(*block));
+  }
+  return module_proto;
+}
+
+proto::Block IRToProto::BlockToProto(const Block& block) {
+  proto::Block block_proto;
+  // TODO: Consider preserving inputs/outputs.
```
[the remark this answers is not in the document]
Cypher1: Filed issues for the TODOs I have added.

https://github.com/google-research/raksha/issues/618
https://github.com/google-research/raksha/issues/619
https://github.com/google-research/raksha/issues/620

src/ir/ir_to_proto_test.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "src/parser/ir/ir_parser.h"
+
+namespace raksha::ir {
+
+class IrToProtoTest : public testing::TestWithParam<std::string_view> {};
+
+TEST_P(IrToProtoTest, RoundTripPreservesAllInformation) {
+  const auto input_program_text = GetParam();
+  IrProgramParser ir_parser;
+  // TODO: IRPrinter cannot take a const SsaNames.
```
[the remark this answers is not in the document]
Cypher1: Thx

src/ir/ir_to_proto.cc
```diff
@@ -0,0 +1,125 @@
...
+  const auto& name = op.name();
+  operation_proto.set_operator_name(std::string(name));
+  for (const auto& input : operation.inputs()) {
+    operation_proto.mutable_inputs()->Add(ValueToProto(input));
+  }
+  for (const auto& attribute : operation.attributes()) {
+    operation_proto.mutable_attributes()->mutable_attributes()->insert(
+        {attribute.first, AttributeToProto(attribute.second)});
+  }
+  // TODO: Consider preserving results and impl_module.
```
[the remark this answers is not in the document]
Cypher1: Filed an issue for all the lossy conversion issues #619

src/ir/ir_to_proto.cc
```diff
@@ -0,0 +1,125 @@
...
+  proto::Block block_proto;
+  // TODO: Consider preserving inputs/outputs.
+  for (const std::unique_ptr<Operation>& operation : block.operations()) {
+    proto::OperationWithId* operation_proto =
+        block_proto.mutable_operations()->Add();
+    (*operation_proto->mutable_operation()) = OperationToProto(*operation);
+    const uint64_t new_operation_id = GetID(*operation);
+    operation_proto->set_id(new_operation_id);
+  }
+  for (const std::pair<std::string, Value> result : block.results()) {
```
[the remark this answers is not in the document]
Cypher1: Thanks!

src/ir/ir_traversing_visitor_with_result.h
Cypher1: I think we can change this later if we find a need.

src/ir/ir_traversing_visitor_with_result_test.cc
Cypher1: I also have a proto deserializer in this change :)

Cypher1:
Hey Gogul and Harsha, I think I've resolved the conversations we had (sorry, I had to rebase my changes to follow main and this has made it harder to follow the conversations [such a recurring issue with Github]).
Thanks for all the feedback :)

src/ir/ir_to_proto.h
Cypher1: Thx. Added

src/ir/proto_to_ir.cc
Cypher1: Ah, good catch! Thanks

src/ir/proto_to_ir.h
Cypher1: Removed stale comment (following code does the check).

src/ir/ir_to_proto.h
Cypher1: I think this has been resolved elsewhere as I can't see a matching comment, resolving for now.

Cypher1:
R.E. Using `AddID` with `SsaNames`: I found that `AddID` takes a `Module` and an `ID`, so you have to call `GetOrCreateID` to get the ID to set. As such I didn't make the recommended change as it does unnecessary work.

PR #593, opened 2022-06-29
Support additional characters in constant names

aferr:
To support the transparent release process, we need to allow
constant names to include: -, +, /, =, spaces, and newlines. This is
needed to include the contents of public key files and URIs in
variables. The test included in `.../src/tests/test_hyphen_id.rs` and
`.../src/test_inputs/hyphenIDs` illustrates the use-case.

As a side-effect, the lex rules for variables (identifiers not in
quotes) and constants (identifiers in quotes) are more cleanly separate,
which should be a strict improvement in the grammar.


rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -139,7 +144,8 @@ ATTRIBUTE: 'attribute';
 
 // Identifiers wrapped in quotes are constants whereas
 // identifiers without quotes are variables.
-ID : ('"')? [_a-zA-Z][_a-zA-Z0-9/.#:]* ('"')?;
+VARIABLE : [_a-zA-Z][_a-zA-Z0-9/.#:]*; 
+CONSTANT : '"' [\n _a-zA-Z+-][\n _a-zA-Z0-9/.#:+=-]* '"';
```
aferr, line 148: @bgogul I want to specifically point out newlines are now allowed in constants since this could be controversial (but I think it's OK). See the hyphenIDs file for the motivation for this which is to allow the contents of public key files as constants

bgogul: Should we use a different delimiters for multiline strings? e.g., C++ raw literals are delimited by `R"(` and `)"` and in Kotlin, we use `"""` for multiline strings.

rust/tools/authorization-logic/src/test/test_hyphen_ids.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub fn test_hyphen_ids() {
+        utils::setup_directories_for_bazeltest(vec!["test_inputs"], vec!["test_outputs"]);
+        // The point of this test is to just see if it parses or if an
+        // error is thrown.
+        compile("test_inputs/hyphenIDs",
+                "test_outputs/hyphenIDs.dl",
+                &Vec::new());
+    }
+
+}
```
bgogul, line 33: add new line at the end of file

rust/tools/authorization-logic/test_inputs/hyphenIDs
```diff
@@ -0,0 +1,10 @@
+.decl ourFavoritePublicKeyIs(k : Key)
+.decl ourFavoriteTrustedBuilderIs(b : Builder)
+.decl variablesInsteadOfConstantsWithHyhensAreOKToo(v : TestVariable)
+"OakTeam" says {
+    ourFavoritePublicKeyIs("-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAErnyr7XY6gxXPnqRGYEpv5lDsbBeX
+izVFLckaF/cbgBuzx7kfmf1qi2j+3DDOQYgISQOj9/LYt4c1cxbt+XtG4g==
+-----END PUBLIC KEY-----").
+    ourFavoriteTrustedBuilderIs("https://github.com/project-oak/transparent-release").
+}
```
bgogul, line 10: add new line at the end of file.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -25,20 +25,25 @@ grammar AuthLogic;
 //-----------------------------------------------------------------------------
 // Parser
 //-----------------------------------------------------------------------------
+id
```
bgogul, line 28: BTW, I thought we just had one copy of `Authlogic.g4`? Are we starting to diverge between C++ and rust implementations?

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -25,20 +25,25 @@ grammar AuthLogic;
 //-----------------------------------------------------------------------------
 // Parser
 //-----------------------------------------------------------------------------
+id 
+    : CONSTANT
```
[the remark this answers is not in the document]
aferr: Yes, this is an argument to a Datalog predicate. This is meant to be an identifier surrounded in quotes, for example like `"example_arg"` in `some_relation("example_arg")` (by contrast to a variable like `other_arg` in `some_other_relation(other_arg)`. It looks like on the souffle-datalog webpage, [they also use the term constant to describe the same thing, TBDs aside](https://souffle-lang.github.io/arguments#intrinsic-functors-and-operations).

Is there some other name for this grammar rule you'd rather have me use?

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: Do we? I think I missed how exactly that was set up. Is there some script that copies from this file to some other location, or does it work some other way?

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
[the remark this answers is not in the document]
aferr: Is `STRING_LITERAL` what you're wanting to call this?

rust/tools/authorization-logic/src/test/test_hyphen_ids.rs
aferr: fixed

rust/tools/authorization-logic/test_inputs/hyphenIDs
aferr: fixed

aferr:
@bgogul PTAL again when you have a moment

rust/tools/authorization-logic/src/test/test_hyphen_ids.rs
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2021 Google LLC.
```
shaper, line 2: 2022 here and rest of PR

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -38,8 +38,22 @@ pub fn parse_program(prog_text: &str) -> AstProgram {
     construct_program(&parse_result)
 }
 
+// The authorizaiton logic frontend supports newlines
```
shaper, line 41: nit typo: authorizaiton

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
shaper: @harsha-mandadi can comment on above, but I am guessing it was manually copied. Can we shift to use a single central version as part of this PR (or a smaller separate single one)?

shaper (commented):
Drive-by review while reading, I defer to @bgogul and @harsha-mandadi

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
harsha-mandadi: `AuthLogic.g4` was copied manually. I made changes to rust build to point to a C++ version of g4 file PR #469

shaper: It looks like the c3po Copybara commit in GH didn't include the rust-side file edit for some reason, from briefly looking at: https://github.com/google-research/raksha/commit/b2729ec8806d16c222e402eef34bb33981a5d726

The PR description is potentially ambiguous/confusing in that it states "move", but if I am reading it correctly, it did not actually delete the original `AuthLogic.g4` file which would have been preferable unless there was a known reason to leave it?

harsha-mandadi: AuthLogic.g4 rust version should have been deleted.

shaper: Just a guess, the Raksha `copy.bara.sky` file looks like it's skipping auth logic via this part of config (`exclude` in the github files section, line 33):

```
        # Don't need auth logic yet.
        "rust/**",
```

Maybe this led to stripping the changes to auth logic as we used `ready-to-pull` label, so it imported, then exported.

Do we merge all changes through that flow, or can/do we also land them in GH directly?

markww:
I think one issue with this change is that it targets the Rust version of the authorization logic, which for Google internal reasons we have been in the process of abandoning. I think @harsha-mandadi is fairly far along in the process of getting rid of the Rust code and this is a change to almost dead code.

I know that we were moving forward the Rust version for some authorization logic purposes to produce a demo on a deadline, but if I understand correctly, that demo has been done and we should refocus the authorization logic effort on the canonical C++ version.

aferr:
@markww There is still some ongoing work for transparent release. We're also considering submitting a workshop paper about transparent release + auth logic this month and if we block on moving the transparent release prototype to C++ now, it would make hitting the workshop deadline nearly impossible. Can we at least put off moving transparent release to the C++ version of authorization logic until we are done with the workshop submission (so in August)?

I will continue to need to do some work on this Rust prototype until then.

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
aferr: fixed

aferr: fixed

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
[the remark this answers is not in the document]
aferr: I think maybe I understand what your point here was now. Is your point really that "variables" and "id"s are usually synonyms and so it doesn't make sense for a "constant" to be a type of "id" (or a type of variable)?

This kind of makes sense to me, but then I don't know what to call the grammar rule that's now called "ID". The grammar rule for "ID" is not just used for predicate arguments, it's used for principals as well. Predicate arguments also include `NUMLITERAL` in addition to the term currently called `ID`. Any suggestions for what to call this grammar rule instead?

Also note that prior to this PR, the lex rule for ID used to effectively include both of `VARIABLE` and `CONSTANT`. This PR splits these lex rules in two. I looked through the uses of "id" in the grammar, and one clear case where we can restrict from "id" to just `VARIABLE` was in `relationDeclaration`, so I changed that.

I suspect these low-level details of the grammar can still be iterated on, but this change is mainly focused on making this useful for a transparent release use-case...

aferr:
> Hi @aferr I just wanted to check in and see how we'd like to proceed here. I think this was discussed in chat but I'm not sure of the latest. @markww is back tomorrow. I'm sensitive to your being blocked/slowed down. Let us know thoughts.

Thanks for checking in! Actually, I remained silent here because it turns out I am not blocked by this since we are planning to submit our paper with a case study rather than experiments, so we are not depending on this for the workshop deadline after all. So we can discuss how we want to proceed long-term soon.

I am imagining the long-term solution will entail migrating the transparency work to use our C++ implementation and deprecating this rust version (as long as the C++ implementation is feature-ready enough for this).

aferr:
OK so at the moment the transparent release code is essentially treating this implementation of authorization logic as canonical (it is using a binary generated as of this in-flight PR). So, in order to do the migration to the C++ implementation, the plan is:
 - we finalize this PR with the understanding that we'll need to make similar changes to the C++ side (so if we think this language change should work differently, please make those comments)
 -  we then implement similar changes to the C++ side along with others to allow the C++ implementation of auth logic work as a standalone binary
 - we migrate transparent release to use the C++ implementation of auth logic.

rust/tools/authorization-logic/test_inputs/hyphenIDs
```diff
@@ -0,0 +1,27 @@
...
+.decl ourFavoriteTrustedBuilderIs(b : Builder)
+.decl variablesInsteadOfConstantsWithHyhensAreOKToo(v : TestVariable)
+"OakTeam" says {
+    ourFavoritePublicKeyIs(
+"""-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAErnyr7XY6gxXPnqRGYEpv5lDsbBeX
+izVFLckaF/cbgBuzx7kfmf1qi2j+3DDOQYgISQOj9/LYt4c1cxbt+XtG4g==
+-----END PUBLIC KEY-----""").
+    ourFavoriteTrustedBuilderIs("https://github.com/project-oak/transparent-release").
+    ourFavoriteTrustedBuilderIs("https://github.com/Attestations/GitHubHostedActions@v1").
```
markww, line 26: Is sanitizing `"` from filepaths tested here? I'm not seeing any `"` in the middle of these files. Maybe this is supposed to address the `"`s at the beginning and end?

markww (commented):
I'm wondering if things are not factored correctly between the parser/lexer and later stages. In particular, rather than `"` and `"""` being part of the contents of a `CONSTANT` or `MULTILINE_CONSTANT`, respectively, shouldn't they be syntactic markers in the grammar that indicate that a `CONSTANT` or `MULTILINE_CONSTANT` is to be read, with those characters not included in the constant token(s)? That would potentially prevent from having to worry about the sanitization of these constants later in the program.

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -38,8 +38,22 @@ pub fn parse_program(prog_text: &str) -> AstProgram {
...
 
+// The authorization logic frontend supports newlines
+// in the names of constants, but souffle does not. As
+// an easy way to support this, we delete newlines from
+// constant names. Since multiline literals/constants
+// are represented with triple quotes, we replace
+// triple quotes with single here.
+fn sanitize_id(name: String) -> String {
+    name.replace("\n", "").replace("\"\"\"", "\"")
+}
```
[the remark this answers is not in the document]
aferr: > After reading more closely, I suspect the answer is "this will not cause issues, we want a multi-line constant with equivalent contents to a single-line constant to be treated identically."

Yes, I think the same way as you do.

I am not sure I follow your second comment. By the time the code in this file is running, antlr has already done all the parsing/lexing, and this piece of code is essentially the "earliest" piece that is not generated by antlr and is directly rust code that we are writing.

This sanitize_id function is essentially doing a very tiny and obvious part of the translation from auth logic into souffle early (just as we are building the AST objects rather than later when we do the rest of the translation). Perhaps your point was really not about doing this early, but about doing this later in case we want to handle the newlines differently later in the compilation pipeline... I don't see a use-case for this yet though, so I think doing this here is OK.

Let me know if I missed your point.

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
```diff
@@ -38,8 +38,22 @@ pub fn parse_program(prog_text: &str) -> AstProgram {
...
+
+fn sanitize_filepath(name: String) -> String {
+    name.replace("\"", "")
+}
```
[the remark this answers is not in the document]
aferr: So in the syntax for auth logic prototyped here file paths are used: 1) as paths for claims being imported/exported, 2) to give paths for public/private cryptographic keys. For either case, the file path as written in the auth logic frontend can either be wrapped in quotes or not. I think the tests write file paths both with and without quotes. The collisions are fine since IIUC file paths usually do not allow quotes in them, so we have no need to be able to refer to file paths that actually have quotes.

I think we can/should iterate on the actual syntax we want to keep, but since we are deprecating this version of auth logic, it is probably best if we turn that nob after we have migrated the transparent release work to use the C++ implementation rather than this one.

rust/tools/authorization-logic/test_inputs/hyphenIDs
aferr: Sanitizing `"` out of filepaths was actually something this code was doing before this PR. I just refactored it here (for example, see the old line 230 where the `replace` was being done where the filename was parsed. It was also doing this `replace` in all the the other spots where a filename was parsed. ) 

So this new test does not test sanitization of filepath names, but other pre-existing tests already did like [this one](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/test_inputs/quotesInExports), so this change must not have caused a regression.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -139,7 +145,9 @@ ATTRIBUTE: 'attribute';
 
 // Identifiers wrapped in quotes are constants whereas
 // identifiers without quotes are variables.
-ID : ('"')? [_a-zA-Z][_a-zA-Z0-9/.#:]* ('"')?;
+VARIABLE : [_a-zA-Z][_a-zA-Z0-9/.#:]*; 
+CONSTANT : '"' [_a-zA-Z+-][@_a-zA-Z0-9/.#:+=-]* '"';
+MULTILINE_CONSTANT : '"""' [\n _a-zA-Z+-][\n _a-zA-Z0-9/.#:+=-]* '"""';
```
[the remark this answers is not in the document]
aferr: OK I tried refactoring it the way you were suggesting [here](https://github.com/google-research/raksha/pull/627/files#diff-f485debf22070ea87700d2b589051e492d0a3a0f9e9084f854cd242dbaad38dd) -- I do think your suggestion is better in principle.

With the way I was trying to implement your suggestion, antlr seems to handle newlines differently in some way that breaks things. When trying to parse a relation declaration, it fails to parse the relation name into the right smaller piece of syntax (here a VARIABLE, but I also tried leaving it as an ID), and this breaks all the tests. Weirdly, if I start some of the tests (like `numStringsInNames` // cargo test test_num_string_names) with a multiline comment, it suddenly parses correctly.

I spent some time trying to get it to work this way, and I could keep trying to pull on this thread if you think it is much better, but otherwise perhaps we can leave this as-is for now since we are likely to revisit the frontend of this language anyway. Let me know what you think.

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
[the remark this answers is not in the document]
aferr: I think your other comment in this thread about separating the quotes out in the antlr grammar is similar to the point you were making here. If so, perhaps this thread "is resolved" and the discussion can continue in the other related thread in `AuthLogic.g4`

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -139,7 +145,9 @@ ATTRIBUTE: 'attribute';
 
 // Identifiers wrapped in quotes are constants whereas
 // identifiers without quotes are variables.
-ID : ('"')? [_a-zA-Z][_a-zA-Z0-9/.#:]* ('"')?;
+VARIABLE : [_a-zA-Z][_a-zA-Z0-9/.#:]*; 
+CONSTANT : '"' [_a-zA-Z+-][@_a-zA-Z0-9/.#:+=-]* '"';
```
markww, line 149: I just noticed this - you're adding `+` and `-` as potential first characters in the pattern for `CONSTANT`. I think what you're trying to do is to say that a constant can start with a `+` or a `-`, indicating a unary plus or minus. But you're entangling that with the concern of "allowing a smaller set of characters to be the first character of an identifier" - which means that while `"0"` is not a legal constant, `"-0"` is. Of course, I'm not entirely clear, mentally, what a constant is in this context, so maybe this is ok.

I do wonder about that first-character restriction for constants at all though. I get it for identifiers, it's a common way to keep from confusing humans reading code, but I'm not sure what function it serves for constants which will be wrapped in `"` or `"""` anyway.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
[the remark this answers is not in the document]
markww: I wonder whether your specification of `CONSTANT` versus `MULTILINE_CONSTANT` is running afoul of this trap in Antlr due to them both ending with a `"`: https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#redundant-string-literals

In that case, the `\n` could actually be disambiguating between the two because `\n` is not allowed in a regular constant.

One option could be to change a multiline constant to use three single quotes instead of double quotes. Or, perhaps, have a syntax closer to C++'s [raw string literal](https://en.cppreference.com/w/cpp/language/string_literal).

markww (approved):
I don't want to hold you up on parser issues you didn't introduce. I'd say just file an issue about it.

Do, however, look at my question about the opening character of a constant.

rust/tools/authorization-logic/src/parsing/astconstructionvisitor.rs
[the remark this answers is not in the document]
markww: yes, that's correct.

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: >  I think what you're trying to do is to say that a constant can start with a + or a -, indicating a unary plus or minus.

Good guess, but this was actually not meant to be a proxy for unary operators. The only unary operator in this language is negation of predicates and that is handled "the right way" by making this an optional char in the parse rule for predicate and storing negation as part of the AST node for predicates. The actual motivation for including `+-=` in CONSTANT was to support principal names that are public keys like [this one](https://github.com/project-oak/transparent-release/blob/2f25e7a46ea571f7ff33d1e0c35dc279247daadd/experimental/auth-logic/endorsement-release/input_policy_examples/oak_endorsement_policy.auth_logic#L27). But actually, cryptographic keys are going to be MULTILINE_CONSTANTs in all cases, so these characters don't need to be in the rule for CONSTANTs which cannot take multiple lines. I left `-` in the middle of CONSTANT because hyphens do show up in desired principal names for transparent release and this seems OK to support in general IMO.

> I do wonder about that first-character restriction for constants at all though.

The difference in first characters vs other characters in these lex rules exists because I started these from standard lex rules for identifiers and then expanded them as needed from the use-cases I was looking at. I think it's fine to re-evaluate if this is how it should work for our policy language going forward. Since this kind of change is really "shallow" to make (it only requires changing the grammar rules and will have no affect on the data structures implied by the grammar so it will not require any changes to the AST nodes or compiler passes internally). I think it may be best to do this kind of language re-evaluation post migration and as a part of our re-evaulation of our policy language design.

PR #598, opened 2022-07-01
Adding return values to our Visitor pattern

Cypher1:
Looking forward to feedback, of course open to changes but won't see them till US Tuesday

### DIFF
--- a/rust/tools/authorization-logic/BUILD
+++ b/rust/tools/authorization-logic/BUILD
@@ -16,7 +16,7 @@
 #-------------------------------------------------------------------------------
 load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//:__subpackages__"])
 
 licenses(["notice"])
 
--- a/rust/tools/authorization-logic/BUILD
+++ b/rust/tools/authorization-logic/BUILD
@@ -14,9 +14,9 @@
 # limitations under the License.
 #
 #-------------------------------------------------------------------------------
-package(default_visibility = ["//visibility:public"])
-
 load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
+
+package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
 
--- a/src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
+++ b/src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
@@ -51,7 +51,7 @@ static DatalogAttributePayload GetPayloadForAttribute(ir::Attribute attr,
   return DatalogAttribute::String("");
 }
 
-void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
+Unit DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
   const ir::Operator &op = operation.op();
   absl::string_view op_name = op.name();
 
@@ -89,6 +89,7 @@ void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
       operation, frontends::sql::DecoderContext::kDefaultOutputName)));
   datalog_facts_.AddIsOperationFact(
       DatalogIsOperationFact(std::move(datalog_operation)));
+  return Unit();
 }
 
 }  // namespace raksha::backends::policy_engine::souffle
--- a/src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
+++ b/src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
@@ -51,7 +51,7 @@ static DatalogAttributePayload GetPayloadForAttribute(ir::Attribute attr,
   return DatalogAttribute::String("");
 }
 
-void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
+std::monostate DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
   const ir::Operator &op = operation.op();
   absl::string_view op_name = op.name();
 
@@ -89,6 +89,7 @@ void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
       operation, frontends::sql::DecoderContext::kDefaultOutputName)));
   datalog_facts_.AddIsOperationFact(
       DatalogIsOperationFact(std::move(datalog_operation)));
+  return {};
 }
 
 }  // namespace raksha::backends::policy_engine::souffle
--- a/src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
+++ b/src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
@@ -51,7 +51,7 @@ static DatalogAttributePayload GetPayloadForAttribute(ir::Attribute attr,
   return DatalogAttribute::String("");
 }
 
-std::monostate DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
+void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
   const ir::Operator &op = operation.op();
   absl::string_view op_name = op.name();
 
@@ -89,7 +89,6 @@ std::monostate DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) 
       operation, frontends::sql::DecoderContext::kDefaultOutputName)));
   datalog_facts_.AddIsOperationFact(
       DatalogIsOperationFact(std::move(datalog_operation)));
-  return {};
 }
 
 }  // namespace raksha::backends::policy_engine::souffle
--- a/src/backends/policy_engine/souffle/datalog_lowering_visitor.h
+++ b/src/backends/policy_engine/souffle/datalog_lowering_visitor.h
@@ -17,6 +17,7 @@
 #ifndef SRC_BACKENDS_POLICY_ENGINE_SOUFFLE_DATALOG_LOWERING_VISITOR_H_
 #define SRC_BACKENDS_POLICY_ENGINE_SOUFFLE_DATALOG_LOWERING_VISITOR_H_
 
+#include <variant>
 #include "src/backends/policy_engine/souffle/raksha_datalog_facts.h"
 #include "src/common/logging/logging.h"
 #include "src/ir/attributes/attribute.h"
@@ -32,13 +33,13 @@
 namespace raksha::backends::policy_engine::souffle {
 
 class DatalogLoweringVisitor
-    : public ir::IRTraversingVisitor<DatalogLoweringVisitor> {
+    : public ir::IRTraversingVisitor<DatalogLoweringVisitor, std::monostate> {
  public:
   // We currently don't have any owner information when outputting IR. We don't
   // need it yet, really, but we do need to output something.
   static constexpr absl::string_view kDefaultPrincipal = "sql";
 
-  void PreVisit(const ir::Operation &operation) override;
+  std::monostate PreVisit(const ir::Operation &operation) override;
 
   const RakshaDatalogFacts &datalog_facts() { return datalog_facts_; }
 
--- a/src/backends/policy_engine/souffle/datalog_lowering_visitor.h
+++ b/src/backends/policy_engine/souffle/datalog_lowering_visitor.h
@@ -18,6 +18,7 @@
 #define SRC_BACKENDS_POLICY_ENGINE_SOUFFLE_DATALOG_LOWERING_VISITOR_H_
 
 #include "src/backends/policy_engine/souffle/raksha_datalog_facts.h"
+#include "src/common/utils/types.h"
 #include "src/common/logging/logging.h"
 #include "src/ir/attributes/attribute.h"
 #include "src/ir/attributes/int_attribute.h"
@@ -38,7 +39,7 @@ class DatalogLoweringVisitor
   // need it yet, really, but we do need to output something.
   static constexpr absl::string_view kDefaultPrincipal = "sql";
 
-  void PreVisit(const ir::Operation &operation) override;
+  Unit PreVisit(const ir::Operation &operation) override;
 
   const RakshaDatalogFacts &datalog_facts() { return datalog_facts_; }
 
--- a/src/backends/policy_engine/souffle/datalog_lowering_visitor.h
+++ b/src/backends/policy_engine/souffle/datalog_lowering_visitor.h
@@ -17,7 +17,6 @@
 #ifndef SRC_BACKENDS_POLICY_ENGINE_SOUFFLE_DATALOG_LOWERING_VISITOR_H_
 #define SRC_BACKENDS_POLICY_ENGINE_SOUFFLE_DATALOG_LOWERING_VISITOR_H_
 
-#include <variant>
 #include "src/backends/policy_engine/souffle/raksha_datalog_facts.h"
 #include "src/common/logging/logging.h"
 #include "src/ir/attributes/attribute.h"
@@ -33,13 +32,13 @@
 namespace raksha::backends::policy_engine::souffle {
 
 class DatalogLoweringVisitor
-    : public ir::IRTraversingVisitor<DatalogLoweringVisitor, std::monostate> {
+    : public ir::IRTraversingVisitor<DatalogLoweringVisitor> {
  public:
   // We currently don't have any owner information when outputting IR. We don't
   // need it yet, really, but we do need to output something.
   static constexpr absl::string_view kDefaultPrincipal = "sql";
 
-  std::monostate PreVisit(const ir::Operation &operation) override;
+  void PreVisit(const ir::Operation &operation) override;
 
   const RakshaDatalogFacts &datalog_facts() { return datalog_facts_; }
 
--- a/src/ir/dot/dot_generator.cc
+++ b/src/ir/dot/dot_generator.cc
@@ -41,8 +41,8 @@ class DotGeneratorHelper : public IRTraversingVisitor<DotGeneratorHelper> {
     return visitor.GetDotGraph();
   }
 
-  void PreVisit(const Block& block) override;
-  void PreVisit(const Operation& operation) override;
+  Unit PreVisit(const Block& block) override;
+  Unit PreVisit(const Operation& operation) override;
 
  private:
   DotGeneratorHelper(DotGeneratorConfig config) : config_(std::move(config)) {}
@@ -244,7 +244,7 @@ std::string DotGeneratorHelper::GetDotGraph() {
   return absl::StrFormat(kDigraphFormat, blocks, operations, edges);
 }
 
-void DotGeneratorHelper::PreVisit(const Operation& operation) {
+Unit DotGeneratorHelper::PreVisit(const Operation& operation) {
   std::string node_name = GetNodeName(operation);
   AddOperationNode(node_name, operation);
   unsigned input_index = 0;
@@ -269,11 +269,13 @@ void DotGeneratorHelper::PreVisit(const Operation& operation) {
       AddEdge(GetNodeName(stored_value->storage()), input_name);
     }
   }
+  return Unit();
 }
 
-void DotGeneratorHelper::PreVisit(const Block& block) {
+Unit DotGeneratorHelper::PreVisit(const Block& block) {
   std::string block_name = GetNodeName(block);
   AddBlockNode(block_name, block);
+  return Unit();
 }
 
 std::string DotGenerator::ModuleAsDot(const Module& module,
--- a/src/ir/dot/dot_generator.cc
+++ b/src/ir/dot/dot_generator.cc
@@ -41,8 +41,8 @@ class DotGeneratorHelper : public IRTraversingVisitor<DotGeneratorHelper> {
     return visitor.GetDotGraph();
   }
 
-  void PreVisit(const Block& block) override;
-  void PreVisit(const Operation& operation) override;
+  std::monostate PreVisit(const Block& block) override;
+  std::monostate PreVisit(const Operation& operation) override;
 
  private:
   DotGeneratorHelper(DotGeneratorConfig config) : config_(std::move(config)) {}
@@ -244,7 +244,7 @@ std::string DotGeneratorHelper::GetDotGraph() {
   return absl::StrFormat(kDigraphFormat, blocks, operations, edges);
 }
 
-void DotGeneratorHelper::PreVisit(const Operation& operation) {
+std::monostate DotGeneratorHelper::PreVisit(const Operation& operation) {
   std::string node_name = GetNodeName(operation);
   AddOperationNode(node_name, operation);
   unsigned input_index = 0;
@@ -269,11 +269,13 @@ void DotGeneratorHelper::PreVisit(const Operation& operation) {
       AddEdge(GetNodeName(stored_value->storage()), input_name);
     }
   }
+  return {};
 }
 
-void DotGeneratorHelper::PreVisit(const Block& block) {
+std::monostate DotGeneratorHelper::PreVisit(const Block& block) {
   std::string block_name = GetNodeName(block);
   AddBlockNode(block_name, block);
+  return {};
 }
 
 std::string DotGenerator::ModuleAsDot(const Module& module,
--- a/src/ir/dot/dot_generator.cc
+++ b/src/ir/dot/dot_generator.cc
@@ -41,8 +41,8 @@ class DotGeneratorHelper : public IRTraversingVisitor<DotGeneratorHelper> {
     return visitor.GetDotGraph();
   }
 
-  std::monostate PreVisit(const Block& block) override;
-  std::monostate PreVisit(const Operation& operation) override;
+  void PreVisit(const Block& block) override;
+  void PreVisit(const Operation& operation) override;
 
  private:
   DotGeneratorHelper(DotGeneratorConfig config) : config_(std::move(config)) {}
@@ -244,7 +244,7 @@ std::string DotGeneratorHelper::GetDotGraph() {
   return absl::StrFormat(kDigraphFormat, blocks, operations, edges);
 }
 
-std::monostate DotGeneratorHelper::PreVisit(const Operation& operation) {
+void DotGeneratorHelper::PreVisit(const Operation& operation) {
   std::string node_name = GetNodeName(operation);
   AddOperationNode(node_name, operation);
   unsigned input_index = 0;
@@ -269,13 +269,11 @@ std::monostate DotGeneratorHelper::PreVisit(const Operation& operation) {
       AddEdge(GetNodeName(stored_value->storage()), input_name);
     }
   }
-  return {};
 }
 
-std::monostate DotGeneratorHelper::PreVisit(const Block& block) {
+void DotGeneratorHelper::PreVisit(const Block& block) {
   std::string block_name = GetNodeName(block);
   AddBlockNode(block_name, block);
-  return {};
 }
 
 std::string DotGenerator::ModuleAsDot(const Module& module,
--- a/src/ir/ir_traversing_visitor.h
+++ b/src/ir/ir_traversing_visitor.h
@@ -37,7 +37,7 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
 
   // Invoked before all the children of `module` is visited.
   virtual Result PreVisit(const Module& module) {
-    return std::move(GetDefaultValue());
+    return GetDefaultValue();
   }
   // Invoked after all the children of `module` is visited.
   virtual Result PostVisit(const Module& module, Result in_order_result) {
@@ -51,7 +51,7 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
   }
   // Invoked before all the children of `block` is visited.
   virtual Result PreVisit(const Block& block) {
-    return std::move(GetDefaultValue());
+    return GetDefaultValue();
   }
   // Invoked after all the children of `block` is visited.
   virtual Result PostVisit(const Block& block, Result in_order_result) {
@@ -65,7 +65,7 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
   }
   // Invoked before all the children of `operation` is visited.
   virtual Result PreVisit(const Operation& operation) {
-    return std::move(GetDefaultValue());
+    return GetDefaultValue();
   }
   // Invoked after all the children of `operation` is visited.
   virtual Result PostVisit(const Operation& operation, Result in_order_result) {
@@ -83,8 +83,8 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
     Result fold_result = common::utils::fold(
         module.blocks(), std::move(pre_visit_result),
         [this, &module](Result acc, const std::unique_ptr<Block>& block) {
-          return std::move(FoldResult(std::move(acc), module,
-                                      block->Accept(*this)));
+          return FoldResult(std::move(acc), module,
+                                      block->Accept(*this));
         });
     return PostVisit(module, std::move(fold_result));
   }
@@ -94,9 +94,8 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
     Result fold_result = common::utils::fold(
         block.operations(), std::move(pre_visit_result),
         [this, &block](Result acc, const std::unique_ptr<Operation>& operation) {
-          return std::move(
-              FoldResult(std::move(acc), block,
-                         operation->Accept(*this)));
+          return FoldResult(std::move(acc), block,
+                         operation->Accept(*this));
         });
     return PostVisit(block, std::move(fold_result));
   }
@@ -105,8 +104,7 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
     Result result = PreVisit(operation);
     const Module* module = operation.impl_module();
     if (module != nullptr) {
-      result = std::move(FoldResult(std::move(result), operation,
-                                    module->Accept<Derived, Result>(*this)));
+      result = FoldResult(std::move(result), operation, module->Accept<Derived, Result>(*this));
     }
     return PostVisit(operation, std::move(result));
   }
--- a/src/ir/ir_traversing_visitor.h
+++ b/src/ir/ir_traversing_visitor.h
@@ -42,36 +42,36 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
     return std::move(GetDefaultValue());
   }
   // Invoked after all the children of `module` is visited.
-  virtual Result PostVisit(const Module& module, Result result) {
-    return std::move(result);
+  virtual Result PostVisit(const Module& module, Result in_order_result) {
+    return std::move(in_order_result);
   }
   virtual Result FoldResult(const Module& module, Result child_result,
-                            Result result) {
-    return std::move(result);
+                            Result in_order_result) {
+    return std::move(in_order_result);
   }
   // Invoked before all the children of `block` is visited.
   virtual Result PreVisit(const Block& block) {
     return std::move(GetDefaultValue());
   }
   // Invoked after all the children of `block` is visited.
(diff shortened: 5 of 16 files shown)


rust/tools/authorization-logic/BUILD
```diff
@@ -14,10 +14,10 @@
 # limitations under the License.
 #
 #-------------------------------------------------------------------------------
-package(default_visibility = ["//visibility:public"])
-
 load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
 
+package(default_visibility = ["//visibility:public"])
```
Cypher1, line 19: This is a change based on linter feedback

markww: I see. I'm not sure that this is a good idea: when we import these things into the Google internal repo, the checks will get mad at us for having `public` visibility. If you want something visible throughout all of Raksha, try doing `default_visibility = ["//:__subpackages__"]`, which is basically the same out here but won't make the internal checks upset.

src/ir/ir_traversing_visitor_test.cc
```diff
@@ -151,6 +154,151 @@ TEST(IRTraversingVisitorTest, TraversesOperationAsExpected) {
...
+
+namespace {
+using ResultType=std::vector<void*>;
+
+// Another small test visitor that collects the nodes as they are visited to make
+// sure that  Pre and Post visits happened correctly.
+// Uses 'returns' rather than side effects.
+class ReturningVisitor : public IRTraversingVisitor<ReturningVisitor, ResultType> {
+ public:
+  enum class TraversalType { PRE = 0x1, POST = 0x2, BOTH = 0x3 };
```
markww, line 167: According to the C++ style guide, enumerator names should be of the form `kEnumeratorName`: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,53 @@ namespace raksha::ir {
...
-    PreVisit(module);
+  Result Visit(const Module& module) final override {
+    Result result = PreVisit(module);
     for (const std::unique_ptr<Block>& block : module.blocks()) {
-      block->Accept(*this);
+      FoldResult(*block, block->Accept<Derived, Result>(*this), result);
     }
-    PostVisit(module);
+    PostVisit(module, result);
+    return result;
```
markww, line 55: If you're going to have `FoldResult`, I think it would make more sense to just have `FoldResult` return `Result`, similar to the other visitor methods. It's more in line with how the functional `fold` behaves and more in line with the [style guide recommendations](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs). Also, you can use the `fold` utility we put in `//src/common/utils`.

Imagine you rewrote the above as follows:

```C++
Result Visit(const Module &module) final override {
  Result result = fold(module.blocks(), PreVisit(module), [&this](Result acc, const std::unique_ptr<Block> &block) {
    return FoldResult(*block, block->Accept<Derived, Result>(*this), std::move(acc));
  });
  PostVisit(module, result);
  return result;
}
```

Now `FoldResult` can just take `result` as a value:
```C++
virtual Result FoldResult(const Block &block, Result child_result, Result acc) {
  return acc;
}
```

You might also want to consider refactoring `PostVisit` to just return its result rather than taking a mutable reference parameter.

markww (commented):
Looks good overall. My big quibbles:

1. The import process will not like `//visibility:public`.
2. We should use `void` instead of `std::monostate` and should not assume that the visitor's default value is sensibly provided via the no-arg constructor.
3. I think there's opportunity to make the code more functional-style by eliminating mutable reference parameters.

rust/tools/authorization-logic/BUILD
Cypher1: Thank you! That's what I was hoping the 'public' setting would do :)

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
```diff
@@ -30,13 +31,13 @@
 namespace raksha::backends::policy_engine::souffle {
 
 class DatalogLoweringVisitor
-    : public ir::IRTraversingVisitor<DatalogLoweringVisitor> {
+    : public ir::IRTraversingVisitor<DatalogLoweringVisitor, std::monostate> {
```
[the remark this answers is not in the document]
Cypher1: The main problem with `void` is that it can't be passed as an argument to a function, except by pointer.
It makes things more complicated, but is possible, I just need to workout the template magic to make the ergonomics better.

Alternatively, I have a change which splits the return from the side effects so that we can have `void` functions and return values via different names. I had actually pushed it, but for some reason Github was showing the earlier change.

I'll investigate this.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,53 @@ namespace raksha::ir {
...
-template <typename Derived>
-class IRTraversingVisitor : public IRVisitor<Derived> {
+template <typename Derived, typename Result=std::monostate>
+class IRTraversingVisitor : public IRVisitor<Derived, Result> {
  public:
   virtual ~IRTraversingVisitor() {}
 
   // Invoked before all the children of `module` is visited.
-  virtual void PreVisit(const Module& module) {}
+  virtual Result PreVisit(const Module& module) { return Result(); }
```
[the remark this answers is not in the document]
Cypher1: Agreed, if your value is not default construct-able it should be trivial to write your particular construction as overrides of `PreVisit`. If all three should be the same you can even

```
template<typename T>
Result PreVisit(const T& _) {
  return Result(special constructor stuff);
}
```

I think this is better than passing around lambdas.

src/ir/ir_traversing_visitor_test.cc
Cypher1: Thanks, this is copied code, would you like me to rename both my copy and the original?

src/ir/ir_traversing_visitor.h
Cypher1: This is originally how I had it. The issue I saw was that this requires moves of arbitrarily large result objects.
I'm happy to change it back though :)

src/ir/dot/dot_generator.cc
```diff
@@ -264,11 +264,13 @@ void DotGeneratorHelper::PreVisit(const Operation& operation) {
       AddEdge(GetNodeName(stored_value->storage()), input_name);
     }
   }
+  return {};
```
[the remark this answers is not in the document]
Cypher1: *nod* I can see that it could confuse some, though I feel default construction should be well known.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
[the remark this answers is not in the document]
Cypher1: The main problem with `void` is that it can't be passed as an argument to a function, except by pointer.
It makes things more complicated, but is possible, I just need to workout the template magic to make the ergonomics better.

Alternatively, I have a change which splits the return from the side effects so that we can have `void` functions and return values via different names. I had actually pushed it, but for some reason GitHub was showing the earlier change.

I'll investigate this.

src/ir/ir_traversing_visitor.h
Cypher1: Ohh, actually I forgot that not everything *is* move-able. This further motivates applying the `child_result` to the `result` via side effects, but I'm still open to either option (a `unique_ptr` to the result value would be fine to avoid the moves).

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,68 @@ namespace raksha::ir {
 // A visitor that also traverses the children of a node and allows performing
 // different actions before (PreVisit) and after (PostVisit) the children are
 // visited. Override any of the `PreVisit` and `PostVisit` methods as needed.
-template <typename Derived>
-class IRTraversingVisitor : public IRVisitor<Derived> {
+template <typename Derived, typename Result=std::monostate>
+class IRTraversingVisitor : public IRVisitor<Derived, Result> {
```
aferr, line 29: Is it useful to give some way of specifying a default value for the returned result that you can pass before visiting starts? For example, if the point of the visitor is to compute a set, you might want to make the default result the empty set or a set containing some specific element. You might want to make this default result value something that subclasses of `IRTraversingVisitor` can override or define some other way.

Or do you have this somehow and I missed it?

aferr (commented):
I looked at this but wanted to explicitly mention that I don't think I have a lot to add on top of what you're already getting out of @markww 's review

src/ir/ir_traversing_visitor_test.cc
Cypher1: Fixed both :)

src/ir/ir_traversing_visitor.h
Cypher1: This is what I hope to get from defaulting to returning the default constructed value. This will give us empty set/vector/maps/zeros etc. for many different types without any override.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,53 @@ namespace raksha::ir {
 // A visitor that also traverses the children of a node and allows performing
 // different actions before (PreVisit) and after (PostVisit) the children are
 // visited. Override any of the `PreVisit` and `PostVisit` methods as needed.
-template <typename Derived>
-class IRTraversingVisitor : public IRVisitor<Derived> {
+template <typename Derived, typename Result=std::monostate>
```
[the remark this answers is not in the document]
Cypher1: Lets continue discussing this elsewhere.

src/ir/ir_traversing_visitor.h
markww: The thing about modern C++ is that it's good at avoiding copies mechanically, even if they appear to occur semantically. C++17 has copy elision, which basically says that if no live name refers to an object at the point when it is copied, the compiler is free to make the copy not occur. Note that I'm using an expansive definition of "name" here, including things like local variable names, field names, pointers to the value, the value being in a container, etc.

While we are stuck with C++'s legacy behavior and a very conservative notion of when a name is live, you can use `std::move` as a hint to the compiler to indicate that a name is no longer live. Using this judiciously can allow you to eliminate the lifetime concerns of passing around pointers while also not eating the performance penalty of a copy.

src/ir/ir_traversing_visitor.h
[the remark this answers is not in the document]
markww: I spoke a little loosely in my comment. When I said "hand it a lambda" I really meant "provide it with a function somehow". The actual recommendation that I think makes more sense is to have a

```C++
Result GetDefaultValue() const = 0;
```

in the visitor, which forces the derived classes to indicate how they want to populate default values for the types. It would then be very easy to do this in a derived class for `void`:

```C++
void GetDefaultValue() const override {}
```

markww:
You know, I'm coming around to your thinking on `std::monostate` instead of `void`. It really does work nicely as a value that can be folded and passed around.

Thinking about it more, my one concern is that it is a bit of a spurious dependency to grab the `<variant>` header just for `std::monostate`. In addition, `std::monostate` is such a silly name that it requires some documentation reading. Perhaps it would be nicer to just have a `Unit` class that we define either in this visitor or in `//src/common/utils` and use that instead of `std::monostate`.

I still think it would be best to get rid of the assumption of a default constructor though. C++ pointer types, reference types, and primitive types do not have default constructors, and it would be unfortunate to exclude them from being returned by the visitor.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
```diff
@@ -49,7 +49,7 @@ static DatalogAttributePayload GetPayloadForAttribute(ir::Attribute attr,
   return DatalogAttribute::String("");
 }
 
-void DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
+std::monostate DatalogLoweringVisitor::PreVisit(const ir::Operation &operation) {
```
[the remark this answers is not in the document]
Cypher1: I've fixed up the `IRVisitor` and IRTraversingVisitor to now be backwards compatible, aliased `std::monostate` as `Unit` and ensured that `IRVisitor` defaults to `void`, so I think these are resolved, except for the default constructor.

If the type is not correct to be default constructed, the writer  will have to provide an alternate construction for their use case via an override (or changing their default constructor to be 'better behaved'. I don't see a way around this.

Marking this resolved. Feel free to reopen if there's other issues.

src/backends/policy_engine/souffle/datalog_lowering_visitor.h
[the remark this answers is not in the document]
Cypher1: I think this is resolved now with the backwards compatibility change and your approval in chat or `std::monostate` via the  `Unit` alias (added in `src/common/utils/types.h`.

src/ir/ir_traversing_visitor.h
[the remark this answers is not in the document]
Cypher1: I think the differences between these are:
- naming (`PreVisit` vs `GetDefaultValue`)
- default behavior (default construction by default vs requiring implementation)
- side effects (i.e. is the 'start value' generated based on a pre visit or a single default value)

Personally, I think having a default constructed value is a good ...default, and this default doesn't preclude start values that are set by Visitor implementations. It just saves typing in the default case.

Additionally, I think it's confusing to have a 'GetDefaultValue' that is different from the `PreVisit` result value.

Interested in @bgogul and @Aferr 's opinions on this though.

src/ir/ir_traversing_visitor.h
Cypher1: Thanks! Done!

src/ir/ir_traversing_visitor.h
Cypher1: Noting here as well just for completeness:

This is also supported by overriding the method `PreVisit`. i.e.

```
template<typename T>
Return PreVisit(const T& _ignored) {
    return Return(/*special arguments*/);
}
```

The benefit of this version, is that the return can be based on `_ignored` and can be different for different structures (e.g. `Module`, `Block`, `Operation`).

I hope I've covered everything you've asked about.

I think this PR is now ready for a full review.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
-class IRTraversingVisitor : public IRVisitor<Derived> {
+template <typename Derived, typename Result = Unit>
+class IRTraversingVisitor : public IRVisitor<Derived, Result> {
  public:
   virtual ~IRTraversingVisitor() {}
 
   // Invoked before all the children of `module` is visited.
-  virtual void PreVisit(const Module& module) {}
+  virtual Result PreVisit(const Module& module) {
+    return Result();
```
markww, line 35: I still think that this is not a good idea. If this visitor is used to translate from the IR to another tree-like structure, the superclass of all nodes within that structure being `Node`, it could potentially be natural to instantiate the visitor with `const Node *`. If you attempt to default-construct `Result` like this, that visitor cannot be instantiated with this class as-is.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
-    PreVisit(module);
+  Result Visit(const Module& module) final override {
+    Result result = PreVisit(module);
     for (const std::unique_ptr<Block>& block : module.blocks()) {
-      block->Accept(*this);
+      result =
+          FoldResult(*block, block->Accept<Derived, Result>(*this), std::move(result));
     }
-    PostVisit(module);
+    return PostVisit(module, std::move(result));
```
markww, line 76: Why not use `fold` from our `utils` package for this and other folds in the visitor? You could write the whole function like so:

```C++
Result pre_visit_result = PreVisit(module);
Result fold_result = fold(module.blocks(), std::move(pre_visit_result),
       [&this](Result acc, const std::unique_ptr<Block> &block) {
         return FoldResult(*block, block->Accept<Derived, Result>(*this), std::move(acc));
       });
return PostVisit(module, std::move(fold_result));
```

This gets rid of the loop and the mutable `result` local. Each of these locals, of course, could be inlined to get rid of the `std::move`, but this is closest to your current form (and possibly the most readable form).

markww (commented):
I'm down to basically style nits and the default constructor thing.

I know that implementing an additional method is ugly, but not every object in C++ can be default constructed. In particular, a very idiomatic C++ pattern would be to return something like a sum type from this visitor by returning a pointer to a base type. Because pointers do not have constructors, this use case cannot be fit into the current way this visitor pattern is written.

src/ir/ir_traversing_visitor.h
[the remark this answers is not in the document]
markww: As I mentioned in a later comment, that isn't the only difference. Certain types that we may wish to return from this visitor (pointers, primitives, types where the default constructor is explicitly deleted) cannot be default constructed. In particular, it is highly idiomatic in C++ to do something akin to returning a sum type by returning a pointer to a base type. Because of that restriction, as currently written, this visitor cannot handle all of the patterns that we may wish to use it for. While ugly, the `GetDefaultValue` method would address this.

src/ir/ir_traversing_visitor_test.cc
```diff
@@ -132,25 +138,176 @@ TEST(IRTraversingVisitorTest, TraversesOperationAsExpected) {
...
+  CollectingVisitor all_order_visitor(CollectingVisitor::TraversalType::kBoth);
   plus_op_instance.Accept(all_order_visitor);
   EXPECT_THAT(
       all_order_visitor.nodes(),
       testing::ElementsAre(std::addressof(plus_op_instance), plus_op_module,
                            plus_op_module, std::addressof(plus_op_instance)));
 }
+}  // namespace
+
+namespace {
```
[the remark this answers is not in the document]
Cypher1: The two sets of tests don't use any overlapping classes or values and doing so would be accidental (as they're intended to test the mutating and non-mutating visitors separately). This enforces that more easily without making things harder to find (like a new file would).

Happy to follow your lead on this one though.

src/ir/ir_traversing_visitor.h
Cypher1: One... 'very good reason'

...I didn't know about it! :)

Thanks great. I'll change it right away

src/ir/ir_traversing_visitor.h
Cypher1: Reaching out via chat to try to close the loop on this, but, so that I'm not leaving your comment hanging:

One reason `PreVisit` is `virtual` is so that this is not required behavior.
Child classes can override `PreVisit` to get different behavior.

Is the problem that you're pointing to that they would need to override all three `PreVisit` functions (which can be done via templating the argument?)

You mention a pointer as an example of a case where the `Result` cannot be default  constructed, but I'd say it's also an example where the pointer cannot be safely constructed without knowledge of which type is being visitws.

Keen to hear your thoughts

src/ir/ir_visitor.h
```diff
@@ -24,13 +24,13 @@ class Operation;
 
 // An interface for the visitor class. We will also pass in the `Derived` class
 // as template argument if we want to support CRTP at a later point.
-template <typename Derived>
+template <typename Derived, typename Result=void>
```
bgogul, line 27: Let us keep this consistent with `IRTraversingVisitor` and make it `Result=Unit`?

src/ir/module.h
```diff
@@ -57,9 +57,9 @@ class Operation {
 
   std::string ToString(SsaNames& ssa_names) const;
 
-  template <typename Derived>
-  void Accept(IRVisitor<Derived>& visitor) const {
-    visitor.Visit(*this);
+  template <typename Derived, typename Result=void>
```
bgogul, line 60: I don't think you need default `Result` to void here and below. This is always called from a visitor, which will have a `Result`.

src/ir/ir_traversing_visitor.h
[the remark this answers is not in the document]
bgogul: I should have caught up on the thread before replying. It looks like what Mark is saying is the following :

```C++
class IRTraversingVisitor<...> {
  Result GetDefaultValue() const = 0;
   virtual Result PreVisit(const Module& module) { return GetDefaultResult(); }
 
}
```

Right?  I am even OK with the following:

```C++
class IRTraversingVisitor<...> {
  Result GetDefaultValue() const { return Result() };
   virtual Result PreVisit(const Module& module) { return GetDefaultResult(); }
 
}
```

If the default `GetDefaultValue()` is not sufficient for the `Result` type in question, we will only have to update it in one place?

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
 
   // Invoked before all the children of `module` is visited.
-  virtual void PreVisit(const Module& module) {}
+  virtual Result PreVisit(const Module& module) {
+    return Result();
+  }
   // Invoked after all the children of `module` is visited.
-  virtual void PostVisit(const Module& module) {}
+  virtual Result PostVisit(const Module& module, Result result) {
+    return result;
```
bgogul, line 46: Does this need to be `std::move(result)` if we want to take advantage of zero copy? @markww @Cypher1 

(This is one of the reasons why I don't quite like `std::move`. There is too much of a cognitive overload that we have to be careful everywhere to make sure we have zero copies. )

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
   virtual ~IRTraversingVisitor() {}
 
   // Invoked before all the children of `module` is visited.
-  virtual void PreVisit(const Module& module) {}
+  virtual Result PreVisit(const Module& module) {
+    return Result();
+  }
   // Invoked after all the children of `module` is visited.
-  virtual void PostVisit(const Module& module) {}
+  virtual Result PostVisit(const Module& module, Result result) {
```
bgogul, line 45: Somehow make it clear that `result` is from after the in-order traversal? e.g.,
```suggestion
  virtual Result PostVisit(const Module& module, Result in_order_result) {
```

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
+    return result;
+  }
+  virtual Result FoldResult(const Module& module, Result child_result,
+                            Result result) {
```
bgogul, line 49: Documentation?  Is this for folding the children of a `module`? Or, is `module` the child?

Also, is `result` the accumulated results so far? Can you name it appropriately? Also, a different order of argument might be better? e.g., 

```suggestion
  virtual Result FoldResult(Result accumulated_result, const Module& child, Result child_result) {
```

Whatever you decide here, please apply to other methods as well.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
-  virtual void PreVisit(const Operation& operation) {}
+  virtual Result PreVisit(const Operation& operation) {
+    return Result();
+  }
   // Invoked after all the children of `operation` is visited.
-  virtual void PostVisit(const Operation& operation) {}
+  virtual Result PostVisit(const Operation& operation, Result result) {
+    return result;
+  }
+  virtual Result FoldResult(const Operation& operation, Result child_result,
```
bgogul, line 72: See other FoldResult

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +25,74 @@ namespace raksha::ir {
...
+  }
   // Invoked before all the children of `operation` is visited.
-  virtual void PreVisit(const Operation& operation) {}
+  virtual Result PreVisit(const Operation& operation) {
+    return Result();
+  }
   // Invoked after all the children of `operation` is visited.
-  virtual void PostVisit(const Operation& operation) {}
+  virtual Result PostVisit(const Operation& operation, Result result) {
+    return result;
```
bgogul, line 70: Do we need `return std::move(result)`?

src/ir/ir_traversing_visitor.h
Cypher1: Added `GetDefaultValue` and set `PreVisit` to return it by default.
I'm not sure this is the right call (as not all types *can* have a meaningful default value), but it's not a deal breaker :)

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +26,80 @@ namespace raksha::ir {
...
-  virtual void PreVisit(const Block& block) {}
+  virtual Result PreVisit(const Block& block) {
+    return GetDefaultValue();
+  }
   // Invoked after all the children of `block` is visited.
-  virtual void PostVisit(const Block& block) {}
+  virtual Result PostVisit(const Block& block, Result result) {
+    return result;
+  }
+  virtual Result FoldResult(const Block& block, Result child_result,
```
bgogul, line 60: See other `FoldResult`

src/ir/ir_traversing_visitor_test.cc
```diff
@@ -132,25 +138,176 @@ TEST(IRTraversingVisitorTest, TraversesOperationAsExpected) {
...
+using ResultType=std::vector<void*>;
+
+// Another small test visitor that collects the nodes as they are visited to make
+// sure that  Pre and Post visits happened correctly.
+// Uses 'returns' rather than side effects.
+class ReturningVisitor : public IRTraversingVisitor<ReturningVisitor, ResultType> {
```
bgogul, line 168: Not a huge deal, but I wonder if there is a way to avoid code duplication here. The code and the tests are quite similar to `CollectingVisitor`.

src/ir/ir_visitor.h
Cypher1: Done (surprised that didn't break anything :D )

bgogul (commented):
Left some comments. Nothing major. PTAL.

src/ir/ir_traversing_visitor_test.cc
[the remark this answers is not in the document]
Cypher1: Guessing that the second namespace is more confusing than it's worth. Changing for now, can change back if I am wrong about this. Cheers

src/ir/ir_traversing_visitor.h
Cypher1: There were some other instances to change so I've made those renames myself, thanks for the diff though

src/ir/ir_traversing_visitor.h
Cypher1: Thanks Gogul, this makes sense to me. I added comments and reordered the arg. Didn't change `PostVisit` as I thought it would be confusing in the `Unit` case.

src/ir/ir_traversing_visitor_test.cc
Cypher1: If we could template the tests on the visitor this could work. I've deduplicated what I could and will look into it.

Cypher1: Ah, on second thought the visitors actually have different base types and different members. I don't know how to do this (Initially I thought we could have them both based on another class which exposed the nodes, but I don't think this works).

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +26,89 @@ namespace raksha::ir {
...
   virtual ~IRTraversingVisitor() {}
 
+  // Gives a default value for all 'PreVisit's to start with.
+  // Should be over-ridden if the Result is not default constructable.
+  virtual Result GetDefaultValue() { return Result(); }
+
   // Invoked before all the children of `module` is visited.
-  virtual void PreVisit(const Module& module) {}
+  virtual Result PreVisit(const Module& module) {
+    return std::move(GetDefaultValue());
```
bgogul, line 40: `std::move` is not needed as `GetDefaultValue()` does not have a name. I just remembered  https://abseil.io/tips/77. It is a good read and gives a simple mental model for `std::move`.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +26,89 @@ namespace raksha::ir {
...
   // Invoked after all the children of `module` is visited.
-  virtual void PostVisit(const Module& module) {}
+  virtual Result PostVisit(const Module& module, Result in_order_result) {
+    return std::move(in_order_result);
+  }
+  // Used to accumulate child results from the Module's children.
+  // Should discard or merge `child_result` into the `accumulator`.
+  virtual Result FoldResult(Result accumulator, const Module& module,
+                            Result child_result) {
+    return std::move(accumulator);
```
bgogul, line 50: It looks like we may not need `std::move` here and in almost all the other places according to https://abseil.io/tips/77

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +26,89 @@ namespace raksha::ir {
...
+    return std::move(GetDefaultValue());
+  }
   // Invoked after all the children of `block` is visited.
-  virtual void PostVisit(const Block& block) {}
+  virtual Result PostVisit(const Block& block, Result in_order_result) {
+    return std::move(in_order_result);
+  }
+  // Used to accumulate child results from the Block's children.
+  // Should discard or merge `child_result` into the `accumulator`.
+  virtual Result FoldResult(Result accumulator, const Block& block,
```
bgogul, line 62: Instead of `block`, name it `parent_block` or simply `parent`? That will serve as additional documentation.

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +26,89 @@ namespace raksha::ir {
...
+    return std::move(GetDefaultValue());
+  }
   // Invoked after all the children of `operation` is visited.
-  virtual void PostVisit(const Operation& operation) {}
+  virtual Result PostVisit(const Operation& operation, Result in_order_result) {
+    return std::move(in_order_result);
+  }
+  // Used to accumulate child results from the Operation's children.
+  // Should discard or merge `child_result` into the `accumulator`.
+  virtual Result FoldResult(Result accumulator, const Operation& operation,
```
bgogul, line 76: `operation` -> `parent_operation` or `parent`.

bgogul (approved):
Looks good. A few minor things:
  - According to https://abseil.io/tips/77, some of the `std::move` may be redundant.
  -  Some minor naming suggestions. 

May be, let @markww have a quick look before you marking it "ready to pull"?

src/ir/ir_traversing_visitor.h
```diff
@@ -24,47 +26,89 @@ namespace raksha::ir {
...
-template <typename Derived>
-class IRTraversingVisitor : public IRVisitor<Derived> {
+template <typename Derived, typename Result = Unit>
+class IRTraversingVisitor : public IRVisitor<Derived, Result> {
  public:
   virtual ~IRTraversingVisitor() {}
 
+  // Gives a default value for all 'PreVisit's to start with.
+  // Should be over-ridden if the Result is not default constructable.
+  virtual Result GetDefaultValue() { return Result(); }
```
markww, line 36: This looks good, but you may run into compilation errors if this is instantiated with a `Result` type without a default constructor.

It looks like overriding `GetDefaultValue` is currently untested. I think you should make a (very simple) visitor to try out returning a type with the no-arg constructor explicitly deleted. So something like:

```C++
class WrappedInt {
  public:
  WrappedInt() = delete;
  WrappedInt(uint64_t val) : val_(val) {}
  
  int64_t operator*() const {
    return val;
  }
  
  private:
  uint64_t val_;
};
```

And just make a visitor using that to count the nodes in the tree.

markww (approved):
Looks good modulo the additional test for `GetDefaultValue` with a `delete`d default constructor and removing extraneous `std::move`s on temporaries.

src/ir/ir_traversing_visitor.h
markww: If you do run into compilation errors with this default `GetDefaultValue`, you can still have a nice default like you have here, but unfortunately, I think getting it may rely on dark template magic:

```C++
template<class Result, class Enable>
struct DefaultValueGetter {
  Result Get() { static_assert("Default value override is required!"); }
}

template<class Result>
struct DefaultValueGetter<Result, std::enable_if_t<std::is_default_constructible_v<Result>>> {
  Result Get() { return Result(); }
};

virtual Result GetDefaultValue() { return DefaultValueGetter<Result>::Get(); }
```

src/ir/ir_traversing_visitor.h
Cypher1: Gotcha, thanks

src/ir/ir_traversing_visitor.h
Cypher1: Ah, I see now, because the template generates the functions for the base class (even though they are shadowed by the `override` we have to provide a 'working' default constructor in this case. That is a pretty bad place for C++ to be in./
I'm sorry that I hadn't foreseen it.

I'll use this template approach, thanks!

Cypher1:
I think this is now ready to pull. Leaving un-pulled to let @markww do a final pass.

If you're have no follow up Mark, I'd appreciate it if you could add the label :)
Thanks for all the review comments and insight!

src/ir/ir_traversing_visitor.h
```diff
@@ -16,55 +16,93 @@
...
 
+template<class Result, class Enable = void>
+struct DefaultValueGetter {
+  static Result Get() { LOG(FATAL) << "Override required for non-default-constructible type."; }
+};
+
+template<class Result>
+struct DefaultValueGetter<Result, std::enable_if_t<std::is_default_constructible_v<Result>>> {
+  static Result Get() { return Result(); }
+};
```
bgogul, line 34: Move this into the `IRTraversingVisitor` class to avoid namespace collisions.

src/ir/ir_traversing_visitor.h
```diff
@@ -16,55 +16,93 @@
...
+#include "src/common/utils/types.h"
 #include "src/ir/ir_visitor.h"
 #include "src/ir/module.h"
 
 namespace raksha::ir {
 
+template<class Result, class Enable = void>
+struct DefaultValueGetter {
+  static Result Get() { LOG(FATAL) << "Override required for non-default-constructible type."; }
+};
```
bgogul, line 29: Leave this for now, but I wonder if we can make this a compile time error rather than runtime error. Can you file an issue?

 I found some tricks to make SFINAE errors more readable. See https://gracicot.github.io/tricks/2017/07/01/deleted-function-diagnostic.html. Can you also mention this link in the filed issue?

src/ir/ir_traversing_visitor_test.cc
```diff
@@ -24,38 +24,81 @@
...
+  WrappedInt() = delete;
+  WrappedInt(uint64_t val) : val_(val) {}
+
+  int64_t operator*() const { return val_; }
+
+ private:
+  uint64_t val_;
+};
+
+// Just a small test visitor that collects the nodes as they are visited to make
```
bgogul, line 38: stale comment

src/ir/ir_traversing_visitor_test.cc
```diff
@@ -24,38 +24,81 @@
...
+
+TEST(IRTraversingVisitorTest,
+     GetDefaultValueWorksForNonDefaultConstructableValues) {
+  Module global_module;
+  global_module.AddBlock(std::make_unique<Block>());
+  global_module.AddBlock(std::make_unique<Block>());
+
+  NonDefaultConstructorVisitor counting_visitor;
+  const WrappedInt result = global_module.Accept(counting_visitor);
+  EXPECT_THAT(*result, testing::Eq(3));
```
bgogul, line 61: `EXPECT_EQ(*result, 3);` is more straightforward.

bgogul (approved):
Looks good.

src/ir/ir_traversing_visitor.h
Cypher1: Opened https://github.com/google-research/raksha/issues/618 and added comment with link.

src/ir/ir_traversing_visitor_test.cc
Cypher1: Updated

Cypher1:
Resolved comments, just letting tests run and I'll merge :)

PR #599, opened 2022-07-01
Fix Bazel canonical build warnings for two external deps.

shaper:
Add sha256 for `hedron_compile_commands` and `com_google_googletest`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -5,6 +5,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 #--------------------
 http_archive(
     name = "com_google_googletest",
+    sha256 = "5cf189eb6847b4f8fc603a3ffff3b0771c08eec7dd4bd961bfd45477dd13eb73",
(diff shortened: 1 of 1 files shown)


shaper:
CI is failing though it works locally. Attempting to view gives me permission-access error for `raksha-ci`, maybe it's just not letting me run CI due to permissions?

shaper:
Fixed buildifier error, PTAL. Still need CI access from whomever can grant it, though.

markww:
@bgogul , what do we need to do to grant @shaper CI access?

markww (approved):
This PR looks good to me. One issue though: you've submitted this PR from an external fork rather than from a dev branch within google-research/raksha. I don't know whether that will work for most PRs with our Copybara setup. Fortunately, this PR has no internal footprint, so I think there shouldn't be an issue with you just merging it.

bgogul:
> One issue though: you've submitted this PR from an external fork rather than from a dev branch within google-research/raksha

It does not matter. Copybara gets triggered if google/main is the target of the merge. It doesn't matter what the source is. We have been creating branches on google-research/raksha itself mainly to facilitate dependent PRs. If you don't need it, you can totally create branches on a fork.

PR #600, opened 2022-07-02
Fix Bazel warning on unsafe function used to create temp dirs.

shaper:
Replace use of `tmpnam` with `mkdtemp`. Example warning prior to this change:

  warning: the use of `tmpnam` is dangerous, better use `mkstemp`

### DIFF
--- a/src/backends/policy_engine/souffle/raksha_datalog_facts_test.cc
+++ b/src/backends/policy_engine/souffle/raksha_datalog_facts_test.cc
@@ -133,11 +133,9 @@ TEST_F(DumpFactsToDirectoryTest, CreatesEmptyRelationsIfRequested) {
 TEST_F(DumpFactsToDirectoryTest, CreatesFilesInRequestedDirectory) {
   std::vector<std::string> empty_relations = {"check", "claim"};
 
-  std::filesystem::path desired_path = std::tmpnam(nullptr);
-  std::error_code error_code;
-  bool tmp_dir_created =
-      std::filesystem::create_directory(desired_path, error_code);
-  ASSERT_TRUE(tmp_dir_created);
+  char filename[] = "/tmp/foo.XXXXXX";
+  ASSERT_NE(::mkdtemp(filename), nullptr);
+  std::filesystem::path desired_path(filename);
 
   absl::Status status =
       facts_.DumpFactsToDirectory(desired_path, empty_relations);
@@ -149,10 +147,14 @@ TEST_F(DumpFactsToDirectoryTest, CreatesFilesInRequestedDirectory) {
   EXPECT_THAT(files, testing::UnorderedElementsAre(
                          "isOperation.facts", "check.facts", "claim.facts"));
   CheckFileContentsInDirectory(desired_path, empty_relations);
+
+  ASSERT_TRUE(std::filesystem::remove_all(desired_path));
 }
 
 TEST_F(DumpFactsToDirectoryTest, ReturnsErrorIfRequestedDirectoryDoesNotExist) {
-  std::filesystem::path desired_path = std::tmpnam(nullptr);
+  std::filesystem::path desired_path("nosuchpath");
+  ASSERT_FALSE(std::filesystem::exists(desired_path));
+
   absl::Status status = facts_.DumpFactsToDirectory(desired_path, {});
   EXPECT_FALSE(status.ok());
   EXPECT_EQ(status.code(), absl::StatusCode::kFailedPrecondition);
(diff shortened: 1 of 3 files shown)


src/common/utils/filesystem.cc
```diff
@@ -34,13 +35,15 @@ absl::StatusOr<std::filesystem::path> CreateTemporaryDirectory() {
         "Unable to access temporary file directory: %s", error_code.message()));
   }
 
-  absl::string_view temp_file = std::tmpnam(nullptr);
-  std::filesystem::path new_temp_path = temp_path / temp_file;
-  if (!std::filesystem::create_directory(new_temp_path, error_code)) {
+  const int BUFFER_SIZE = 1024;
```
shaper, line 38: This feels old-school but I looked at `string_view` and `std::string` and in the end since `mkdtemp` needs a non-const `char *`, whatever we do we need to build that, and this seemed not worse than other approaches, but I welcome alternatives.

It also seems silly there isn't a cleaner C++ filesystem lib way to create a temporary directory with a known unique name, but this seems to be the thing, or at least one version of it, again feedback welcome.

The complaint seems to come from `g++` so we could also maybe turn it off, but one presumes it's well intentioned.

markww: The warning on `tmpnam` comes from the fact that the operating system tries really hard to guess a unique name, but it cannot guarantee that it truly is unique nor that someone else just happened to call `tmpnam` or similar and create that name between when you call `tmpnam` and when you use it. In general, that opens the door to TOCTOU attacks.

Now, TOCTOU attacks are bad, but I think this might not fix the underlying issue. We are only using the temporary directory so that we can communicate with Souffle by handing it a directory with input facts and reading the output facts from that directory. Because we are creating the directory so that we can, at a later point, read the output of some other program that was placed in files in that directory, the TOCTOU window is probably wide open regardless of how atomically the directory itself is created. If some other program accidentally or maliciously places `.fact` files over the Souffle output, we would trustingly read them as if they were the Souffle output files.

This solution does make the creation of the directory atomic, but it relies upon unix-specific functions (`mkdtemp`), and so wouldn't work on other systems (say, a Windows box) while this less-atomic system is more portable.

In the future, we'd like to talk to Souffle entirely through an in-memory interface, which would sidestep the TOCTOU concerns (and probably be a lot faster too). We'd likely eliminate the temporary directory entirely, or use it only for debugging. So I'm not sure if this change is worth the portability and readability cost.

src/common/utils/filesystem.cc
```diff
@@ -34,13 +35,15 @@ absl::StatusOr<std::filesystem::path> CreateTemporaryDirectory() {
...
+  char new_temp_path[BUFFER_SIZE];
+  snprintf(new_temp_path, BUFFER_SIZE, "%s/XXXXXX", temp_path.c_str());
+  if (::mkdtemp(new_temp_path) == nullptr) {
     return absl::FailedPreconditionError(
         absl::StrFormat("Unable to create directory `%s`: %s", new_temp_path,
-                        error_code.message()));
+                        strerror(errno)));
   }
+
   return new_temp_path;
```
markww, line 47: This line will return `new_temp_path`, which is a pointer to stack space now that `new_temp_path` is a `char` array, and thus looks a lot like it introduces a memory error. It may be that an implicit conversion to `std::filesystem::path` just does the right thing, but this should be rewritten to make that more obvious.

src/backends/policy_engine/souffle/raksha_datalog_facts_test.cc
```diff
@@ -133,11 +133,9 @@ TEST_F(DumpFactsToDirectoryTest, CreatesEmptyRelationsIfRequested) {
 TEST_F(DumpFactsToDirectoryTest, CreatesFilesInRequestedDirectory) {
   std::vector<std::string> empty_relations = {"check", "claim"};
 
-  std::filesystem::path desired_path = std::tmpnam(nullptr);
-  std::error_code error_code;
-  bool tmp_dir_created =
-      std::filesystem::create_directory(desired_path, error_code);
-  ASSERT_TRUE(tmp_dir_created);
+  char filename[] = "/tmp/foo.XXXXXX";
```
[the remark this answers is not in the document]
markww: `/tmp` is not guaranteed to be the temporary directory. The temporary directory should be gotten via `std::filesystem::temp_directory_path(error_code);`.

src/common/utils/filesystem.cc
shaper: Rewrote to use `absl::StrFormat`.

src/backends/policy_engine/souffle/raksha_datalog_facts_test.cc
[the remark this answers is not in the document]
shaper: Reworked to use filesystem tmpdir utility routine.

src/common/utils/filesystem.cc
shaper: Rewrote this to use `std::string` and `absl::StrFormat`.

bgogul (approved):

Issue #601, opened 2022-07-02
Update to Souffle v2.3.


shaper:
There were some changes @asejfia was interested in post-v2.1 we're using now, and we foresee some other changes we're upstreaming to Souffle, so being closer to the most recent Souffle release seems helpful.

We'd need to update both externally and internally.

I started exploring what's involved as a way to get up to speed, so I'm filing this to track the work. Feel free to suggest we wait if you'd rather not have this happen.

FYI @markww @bgogul

shaper:
Update -- the internal change to upgrade to v2.3 is sent, and I've emailed the Kythe folks to discuss landing it.

shaper:
The internal change is submitted.

PR #603, opened 2022-07-02
Upgrade to Souffle v2.3 from v2.1.

shaper:
External part of #601

### DIFF
--- a/rust/tools/authorization-logic/BUILD
+++ b/rust/tools/authorization-logic/BUILD
@@ -134,11 +134,14 @@ filegroup(
     srcs = glob(["raksha_examples/**"]),
 )
 
-# Specifing `@souffle` as part of `data` in the `authorization_logic_test` target
+# Specifying `@souffle` as part of `data` in the `authorization_logic_test` target
 # triggers a build of souffle with `-fno-exceptions`, which causes the build
 # to fail. Instead, we extract the souffle binary using a genrule by making
 # it part of the `tools` (which does not trigger a rebuild) and use it when
 # running the authorization logic tests via `authorization_logic_test`.
+#
+# TODO(#607): Investigate warning as below on build:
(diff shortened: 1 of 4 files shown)


rust/tools/authorization-logic/BUILD
```diff
@@ -134,11 +134,14 @@ filegroup(
...
 )
 
-# Specifing `@souffle` as part of `data` in the `authorization_logic_test` target
+# Specifying `@souffle` as part of `data` in the `authorization_logic_test` target
 # triggers a build of souffle with `-fno-exceptions`, which causes the build
 # to fail. Instead, we extract the souffle binary using a genrule by making
 # it part of the `tools` (which does not trigger a rebuild) and use it when
 # running the authorization logic tests via `authorization_logic_test`.
+#
+# TODO(wkorman): Investigate warning as below on build:
```
bgogul, line 143: Can we file an issue and use that number in TODO? I think there is very straightforward fix, we should simply change the rule name to something like below:

```
genrule(
  name = "souffle_tool",
  outs = ["souffle"],
   ...
)
```

We will need to change the current uses of souffle though.

bgogul (approved):
Thanks, Walter!

@markww, I assume we are already out-of-sync with internal souffle that this change is fine?

rust/tools/authorization-logic/BUILD
markww: Yes, that's probably the right solution. And you are correct, we would need to update our usages of `souffle`.

markww:
Yes, no (additional) harm in becoming further out of sync with internal Souffle. But ideally, we should update internal Souffle to also be version 2.3.

rust/tools/authorization-logic/BUILD
shaper: Done, filed #607

markww (approved):

Issue #468, opened 2022-03-18
Emit Query Results To a Single CSV


aferr:
Currently, the results of queries are emitted as a collection of CSVs where for each query in the input auth logic program named `query_1, ..., query_n` there are CSVs called `query_1.csv, ..., query_n.csv` and the CSV will be empty if it is false or will contain just `dummy_var` if it is true. The directory where these CSVs are placed can be specified in the CLI, but the file names cannot be.

Two problems with this are:
 - the output file names and locations cannot be completely specified within the CLI
 - this interface makes writing bazel rules difficult -- if the locations of output files are completely contained within the command of a genrule, bazel can locate these files easily. If not, extra work needs to be done in the bazel rule to locate the files or place them somewhere they are recognized.

To fix this, the query outputs should be generated in a single CSV that contains the results of all the queries. If the queries in the input program are `query_1, ..., query_n`, where query_1 is true or query_n is false, the resulting CSV will look like:
```
query_1
...
```
where the false queries are absent from the CSV and the true ones are present

To implement this, the translation for queries which is [implemented here](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/lowering_ast_datalog.rs)
would need to change. Currently queries are translated into predicates with the same name as the query, and the queries have one parameter and they are true for the argument "dummy_var" if the query is true. A way to implement this change would be to instead use a single predicate (e.g. "queries") and use _arguments_ named after each specific query to indicate the truth value of each query (e.g. `queries("query_1") : ... ` is true if query_1 is true.

This one CSV can also be parsed into other formats like JSON as described here (https://github.com/google-research/raksha/issues/455)

aferr:
@bgogul @harsha-mandadi Since you were discussing how to represent queries for the C++ implementation on Monday, I wanted to CC you on this because I think the way I made queries work in the original implementation was not ideal, and this has some thoughts about how to do it better. You might as well not repeat my mistakes for the C++ version :)

Issue #464, opened 2022-03-17
The test //rust/tools/authorization-logic:authorization_logic_test may be nondeterministic.


markww:
Recently, I was running my tests to ensure everything was working as expected and was surprised to see that the test `//rust/tools/authorization-logic:authorization_logic_test` failed. In hopes of getting the failure message on a cleaner terminal output, I reran `bazel test ...`. After that, the test passed. Unfortunately, this also appears to have deleted the failing logs.

Because of the above behavior, it seems likely that `//rust/tools/authorization-logic:authorization_logic_test` is nondeterministic in some way, perhaps due to a dependency that Bazel does not have insight into. In any case, we should look into this if it pops up again.

Cypher1:
I ran
```
bazel test //rust/tools/authorization-logic:authorization_logic_test --runs_per_test=100
```
and got
```
FAILED in 55 out of 100 in 1.0s
```

So this is definitely flakey. Can look into it further but not sure where to start.

It looks like the failures are related to the file system but here's the numbers from
```
cat  /home/jopra/.cache/bazel/_bazel_jopra/ad57f266c437944f14d502fbbd0016f5/execroot/__main__/bazel-out/k8-fastbuild/testlogs/rust/tools/authorization-logic/authorization_logic_test/run_*_of_100/test.log | grep "panicked at" | sort | uniq -c
```
Sorted and counted:
```
5 thread 'test::test_dots_and_quotes::test::test_quotes_in_paths' panicked at 'failed to read input in input_to_souffle_file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', rust/tools/authorization-logic/src/compilation_top_level.rs:38:10
      1 thread 'test::test_multimic_no_overrides::test::test_multimic_no_overrides_scenario2' panicked at 'assertion failed: is_file_empty(&queryfile) != intended_result', rust/tools/authorization-logic/src/test/test_queries.rs:64:13
      2 thread 'test::test_multimic_overrides::test::test_multimic_overrides' panicked at 'assertion failed: is_file_empty(&queryfile) != intended_result', rust/tools/authorization-logic/src/test/test_queries.rs:64:13
     32 thread 'test::test_num_compare::test::test_basic_number_comparisons' panicked at 'failed to read input in input_to_souffle_file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', rust/tools/authorization-logic/src/souffle/souffle_interface.rs:45:10
[...]

PR #604, opened 2022-07-02
Fix Rust dead code warnings.

### LINKED ISSUE #602: `souffle_interface.rs` reports dead code on clean builds.
Clean build reports dead code per below, it looks like it's due to the interface being built into a `rust_library` for use by other targets. Capturing my findings for commentary from others. This is literally my first time interacting with Rust code directly myself.

https://stackoverflow.com/questions/25877285/how-to-disable-unused-code-warnings-in-rust suggestions include:

```
#[allow(dead_code)]
```

atop file/function/etc. with other options that seem less desirable/focused (setting allow on crate, passing allow to `rustc` via flag/env-var, etc.), so I'm thinking to try the above.

Sample warnings:

```
INFO: From Compiling Rust staticlib authorization_logic_lib (32 files) [for host]:
warning: function is never used: `store_new_keypair_cleartext`
  --> rust/tools/authorization-logic/src/signing/tink_interface.rs:56:8
   |
56 | pub fn store_new_keypair_cleartext(pub_key_file: &str, priv_key_file: &str) {
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: function is never used: `push_if_constant`
  --> rust/tools/authorization-logic/src/souffle/universe_translation.rs:68:4
   |
68 | fn push_if_constant(hash_set: &mut HashSet<String>, name: String) {
   |    ^^^^^^^^^^^^^^^^

warning: function is never used: `emit_souffle`
  --> rust/tools/authorization-logic/src/souffle/souffle_interface.rs:31:8
   |
31 | pub fn emit_souffle(filename: &str) {
   |        ^^^^^^^^^^^^

warning: function is never used: `input_to_souffle_file`
  --> rust/tools/authorization-logic/src/souffle/souffle_interface.rs:42:8
   |
42 | pub fn input_to_souffle_file(input_file_path: &str, output_file_path: &str) {
   |        ^^^^^^^^^^^^^^^^^^^^^

warning: function is never used: `run_souffle`
  --> rust/tools/authorization-logic/src/souffle/souffle_interface.rs:71:8
   |
71 | pub fn run_souffle(input_file_path: &str, queries_outdir: &str) {
   |        ^^^^^^^^^^^

warning: function is never used: `is_file_empty`
  --> rust/tools/authorization-logic/src/souffle/souffle_interface.rs:89:8
   |
89 | pub fn is_file_empty(filename: &str) -> bool {
   |        ^^^^^^^^^^^^^

warning: 6 warnings emitted
```

FYI @aferr @Cypher1

shaper:
Fixes #602.

Relates to #464 and #468.


rust/tools/authorization-logic/src/signing/tink_interface.rs
```diff
@@ -53,6 +53,7 @@ pub fn deserialize_from_file(filename: &str) -> Result<Vec<AstAssertion>, Error>
 // At present, this function is only used in tests, so a warning will be given
 // that this is not used. In the future, a script for generating keys might
 // also use this outside of tests.
+#[allow(dead_code)]
```
shaper, line 56: Code style wise I feel deviant here, in that I just sprinkled this on the one fn that is noted as dead, but perhaps style-wise we should say, if it's an interface and built in a library, we tag all fn's? Or, do it via allow-dead on the BUILD library definition (but that seems wider-swathe and we'd risk missing more).

So I went for minimal intrustion, but feedback is welcome (see issue linked in PR description for more background).

rust/tools/authorization-logic/src/souffle/souffle_interface.rs
```diff
@@ -28,6 +29,7 @@ use crate::{
 /// Given a filename containing policy code this function parses it,
 /// constructs an AST, translates the AST to the datalogIR, and emits a string
 /// contianing souffle code.
+#[allow(dead_code)]
```
shaper, line 32: StackOverflow makes it sound like there's a way to state that one wants all code/fns in a file to have this allow, but I tried at top of file and it didn't seem to work.

rust/tools/authorization-logic/src/souffle/universe_translation.rs
```diff
@@ -65,12 +65,6 @@ fn push_option_to_vec<T>(vec: &mut Vec<T>, elt: Option<T>) {
     }
 }
 
-fn push_if_constant(hash_set: &mut HashSet<String>, name: String) {
```
shaper, line 68: This seems like a legitimately dead fn, @aferr if you'd like to keep it let me know and we can tag it.

rust/tools/authorization-logic/src/test/test_queries.rs
```diff
@@ -66,7 +66,8 @@ pub mod test {
 
     pub fn setup_dirs_and_run_query_test(t: QueryTest) {
         utils::setup_directories_for_bazeltest(vec![&t.input_dir], vec![&t.output_dir]);
-        run_query_test(t)
+        run_query_test(t);
+        clean_test_dir()
```
shaper, line 70: @aferr was this the intent? It is actually dead code otherwise.

rust/tools/authorization-logic/src/signing/tink_interface.rs
aferr: I think I like the idea of putting this on all the functions that define interfaces for libraries most. I think doing it this way forces us to look at the public functions and think about if these really should be a part of the library API, or not.

> Or, do it via allow-dead on the BUILD library definition

I'm not sure I follow how this works. Is this a change to the BUILD file rather than the rust code? If so, does this actually stop the cargo warnings?

rust/tools/authorization-logic/src/souffle/souffle_interface.rs
aferr: What you have seems OK to me (listing it next to every function).

[This](https://stackoverflow.com/questions/25877285/how-to-disable-unused-code-warnings-in-rust) might be exactly the same stackoverflow you already looked at, but this one seems to suggest that you have to write `#![allow(dead_code)]` with the `!` for it to cover the whole crate. Since it says "crate-level" and not "file-level", maybe this means you have to put it in the `mod.rs` file rather than whichever file you are defining the function in. I did not try this myself, though!

rust/tools/authorization-logic/src/souffle/universe_translation.rs
aferr: I think we should just delete this one, then.

rust/tools/authorization-logic/src/test/test_queries.rs
aferr: Yes, that's the intent. I'm not sure why it wasn't like this already...

aferr (approved):

rust/tools/authorization-logic/src/souffle/souffle_interface.rs
Cypher1: I think the 'top of file' one needs a 'bang'

```
#![allow(dead_code)]
#![allow(unused_variables)]
```

Still I believe these individual ones are better as the 'unused' message persists in code, and the warning is silenced.

Cypher1 (approved):

rust/tools/authorization-logic/src/signing/tink_interface.rs
Cypher1: If the function is an interface, it should be `pub` at which point the dead code warning should go away. Rust is pretty good at detecting this kind of thing.

markww:
I think getting rid of warnings is laudable, but this might not be very high impact in the Rust code, as we are planning to delete it and move to C++ for this anyway.

shaper:
> I think getting rid of warnings is laudable, but this might not be very high impact in the Rust code, as we are planning to delete it and move to C++ for this anyway.

Understood, it's nice to keep the build clean in any case (there were some hidden warnings that pointed to real dead code that had possible interim ramifications), but I'll minimize time spent on Rust.

PR #609, opened 2022-07-07

Cypher1:
This draft is based on https://github.com/google-research/raksha/pull/598
I'm sending it out as a preview of future work we could do.

Needs:
- Tests for non-const IR traversal (i.e. traversal while performing mutations)
- Static checks for CopyConst and SelectType (there's tests but these are pretty limited).
- Comments

### DIFF
--- a/rust/tools/authorization-logic/BUILD
+++ b/rust/tools/authorization-logic/BUILD
@@ -16,7 +16,7 @@
 #-------------------------------------------------------------------------------
 load("@rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//:__subpackages__"])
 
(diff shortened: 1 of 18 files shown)


Cypher1:
@markww @aferr @bgogul FYI. Not sure we need this but thought I'd put it up for reference.

Cypher1:
Thanks @markww I think I have to close this to do that. Brb

markww:
J, you don't have to close the PR to change the branch that it merges into. If you press the "Edit" button on the PR, you should get a dropdown with all of the active branches and you can just select a different one. Sorry, I didn't intend for you to have to wrestle with GitHub this much.

Cypher1:
Thanks Mark,
Yeah, I thought that'd work but for me it wasn't showing up. It was no big
deal to fix.

J Pratt | (they/them) | Arcs Syd | ***@***.***

PR #610, opened 2022-07-07
Modify antlr grammar for g4 to allow numbers as strings

harsha-mandadi:
Ex: %3 = core.mult [lhs: 10, rhs: **"59"**](%0.out, %2.out)


src/parser/ir/Ir.g4
```diff
@@ -20,10 +20,10 @@ grammar Ir;
 // Parser
 //-----------------------------------------------------------------------------
 
-
+stringLiteral : NUMLITERAL | ID;
```
markww, line 23: Nit: Why the style difference between `stringLiteral` vs. `NUMLITERAL` or `ID`?

markww (approved):
LGTM other than a style quibble.

PR #611, opened 2022-07-07

Cypher1:
From #609
An addition to #598 to support mutating and non-mutating IRVisitors.


Cypher1:
Hit the wrong branch. One more try

PR #612, opened 2022-07-07

Cypher1:
Re-opening of #609 with the visitor change as a base for easier review.

Accidentally created #611 with the wrong base. Please disregard.

### DIFF
--- a/src/common/utils/types.h
+++ b/src/common/utils/types.h
@@ -16,32 +16,15 @@
 #ifndef SRC_COMMON_UTILS_TYPES_H_
 #define SRC_COMMON_UTILS_TYPES_H_
 
+#include <type_traits>
 #include <variant>
 
 namespace raksha {
 // Type alias for monostate that sounds less scary.
 using Unit = std::monostate;
 
-// An implementation of a static, type-level, 'if'.
-// Useful for writing typesafe, code generically (especially for generic constness).
-// Based on SAT's laser (specifically 'misc_utils.h').
-template <bool Flag, typename T, typename F>
-struct SelectType {
-    using Result = T;
-};
-
-template <typename T, typename F>
-struct SelectType<true, T, F> {
-    using Result = T;
-};
-
-template <typename T, typename F>
-struct SelectType<false, T, F> {
-    using Result = F;
-};
-
 template <bool IsConst, typename T>
-using CopyConst = typename SelectType<IsConst, const T, T>::Result;
+using CopyConst = typename std::conditional_t<IsConst, std::add_const_t<T>, T>;
 
 }
 
--- a/src/common/utils/types.h
+++ b/src/common/utils/types.h
@@ -19,8 +19,30 @@
 #include <variant>
 
 namespace raksha {
-  // Type alias for monostate that sounds less scary.
-  using Unit = std::monostate;
+// Type alias for monostate that sounds less scary.
+using Unit = std::monostate;
+
+// An implementation of a static, type-level, 'if'.
+// Useful for writing typesafe, code generically (especially for generic constness).
+// Based on SAT's laser (specifically 'misc_utils.h').
+template <bool Flag, typename T, typename F>
+struct SelectType {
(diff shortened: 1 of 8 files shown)


src/common/utils/types_test.cc
```diff
@@ -0,0 +1,49 @@
...
+}
+
+TEST(TypesTest, SelectTypeFalse) {
+  SelectType<false, int, std::string>::Result a = "Hi";
+  EXPECT_EQ(a, "Hi");
+}
+
+TEST(TypesTest, CopyConstTypeTrue) {
+  CopyConst<true, int> a = 3;
+  // a++; // Should not compile.
```
Cypher1, line 38: I'd love to find a way to statically assert this but it's not something I've worked out how to do.
Open to ideas.

markww: Here's how you'd do it:

```C++
ASSERT_EQ(typeid(&a), typeid(const int *));
ASSERT_NE(typeid(&a), typeid(int *));
```

See https://en.cppreference.com/w/cpp/language/typeid . I'm taking the address of `a` because `typeid` strips the top-level const (and volatile) qualifiers, but it leaves alone any qualifiers below the top level. But because of this quirk, I like to also check that the result is not equal to the non-const qualified thing, just to make sure I haven't messed up somehow.

Cypher1:
Added tests thanks to @markww. Great stuff :)

Cypher1:
Adding @bgogul in place of @markww.

src/common/utils/types.h
```diff
@@ -19,8 +19,30 @@
...
+    using Result = T;
+};
+
+template <typename T, typename F>
+struct SelectType<false, T, F> {
+    using Result = F;
+};
+
+template <bool IsConst, typename T>
+using CopyConst = typename SelectType<IsConst, const T, T>::Result;
```
markww, line 44: I just realized, I think `SelectType` is the same as `std::conditional` (https://en.cppreference.com/w/cpp/types/conditional). That could reduce the size of this PR.

Cypher1: Thanks Mark, will switch to using that and remove the associated new tests and types.

markww: Also, this could perhaps be more robust if `std::add_const` were used: https://en.cppreference.com/w/cpp/types/add_cv

Cypher1: Another good find :) Thanks Mark. Seems the std lib has nice helpers `std::conditional_t` and `std::add_const_t` that make things a little easier.

Done! :)

markww (approved):
LGTM, thanks for this :)

src/ir/ir_traversing_visitor.h
```diff
@@ -53,56 +53,60 @@ class IRTraversingVisitor : public IRVisitor<Derived, Result> {
...
   }
 
-  Result Visit(const Operation& operation) final override {
+  Result Visit(CopyConst<IsConst, Operation>& operation) final override {
     Result result = PreVisit(operation);
-    const Module* module = operation.impl_module();
+    auto* module = operation.impl_module();
     if (module != nullptr) {
-      result = FoldResult(std::move(result), module->Accept<Derived, Result>(*this));
+      result = FoldResult(result, module->Accept(*this));
```
markww, line 109: Why is `std::move` removed here? I believe that using `std::move` is correct: you are erasing the only name (`result`) pointing to the `PreVisit` result so that `Result` may be moved into `FoldResult` (and the result of that operation put back into `result`). If you're worried about the fact that the parameter may be const, I wouldn't: `std::move` does not activate the move constructor when used on a `const` name.

Cypher1: Thanks Mark! I thought I explicitly disabled the copy constructor for one of the test result values (and enabled the move constructor) but I hadn't.
I've now done that and gotten rid of any potential for accidental copying AND gotten rid of any unnecessary moves (the removing the ones there now cause build failure).

Thanks for the catch

PR #613, opened 2022-07-08
Tracking sensitivity analysis and tests

asejfia:
@bgogul @markww @aferr FYI, this is my code to track sensitivities so far.


build_defs/test_helpers.bzl
```diff
@@ -231,3 +231,70 @@ def output_rule_diff_test(
         data = [expected_contents_csv, ":resultFact"],
         srcs = ["//src/analysis/souffle/tests/arcs_fact_tests:diff_wrapper.sh"],
     )
+    
+def run_sensitivity_analysis_compare_results(
+        name,
+        test_bin,
+        input_files,
+        visibility = None):
```
markww, line 239: This looks largely like a copy of the version for taint analysis. What are the differences? Is it possible to refactor this so that they share most of their code?

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
...
+#define SENSITIVITY_ANALYSIS_DL_
+//this file defines the calculations for sensitivity analysis 
+//to be used in dp calculations
+
+#include "src/analysis/souffle/taint.dl"
+
+//one isSqlOperator fact per each sql operator, currently supporting one output
+//referring to results as 'value' (sqlValue declared below) as per the discussion in the doc 
+.decl isSqlOperator(sqlOperator: Operation)
+isSqlOperator([owner, operator, result, input_list, attribute_list]) :- isOperation([owner, operator, result, input_list, attribute_list]), match("sql.*", operator).
```
markww, line 11: We have [some handy relations](https://github.com/google-research/raksha/blob/main/src/analysis/souffle/operations.dl#L52) for getting fields of `Operation`s that should allow you to not fully deconstruct the `Operation` record type.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
+#ifndef SENSITIVITY_ANALYSIS_DL_
+#define SENSITIVITY_ANALYSIS_DL_
+//this file defines the calculations for sensitivity analysis 
+//to be used in dp calculations
+
+#include "src/analysis/souffle/taint.dl"
+
+//one isSqlOperator fact per each sql operator, currently supporting one output
+//referring to results as 'value' (sqlValue declared below) as per the discussion in the doc 
+.decl isSqlOperator(sqlOperator: Operation)
```
markww, line 10: Why does `isSqlOperator` contain an `Operation` rather than an `Operator`?

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
...
+isSqlOperator([owner, operator, result, input_list, attribute_list]) :- isOperation([owner, operator, result, input_list, attribute_list]), match("sql.*", operator). 
+
+
+.decl isSqlValue(value: AccessPath)
+isSqlValue(value) :- isSqlOperator([_, _, value, _, _]).
```
markww, line 15: It looks like you're conflating the concepts of `isSqlOperation` and `isSqlOperator`. I'd say make both relations, and make each of them take as their single argument the type that forms the suffix of the relation name.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
...
+//used in finding the maximm sensitvity 
+.decl sqlValueHasSqlInputListSuffix(value: AccessPath, partialList: OperandList)
+
+sqlValueHasSqlInputListSuffix(value, input_list) :-
+  sqlValueHasSqlInputList(value, input_list).
+sqlValueHasSqlInputListSuffix(value, tail) :-
+  sqlValueHasSqlInputListSuffix(value, [_, tail]).
+
+//calculate the length of each sql operator input list by iterating starting from the end
+.decl sqlOperatorInputListLength(input_list: OperandList, list_length: number)
```
markww, line 31: I think this is duplicated work. You can probably just use the relation already populated [here](https://github.com/google-research/raksha/blob/main/src/analysis/souffle/operations.dl#L89).

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
...
+
+//assign the length of each input to the corresponding sqlValue
+.decl sqlValueSqlOperatorInputListLength(value: AccessPath, input_list_length: number)
+
+sqlValueSqlOperatorInputListLength(value, len) :-
+  isSqlValue(value), sqlValueHasSqlInputList(value, input_list),
+  sqlOperatorInputListLength(input_list, len).
+
+//indexing the input per value 
+.decl sqlValueHasSqlInputAtIndex(value: AccessPath, sql_input: AccessPath, index:number)
```
markww, line 45: I think this is a duplicate of [this relation](https://github.com/google-research/raksha/blob/main/src/analysis/souffle/operations.dl#L97), you can just use that relation instead of rewriting it.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
...
+
+sqlValueHasSqlInputAtIndex(value, sql_input, sql_input_list_len - suffix_list_len) :-
+       sqlValueHasSqlInputListSuffix(value, [sql_input, tail]), 
+       sqlOperatorInputListLength([sql_input, tail], suffix_list_len), 
+       sqlValueSqlOperatorInputListLength(value, sql_input_list_len). 
+
+//keeping track of all initial inputs as they all have sennsitivity score 1
```
markww, line 52: Nit: sennsitivity -> sesnsitivity.

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
```diff
@@ -0,0 +1,16 @@
...
+//Only outputs value sensitivities
+
+.input isOperation(delimiter=";")
+.input isInitialInput(delimiter=";")
+.input expectedResult(delimiter=";")
+.output actualResult(delimiter=";")
+.output expectedResult(delimiter=";")
+
+
+#endif SENSITIVITY_ANALYSIS_UNIT_TEST_INTERFACE_DL_
```
markww, line 16: Add newline at EOF.

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
```diff
@@ -0,0 +1,16 @@
+#ifndef SENSITIVITY_ANALYSIS_UNIT_TEST_INTERFACE_DL_
+#define SENSITIVITY_ANALYSIS_UNIT_TEST_INTERFACE_DL_
+
+#include "src/analysis/souffle/sensitivity_analysis.dl"
+
+//An interface used for unit testing of the sensitivity analysis. 
+//Only outputs value sensitivities
+
+.input isOperation(delimiter=";")
+.input isInitialInput(delimiter=";")
```
markww, line 10: I'm intrigued by the decision to make `isInitialInput` an input relation. Wouldn't it be safer to just look for operations that grab columns out of tables for which there are no predecessors to discover the set of initial inputs?

Another thought: we basically only care about sensitivity analysis for columns that have user data or other sensitive information in them. So perhaps this should instead be something like `isSensitiveColumn`.

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
```diff
@@ -0,0 +1,16 @@
...
+.input expectedResult(delimiter=";")
+.output actualResult(delimiter=";")
+.output expectedResult(delimiter=";")
```
markww, line 13: Why is `expectedResult` both an input and an output relation?

src/analysis/souffle/tests/dp_sensitivity_tests/expectedResult.facts
```diff
@@ -0,0 +1,5 @@
+revenue_table_where1;1
+revenue_table_group_by;2
+revenue_table_where2;1
+user;2
+address;2
```
markww, line 5: End file with newline. If you're using vscode, it should be possible to configure it to always put a newline at the end of your files.

src/analysis/souffle/tests/dp_sensitivity_tests/isInitialInput.facts
```diff
@@ -0,0 +1 @@
+"revenue_table"
```
markww, line 1: End file with newline.

src/analysis/souffle/tests/dp_sensitivity_tests/isOperation.facts
```diff
@@ -0,0 +1,5 @@
+["template_owner", "sql.where", "revenue_table_where1", ["revenue_table", nil], nil]
+["template_owner", "sql.group_by", "revenue_table_group_by", ["revenue_table", nil], nil]
+["template_owner", "sql.where", "revenue_table_where2", ["revenue_table", nil], nil]
+["template_owner", "sql.select", "user", ["revenue_table_where1", ["revenue_table_group_by", ["revenue_table_where2", nil]]], nil]
+["template_owner", "sql.select", "address", ["revenue_table_where1", ["revenue_table_group_by", nil]], nil]
```
markww, line 5: End file with newline.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,132 @@
...
+      sensitivity_score >= input_sensitivity_score. 
+
+
+sqlValueHasSensitivityScoreAtIndex(value, sensitivity_score, index) :- 
+      isSqlValue(value),
+      sqlValueHasSqlInputAtIndex(value, sql_input, index), 
+      isSqlInput(value, sql_input), 
+      sqlValueSensitivity(sql_input, sensitivity_score), 
+      sqlValueHasSensitivityScoreAtIndex(value, prev_sensitivity_score, index-1), 
+      sensitivity_score >= prev_sensitivity_score.
```
markww, line 100: I'm pretty confused by the relationship between these two relations. Are they trying to do the same thing, or something slightly different?

asejfia: They're supposed to do slightly different things (see last line). The first one makes sure the sensitivity score is greater or equal then the sensitivity score of the input at  *index*; the second one makes sure the sensitivity score is greater or equal than the previous sensitivity score; combined, they should update the current score only if the sensitivity score of input at *index* is greater than the previous sensitivity score (*index-1*); otherwise, just copy it. My souffle knowledge is still not very well formed so please let me know if you think I should change anything or if the code is plain wrong. Alternatively, would adding a comment explaining what I just did help? 

Sidenote: Gogul mentioned I might be able to combine this in one line using an *or* operator. Would that help?

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
asejfia: It was an attempt to fit in with the existing bazel rule. I reused another one written by you and it doesn't have this problem anymore.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Yes. I reused that code in the more recent commits. Thank you!

src/analysis/souffle/sensitivity_analysis.dl
asejfia: the decision came after a discussion with @bgogul. AFAIK, all input to the datalog facts has to be formatted as an Operation. The name 'sql operator' was also discussed in the google doc and I switched to operator after @bgogul suggested it. If I misunderstood something, I'll be happy to revert/make more changes.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: I am not sure I understand what that would be useful for specifically.

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
asejfia: Is this a concern so that the user doesn't have to add any more annotations? This relation is supposed to distinguish data sources that have been given as input vs. intermediate or final ones produced by running queries on them, so not entirely sure isSensitiveColumn makes that difference clear.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
...
+//keep track at the score at each index, only update if the new input score is higher
+sqlValueHasSensitivityScoreAtIndex(value, sensitivity_score, index) :- 
+      isSqlValue(value),
+      isSqlOperator(operation), 
+      operation = [_, _, value, _, _],
+      operationHasOperandAtIndex(operation, sql_input, index),
+      isSqlInput(value, sql_input), 
+      sqlValueSensitivity(sql_input, input_sensitivity_score), 
+      sqlValueHasSensitivityScoreAtIndex(value, sensitivity_score, index-1),
+      sensitivity_score >= input_sensitivity_score.
```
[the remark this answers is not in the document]
asejfia: This relation is supposed to do something like what you described. It just finds the maximum of all input sensitivities (updates the ith score only if the ith input's score is higher than i-1; so the last index contains the maximum score). Do you think we should change the name for it?

The sqlValueSensitivity relation defines the final score and that's where you see the multiplication happen.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
...
+
+//base cases (index 0)
+//for initial inputs, sensitivity is 1
+sqlValueHasSensitivityScoreAtIndex(value, 1, 0) :- isSqlOperator(operation), 
+      operation = [_, _, value, _, _],
+      operationHasOperandAtIndex(operation, sql_input, 0),
+      isInitialInput(sql_input).
+
+//for all others, get the final sensitivity score
+sqlValueHasSensitivityScoreAtIndex(value, sensitivity_score, 0) :- isSqlOperator(operation),
```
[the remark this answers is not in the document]
asejfia: Yes, we had a problem with stratification and that's why we had to have the maximum decomposed in two different relations.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
+#ifndef SENSITIVITY_ANALYSIS_DL_
```
bgogul, line 1: Full path in the header guard: `SRC_ANALYSIS_SOUFFLE_SENSITIVITY_ANALYSIS_DL_`

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
...
+#define SENSITIVITY_ANALYSIS_DL_
+//this file defines the calculations for sensitivity analysis 
+//to be used in dp calculations
+
+#include "src/analysis/souffle/taint.dl"
+
+//one isSqlOperator fact per each sql operator, currently supporting one output
+//referring to results as 'value' (sqlValue declared below) as per the discussion in the doc
```
bgogul, line 9: nit: space after `//`
```suggestion
// one isSqlOperator fact per each sql operator, currently supporting one output
// referring to results as 'value' (sqlValue declared below) as per the discussion in the doc 
```

src/analysis/souffle/sensitivity_analysis.dl
bgogul: May be call it `isSqlOperation`?

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
...
+//one isSqlOperator fact per each sql operator, currently supporting one output
+//referring to results as 'value' (sqlValue declared below) as per the discussion in the doc 
+.decl isSqlOperator(sqlOperator: Operation)
+isSqlOperator([owner, operator, result, input_list, attribute_list]) :- isOperation([owner, operator, result, input_list, attribute_list]), match("sql.*", operator). 
+
+
+.decl isSqlValue(value: AccessPath)
+isSqlValue(value) :- isSqlOperator([_, _, value, _, _]).
+
+//keeping track of all initial inputs as they all have sensitivity score 1
```
bgogul, line 17: nit: space after `//`

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
...
+.decl isInitialInput (ingress_point: AccessPath)
```
bgogul, line 18: nit: space
```suggestion
.decl isInitialInput(ingress_point: AccessPath)
```

src/analysis/souffle/sensitivity_analysis.dl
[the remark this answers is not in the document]
bgogul: May be add a comment to this effect to avoid people reading the code from wondering the same thing as the discussion here.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,95 @@
...
+
+//flattening each individual input per value
+.decl isSqlInput(value: AccessPath, sql_input: AccessPath)
+
+isSqlInput(value, sql_input) :- isSqlOperator([_, _, value, input_list, _]), 
+                              flatten(input_list, sql_input). 
+
+//keeping track of sensitivity score per each input at index 
+//used to calculate the maximum input sensitivity 
+.decl sqlValueHasSensitivityScoreAtIndex(value: AccessPath, sensitivity_score: number, index: number)
```
bgogul, line 37: This is capturing max of sensitivity for an `operation` (not value) upto an index, right? Can we name it and change the signature accordingly? e.g., 

`sqlOperationHasSensitivityScoreUptoInputIndex(op: Operation, sensitivity_score: number, index: number)`

You can use any other name that you think is meaningful.  The key thing is that we should key on `Operation` instead of `AccessPath`?

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
```diff
@@ -0,0 +1,13 @@
+#ifndef SENSITIVITY_ANALYSIS_UNIT_TEST_INTERFACE_DL_
+#define SENSITIVITY_ANALYSIS_UNIT_TEST_INTERFACE_DL_
+
+#include "src/analysis/souffle/sensitivity_analysis.dl"
+
+//An interface used for unit testing of the sensitivity analysis.
```
bgogul, line 6: nit: space after `//`

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
```diff
@@ -0,0 +1,13 @@
+#ifndef SENSITIVITY_ANALYSIS_UNIT_TEST_INTERFACE_DL_
```
bgogul, line 1: full path in header guard macro.

bgogul (commented):
Left some suggestions for cleaning up.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: We're currently tracking sensitivities per value though (value being the output of an sqlOperation). I implemented this based on our discussions in the design doc. I am not sure I see an advantage to converting to Operation (as we would have to add checks for Operation's output everywhere). What do you think?

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Done.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Got it now. Updated!

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Changed.

src/analysis/souffle/sensitivity_analysis_unit_test_interface.dl
asejfia: Yes, I introduced this change in the other PR which depends on this.

src/analysis/souffle/sensitivity_analysis.dl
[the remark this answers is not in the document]
aferr: OK sounds good. 

It might be useful to add a comment that explains that index means depth in the IR graph.

aferr (approved):
I had a minor suggestion about adding a comment, but I'll approve because of the time zone difference. (The comment is not too important anyway)

PR #616, opened 2022-07-12
Fix `ir_printer.h` include that breaks external builds.

shaper:
I see this at ToT here: https://github.com/google-research/raksha/blame/main/src/ir/block_builder_test.cc#L23

But it looks correct in https://github.com/google-research/raksha/pull/590/files#diff-4331b53480b9ef191adbc3c4fb8b7726a61ec9a4cea83663f0036c91e643f3a1 so I'm not sure what happened.

### DIFF
--- a/src/ir/block_builder_test.cc
+++ b/src/ir/block_builder_test.cc
@@ -20,7 +20,7 @@
 #include "src/common/testing/gtest.h"
 #include "src/ir/attributes/attribute.h"
 #include "src/ir/ir_context.h"
-#include  "third_party/raksha/src/ir/ir_printer.h"
+#include "src/ir/ir_printer.h"
 #include "src/ir/ssa_names.h"
 #include "src/ir/types/entity_type.h"
 #include "src/ir/types/type.h"


harsha-mandadi (approved):

src/ir/block_builder_test.cc
```diff
@@ -20,7 +20,7 @@
 #include "src/common/testing/gtest.h"
 #include "src/ir/attributes/attribute.h"
 #include "src/ir/ir_context.h"
-#include  "third_party/raksha/src/ir/ir_printer.h"
+#include "src/ir/ir_printer.h"
```
cromwellian, line 23: Does the internal build still work after this change? (e.g. this fixes external, but it still works internally?)

cromwellian (commented):
Does the internal build still work after this change? (e.g. this fixes external, but it still works internally?)

src/ir/block_builder_test.cc
shaper: I think it will, yes -- internal code is transformed by copybara to add the prefix. The question is how the prefix made it back out, it should have been stripped (again, if needed). I theorize the extra space may have broken the regex. Am talking about it in Raksha chat.

bgogul:
> But it looks correct in https://github.com/google-research/raksha/pull/590/files#diff-4331b53480b9ef191adbc3c4fb8b7726a61ec9a4cea83663f0036c91e643f3a1 so I'm not sure what happened.

We should check the copybara config to see why the transformation of include directory did not happen.

bgogul:
> > But it looks correct in https://github.com/google-research/raksha/pull/590/files#diff-4331b53480b9ef191adbc3c4fb8b7726a61ec9a4cea83663f0036c91e643f3a1 so I'm not sure what happened.
> 
> We should check the copybara config to see why the transformation of include directory did not happen.

I think I know why the transformation did not happen. In this case, there was an extra space in  `#include  "third_party/raksha/...`. 

Surprisingly, the test failures did not block the merge from happening. We should look into it.

bgogul:
> Surprisingly, the test failures did not block the merge from happening. We should look into it.

I think I know why as well. The PR that @harsha-mandadi  cloned internally had the tag `GIT_ORIGIN_REV_ID`, which disabled tests. (Works as intended.)

harsha-mandadi:
Interesting!! I did not intentionally set any tags while cloning. Does it happen automatically and how can we avoid this in the future??

bgogul:
> Interesting!! I did not intentionally set any tags while cloning. Does it happen automatically? How can we avoid this in the future??

We had an internal discussion. Replying here also to close this thread.  Yes, the copybara service adds this tag automatically whenever a PR is imported. We disable external tests when `GIT_ORIGIN_REV_ID` is present. The idea is that before importing a PR, the external tests would have run away and should need to be run again.

PR #621, opened 2022-07-15
Initiating end-to-end dp analysis

asejfia:
Data and code to test end-to-end integration for sensitivity analysis. Had to do some changes to the sensitivity analysis from the other branch too to address some of the new operations.


src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -112,13 +95,29 @@ sqlValueSensitivity(value, sensitivity_score) :-
...
+
+.decl isDiffPrivacyMechanism(value: AccessPath, epsilon: number)
+isDiffPrivacyMechanism(value, epsilon) :- isOperation([_, "privacy_mechanism", value, _, [["epsilon", $NumberAttributePayload(epsilon)], _]]). 
+
+//sequential composition
+.decl epsilonDiffPrivacy(value: AccessPath, epsilon: number)
+epsilonDiffPrivacy(value, epsilon) :- isDiffPrivacyMechanism(value, _),
+      epsilon = sum y : { isDiffPrivacyMechanism(value, y)}. 
+
+violatesPolicy(value) :- isSqlOperator([_, _, value, _, _]),
```
bgogul, line 110: We should probably only check for value that satisfy `isSqlOutput`?

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -112,13 +95,29 @@ sqlValueSensitivity(value, sensitivity_score) :-
...
+//sequential composition
+.decl epsilonDiffPrivacy(value: AccessPath, epsilon: number)
+epsilonDiffPrivacy(value, epsilon) :- isDiffPrivacyMechanism(value, _),
+      epsilon = sum y : { isDiffPrivacyMechanism(value, y)}. 
+
+violatesPolicy(value) :- isSqlOperator([_, _, value, _, _]), 
+      !sqlValueSensitivity(value, _),
+      epsilonDiffPrivacy(value, number), 
+      isGlobalEpsilon(global_epsilon),
+      number > global_epsilon;
```
bgogul, line 114: keep style consistent with else case: `global_epsilon <= number`

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -112,13 +95,29 @@ sqlValueSensitivity(value, sensitivity_score) :-
...
+.decl isDiffPrivacyMechanism(value: AccessPath, epsilon: number)
+isDiffPrivacyMechanism(value, epsilon) :- isOperation([_, "privacy_mechanism", value, _, [["epsilon", $NumberAttributePayload(epsilon)], _]]). 
+
+//sequential composition
+.decl epsilonDiffPrivacy(value: AccessPath, epsilon: number)
+epsilonDiffPrivacy(value, epsilon) :- isDiffPrivacyMechanism(value, _),
+      epsilon = sum y : { isDiffPrivacyMechanism(value, y)}. 
+
+violatesPolicy(value) :- isSqlOperator([_, _, value, _, _]), 
+      !sqlValueSensitivity(value, _),
```
bgogul, line 111: What do the disjunctions in this disjunction correspond to?

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -112,13 +95,29 @@ sqlValueSensitivity(value, sensitivity_score) :-
       sqlValueHasSensitivityScoreAtIndex(value, input_sensitivity, input_list_length-1), 
       sensitivity_score = input_sensitivity * function_sensitvity.
 
+.decl isGlobalEpsilon(epsilon: number)
+//TODO: is there a way to set a deafault option?
```
bgogul, line 99: We could probably add this as an input fact.

bgogul (commented):
Left some clarification questions.

src/backends/policy_engine/souffle/dp_testdata/dp_analysis.ir
```diff
@@ -0,0 +1,13 @@
+module m0 {
+  block b0 {
+        %0 = core.input [name: "revenue_table"] ()  
+        %epsilon = global_epsilon [epsilon: "2"] ()
+        %1 = sql.filter [column: "revenue"] (%0.out) // { revenue > 100 }
```
harsha-mandadi, line 5: To avoid tests from failing because of "Value not found" error, please change %0.out to %0. Similarly in all other places below.

asejfia:
btw: had to rebase from main because of changes to IR that were newer, that's why there are so many commits here.

src/backends/policy_engine/souffle/dp_testdata/dp_analysis.ir
asejfia: done.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: 1. There is a privacy mechanism applied to value, but the value does not need sensitivity analysis
2. There is a privacy mechanism applied to value and the value has a particular sensitivity score 
3. (Just realized this one is incomplete, will fix on the next commit) There is no privacy mechanism on a value that has a particular sensitivity score.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: We currently do not have such a relation/fact. What would that model (that is not currently modeled by the output of an Operation)?

asejfia: I just saw the relation defined in another file. I think with the new isAggregationOperation relation this would no longer be  necessary, right?

asejfia:
@markww @aferr can you take a look at these changes whenever you have time? TY!

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+#define SQL_COUNT_OPERATOR as("sql.count", Operator)
+#define SQL_MAX_OPERATOR as("sql.max", Operator)
+#define SQL_AVERAGE_OPERATOR as("sql.average", Operator)
+
+// One isSqlOperator fact per each sql operator, currently supporting one output.
+// We refer to results as 'value' (sqlValue declared below)
+.decl isSqlOperation(sqlOperation: Operation)
+isSqlOperation(operation) :- isOperation(operation), 
+      operationHasOperator(operation, operator), 
+      match("sql.*", operator).
```
markww, line 26: For discovering a prefix, `match` is probably more power than necessary and potentially slow. I'd say use `substr` instead.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+      ).
+// Aggregation operations point to where noise needs to be added. 
+.decl isAggregationOperation(sql_operation: Operation)
+isAggregationOperation(operation) :- isSqlOperation(operation), 
+      (
+            operationHasOperator(operation, SQL_SUM_OPERATOR); 
+            operationHasOperator(operation, SQL_COUNT_OPERATOR); 
+            operationHasOperator(operation, SQL_AVERAGE_OPERATOR);
+            operationHasOperator(operation, SQL_MAX_OPERATOR)
+      ).
```
markww, line 45: Rather than listing out the operators with `;`s between them like this, it might be useful to populate a separate relation with them. Imagine something like this:

```
.decl isTransformationOperator(op: Operator)
isTransformationOperator(SQL_WHERE_OPERATOR).
isTransformationOperator(SQL_GROUP_BY_OPERATOR).
isTransformationOperator(SQL_SELECT_OPERATOR).
```

This eliminates the need for the `;` separators in `isTransformationOperation`, which can now be written like so:

```
isTransformationOperation(operation) :- isSqlOperation(operation),
operationHasOperator(op), isTransformationOperator(op).
```

This also factors things nicely such that `isTransformationOperator` may be used in other relations if so desired (or even populated as an input relation if we want to allow user extensions in the future).

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+// TODO: add other cases 
+.decl sqlFunctionSensitivity(operator: Operator, sensitivity_score: number)
+
+sqlFunctionSensitivity(operator, 1) :- operator = SQL_FILTER_OPERATOR;
+      operator = SQL_WHERE_OPERATOR; 
+      operator = SQL_SELECT_OPERATOR;
+      operator = SQL_COLUMN_ACCESS_OPERATOR;
+      operator = SQL_OUTPUT_OPERATOR; 
+      isAggregationOperation(operation),
+      operationHasOperator(operation, operator).
```
markww, line 118: I'd express the constant cases as base facts like so:

```
sqlFunctionSensitivity(SQL_FILTER_OPERATOR, 1).
sqlFunctionSensitivity(SQL_WHERE_OPERATOR, 1).
...
```

I know this is more verbose, but using base facts (rules without a `:-`) when possible reduces the complexity of the datalog in a number of important ways (such as being trivially stratifiable) which makes them worth the additional characters.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+      operationHasOperandAtIndex(operation, sql_input, index),
+      isSqlInput(operation, sql_input), 
+      ( 
+            sqlValueSensitivity(sql_input, input_sensitivity_score), 
+            sqlOperationHasSensitivityScoreAtIndex(operation, sensitivity_score, index-1), 
+            sensitivity_score >= input_sensitivity_score; 
+            sqlValueSensitivity(sql_input, sensitivity_score), 
+            sqlOperationHasSensitivityScoreAtIndex(operation, prev_sensitivity_score, index-1), 
+            sensitivity_score >= prev_sensitivity_score
+      ).
```
markww, line 106: I'm having trouble understanding the logic here. Can this be refactored to simplify it or commented to explain it?

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+
+// Inductive cases.  
+// We keep track at the score at each index, and only update if the new input score is higher. 
+sqlOperationHasSensitivityScoreAtIndex(operation, sensitivity_score, index) :- 
+      isSqlOperation(operation), 
+      operationHasOperandAtIndex(operation, sql_input, index),
+      isSqlInput(operation, sql_input), 
+      ( 
+            sqlValueSensitivity(sql_input, input_sensitivity_score), 
+            sqlOperationHasSensitivityScoreAtIndex(operation, sensitivity_score, index-1),
```
markww, line 101: In general, try to express your inductive cases using `+1` rather than `-1`. So that would make this rule look more like:

```
sqlOperationHasSensitivityScoreAtIndex(operation, sensitivity_score, prev_index + 1) :-
...
  (
    sqlValueSensitivity(sql_input, input_sensitivity_score),
    sqlOperationHasSensitivityScoreAndIndex(operation, sensitivyt_score, prev_index)
```

It's just more in line with the classical inductive structure of "Prove for 0, then prove that true for N -> true for N + 1".

src/analysis/souffle/policy_verifier_interface.dl
```diff
@@ -20,6 +20,7 @@
 #include "src/analysis/souffle/sql_output.dl"
 #include "src/analysis/souffle/taint.dl"
 #include "src/analysis/souffle/tag_transforms.dl"
+#include "src/analysis/souffle/epsilon_analysis.dl"
```
markww, line 23: Is this include necessary here at this time?

src/analysis/souffle/tests/dp_sensitivity_tests/isOperation.facts
```diff
@@ -0,0 +1,5 @@
+["template_owner", "sql.where", "revenue_table_where1", ["revenue_table", nil], nil]
```
markww, line 1: One question that I have that I don't know is answered by the current tests: does this approach do the right thing if an input is calculated with a value without a sensitivity score, such as a literal?

src/backends/policy_engine/souffle/dp_testdata/dp_analysis_1q_failing.ir
```diff
@@ -0,0 +1,14 @@
+//modeling a single query that fails the dp analysis
```
markww, line 1: In these comments at the top of the files, could you expand a bit more on why these should pass or fail the dp analysis? It is a bit hard to follow just reading the IR language.

markww (commented):
The approach looks good. Have some questions and comments, largely to make it more readable.

src/analysis/souffle/epsilon_analysis.dl
```diff
@@ -0,0 +1,77 @@
...
+
+#define GLOBAL_EPSILON_OPERATOR as("global_epsilon", Operator)
+#define PRIVACY_MECHANISM_OPERATOR as("privacy_mechanism", Operator)
+
+// Global epsilon is a value determined by users of Raksha. 
+// We could eventually set one as the default. 
+.decl isGlobalEpsilon(epsilon: number)
+isGlobalEpsilon(epsilon) :- isOperation(operation), 
+    operationHasOperator(operation, GLOBAL_EPSILON_OPERATOR),
+    operationHasAttribute(operation, ["epsilon", $NumberAttributePayload(epsilon)]).
```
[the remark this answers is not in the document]
asejfia: Do you mean have .input isGlobalEpsilon()? I went this route cause I thought isOperation is the only way to translate IR into datalog facts automatically.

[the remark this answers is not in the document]
markww: Hm, you raise a good point. It would probably be labor intensive to follow my suggestion. Also, it seems like the global epsilon is a temporary situation, so I think this is OK for now.

src/analysis/souffle/policy_verifier_interface.dl
asejfia: We are checking for violation of policy, so yes.

src/analysis/souffle/epsilon_analysis.dl
```diff
@@ -0,0 +1,77 @@
...
+
+isSqlPath(sql_input, sql_output):- isSqlOperation(operation), 
+    isSqlInput(operation, sql_input), 
+    operationHasResult(operation, intermediate_output), 
+    isSqlPath(intermediate_output, sql_output);
+    isDiffPrivacyMechanism(operation), 
+    operationHasOperandList(operation, input_list), 
+    flatten(input_list, sql_input), 
+    isSqlPath(sql_input, intermediate_output),
+    isSqlPath(intermediate_output, sql_output).
```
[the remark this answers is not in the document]
asejfia: Is this missing the recursive part of the definition or am I failing to see it? Also, why would writing the relation like this simplify things? I feel like it would add unnecessary code.

src/analysis/souffle/epsilon_analysis.dl
```diff
@@ -0,0 +1,77 @@
...
+// 2. The local epsilon for an input to an aggregation operation is greater than the global epsilon. 
+violatesPolicy(value) :- isAggregationOperation(operation), 
+    operationHasResult(operation, sql_output), 
+    isSqlPath(value, sql_output), 
+    epsilonDiffPrivacy(value, epsilon), 
+    isGlobalEpsilon(global_epsilon),
+    (
+        sqlValueSensitivity(sql_output, sensitivity),
+        global_epsilon < epsilon * sensitivity; 
+        global_epsilon < epsilon
```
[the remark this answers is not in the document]
asejfia: The breakdown here serves to handle cases when there is no sensitivity analysis (also answers your question below about cases when there is no sensitivity analysis).

src/analysis/souffle/tests/dp_sensitivity_tests/isOperation.facts
asejfia: Answered above, we do check for cases like these too.

src/backends/policy_engine/souffle/dp_testdata/dp_analysis_1q_failing.ir
asejfia: Done. LMK if you think I should add more detail.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Done.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Added a comment, please let me know what you think.

src/analysis/souffle/epsilon_analysis.dl
```diff
@@ -0,0 +1,77 @@
...
+    operationHasOperator(operation, PRIVACY_MECHANISM_OPERATOR). 
+
+// The epsilon of each privacy mechanism is tracked and later attached to values. 
+.decl diffPrivacyMechanismHasEpsilon(operation: Operation, epsilon: number)
+diffPrivacyMechanismHasEpsilon(operation, epsilon) :- isDiffPrivacyMechanism(operation),
+    operationHasAttribute(operation, ["epsilon", $NumberAttributePayload(epsilon)]).
+
+// We propagate privacy mechanisms backwards to inputs as per the PINQ paper.
+// This relation tracks every input to its (direct and indirect) outputs.   
+.decl isSqlPath(sql_input: AccessPath, sql_output: AccessPath)
```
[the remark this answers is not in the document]
asejfia: Done.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+            sqlOperationHasSensitivityScoreAtIndex(operation, sensitivity_score, index-1), 
+            sensitivity_score >= input_sensitivity_score; 
+            sqlValueSensitivity(sql_input, sensitivity_score), 
+            sqlOperationHasSensitivityScoreAtIndex(operation, prev_sensitivity_score, index-1), 
+            sensitivity_score >= prev_sensitivity_score
+      ). 
+
+// Function sensitivity scores per each sql operator as defined in PINQ
+// TODO: add other cases 
+.decl sqlFunctionSensitivity(operator: Operator, sensitivity_score: number)
```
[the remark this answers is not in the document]
asejfia: Done.

src/analysis/souffle/epsilon_analysis.dl
```diff
@@ -0,0 +1,77 @@
...
+
+isSqlPath(sql_input, sql_output):- isSqlOperation(operation), 
+    isSqlInput(operation, sql_input),
+    operationHasResult(operation, sql_output); 
+    isDiffPrivacyMechanism(operation), 
+    operationHasOperandList(operation, input_list), 
+    flatten(input_list, sql_input),
+    operationHasResult(operation, sql_output).
```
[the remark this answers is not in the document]
asejfia: Great point. Done.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,134 @@
...
+// Helper functions to flatten input/operand lists. 
+.decl intermediateFlatten(list:OperandList, sublist:OperandList, list_element:AccessPath)
+
+intermediateFlatten(list, sublist, list_element) :- isOperation(operation), 
+      operationHasOperandList(operation, list), 
+      list = [list_element, sublist].
+intermediateFlatten(list, sublist, list_element) :- intermediateFlatten(list, [list_element, sublist], _).
+
+.decl flatten(list:OperandList, list_element:AccessPath)
+flatten(list, list_element) :- intermediateFlatten(list, _, list_element).
```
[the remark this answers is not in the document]
asejfia: Done.

src/analysis/souffle/epsilon_analysis.dl
```diff
@@ -0,0 +1,77 @@
...
+// We track every mechanism that adds noise. 
+// Right now, we are not tracking the type of noise (e.g. via Laplace of Gaussian mechanisms).
+.decl isDiffPrivacyMechanism(operation: Operation)
+isDiffPrivacyMechanism(operation) :- isOperation(operation), 
+    operationHasOperator(operation, PRIVACY_MECHANISM_OPERATOR). 
+
+// The epsilon of each privacy mechanism is tracked and later attached to values. 
+.decl diffPrivacyMechanismHasEpsilon(operation: Operation, epsilon: number)
+diffPrivacyMechanismHasEpsilon(operation, epsilon) :- isDiffPrivacyMechanism(operation),
+    operationHasAttribute(operation, ["epsilon", $NumberAttributePayload(epsilon)]).
```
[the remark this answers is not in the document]
asejfia: Done.

src/backends/policy_engine/souffle/dp_testdata/dp_analysis_2q_passing.ir
```diff
@@ -0,0 +1,19 @@
...
+        %dp_data = privacy_mechanism [epsilon: 1] (%2) //epsilon set to take into account sens. analysis 
+        %3 = sql.average [] (%dp_data)
+        %4 = sql.output [] (%3) // Report the noisy average
+        //second query
+        %5 = sql.column_access [colum: "profit"] (%0)
+        %6 = privacy_mechanism [epsilon: 1] (%5)
+        %7 = sql.average [] (%6)
+        %8 = sql.output [] (%7)
+    }
+}
```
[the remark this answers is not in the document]
asejfia: Ahh, I though VS Code would handle that for me automatically. Done.

src/analysis/souffle/epsilon_analysis.dl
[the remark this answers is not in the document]
markww: I see, the LHS of the `;` will be taken if there is a `sqlValueSensitivity` associated with the `sql_output`, the RHS otherwise. That makes sense.

In that case, this is fine. It might improve readability to wrap the LHS of the `;` in parens, but it might be too cluttered, I'd say try it out and see how it looks.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -0,0 +1,142 @@
...
+      operationHasOperandAtIndex(operation, sql_input, prev_index+1),
+      isSqlInput(operation, sql_input), 
+      ( 
+            sqlValueSensitivity(sql_input, input_sensitivity_score), 
+            sqlOperationHasSensitivityScoreAtIndex(operation, sensitivity_score, prev_index), 
+            sensitivity_score >= input_sensitivity_score; // ensure the prev_index sensitivity_score is higher than that of the input 
+            sqlValueSensitivity(sql_input, sensitivity_score), 
+            sqlOperationHasSensitivityScoreAtIndex(operation, prev_sensitivity_score, prev_index), 
+            sensitivity_score >= prev_sensitivity_score // ensure the input sensitvity_score is higher than that of prev_index 
+      ).
```
markww, line 114: Is this equivalent to the following?

```
sqlOperationHasSensitivityScoreAtIndex(operation, max(prev_sensitivity_score, input_sensitivity_score), prev_index + 1) :-
  isSqlOperation(operation),
  operationHasOperandAtIndex(operation, sql_input, prev_index + 1),
  isSqlInput(operation, sql_input),
  sqlValueSensitivity(sql_input, input_sensitivity_score),
  sqlOperationHasSensitivityScoreAtIndex(operation, prev_sensitivity_score, prev_index).
```

If so, that might be a bit easier to follow.

asejfia: Yes! Wow, this is much better. I didn't know you could use max in the head of a rule, but now that I think about it, it makes a lot of sense. Thank you, Mark. Pushing another commit now that changes the code.

src/analysis/souffle/epsilon_analysis.dl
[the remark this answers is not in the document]
markww: No, you're not failing to see it, I wrote it wrong. Trying again:

```
isSqlPath(sql_input, sql_input) :- isSqlInput(operation, sql_input).
isSqlPath(diff_priv_input, diff_priv_input) :- isDiffPrivacyMechanism(operation), 
    operationHasOperandList(operation, input_list), 
    flatten(input_list, sql_input).
    
isSqlPath(sql_input, sql_output) :-
  isSqlPath(sql_input, intermediate),
  operationHasOperandList(operation, input_list), flatten(input_list, intermediate),
  operationHasResult(operation, sql_output).
```

Here's what I'm trying to do: in the original code, I see you handling two cases, a case involving a sql operation and a case involving a differential privacy mechanism. It looks to me like the rules generated are entangling two separate concerns:

1. Starting a SQL path using a differential privacy mechanism and/or SQL operation.
2. Traversing a SQL path across an operation.

I'm trying to separate those concerns by just introducing a zero-length path for the initial case and then uniformly handling the paths afterwards. I may be misunderstanding your propagation code, though, so if this does not work, do feel free to keep what you have.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Done.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Done.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Done.

markww (commented):
LGTM. I have a few comments outstanding, but nothing showstopping, feel free to get this in once you address the comments as you see fit (anywhere between ignore and fix).

markww (approved):

src/analysis/souffle/epsilon_analysis.dl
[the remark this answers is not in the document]
asejfia: Yes, that makes a lot of sense. I changed the code.

Cypher1:
Did a quick review (for security and build breaking) internally, copying comment here for overcommunication:

Lots of trailing white-space (please adjust your editor to remove this as it can make tracking changes a little harder), otherwise LGTM

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -1,90 +1,134 @@
...
 isSqlInput(operation, sql_input) :- isSqlOperation(operation), 
       operationHasOperandList(operation, input_list), 
       flatten(input_list, sql_input). 
 
-// keeping track of sensitivity score per each input at index 
-// used to calculate the maximum input sensitivity 
-// decomposed into base and inductive cases because recursive aggregates are not supported yet 
-// for the compiler case in souffle
-.decl sqlOperationHasSensitivityScoreAtIndex(operation: Operation, sensitivity_score: number, index: number) 
+// Keeping track of sensitivity score per each input at index.
```
aferr, line 83: I think it could be useful to explain in the comment that `index` is the depth in the graph from the initial inputs (where the initial inputs have index 0).

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -1,90 +1,134 @@
...
-// keeping track of sensitivity score per each input at index 
-// used to calculate the maximum input sensitivity 
-// decomposed into base and inductive cases because recursive aggregates are not supported yet 
-// for the compiler case in souffle
-.decl sqlOperationHasSensitivityScoreAtIndex(operation: Operation, sensitivity_score: number, index: number) 
+// Keeping track of sensitivity score per each input at index. 
+// This is used to calculate the maximum input sensitivity. 
+// It's decomposed into base and inductive cases because recursive 
+// aggregates are not supported yet for the compiler case in souffle
+.decl sqlOperationHasSensitivityScoreAtIndex(operation: Operation, sensitivity_score: SensitvityScore, index: number)
```
aferr, line 87: (This comment is also based on the definition of the relation below).

I think you need some relation that gives the sensitivity of **operand**s  (AccessPaths) rather than operations, so it should be more like: `.decl  pathHasSensitivity(operand: AccessPath, sensitivity: SensitivityScore)`

and the sensitivity of the path should depend on the _operation_ that defines this path. It should be the function sensitivity of the operation in the definition multiplied by the maximum over the sensitivities of the operations in the inputs. You might need an auxiliary definition to compute the maximum over the list of inputs. So the following should almost work, but it might not be stratifiable:
```
.decl maxSensitivityInOperandList(operand_list: OperandList, sensitivity_score: SensitivityScore)
maxSensitivityInOperandList(nil, 0).
maxSensitivityInOperandList([path, subList], max(pathSensitivity, listSensitivity)) :- 
        isOperandList(subList),
        pathHasSensitivity(path, pathSensitivity),
        maxSensitivityInOperandList(subList, listSensitivity).

pathHasSensitivity(result_path, sensitivity) :- 
        operationHasResult(result_path, operation),
        operationHasOperandList(operation, input_list), 
        operationHasOperator(operation, operator),
        sqlFunctionSensitivity(operator, function_sensitivity),
        maxSensitivityInOperandList(input_list, max_input_sensitivity),
        sensitivity = function_sensitivity * max_input_sensitivity
```

To make this stratifiable we might need to add the index and do induction on indexes as you are doing, or maybe another way is to merge the functionality of `pathSensitivity` and `maxSensitivityInOperandList` somehow...

aferr: OK my mistake, when reading this code, the drop-down arrow was hiding the rest of the definition of `sqlValueSensitivity`, which does look right to me! I should have noticed the arrow was hiding it...

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -1,90 +1,134 @@
...
+sqlFunctionSensitivity(SQL_OUTPUT_OPERATOR, 1). 
+sqlFunctionSensitivity(SQL_GROUP_BY_OPERATOR, 2).
+
+// The final value sensitivity score, equal to the max of input 
+// multiplied by the function sensitivity.
+.decl sqlValueSensitivity(value: AccessPath, sensitivity_score: SensitvityScore)
 
-// final value sensitivity score, the max of input multiplied by the function sensitivity 
-.decl sqlValueSensitivity(value: AccessPath, sensitivity_score: number)
 sqlValueSensitivity(value, sensitivity_score) :-
```
aferr, line 132: As a minor point, I think I would find this easier to read if this relation definition comes *before* the definition of `sqlOperationHasSensitivityScoreAtIndex` since `sqlValueSensitivity` is the main one that is consumed by the epsilon part of the analysis.

src/analysis/souffle/sensitivity_analysis.dl
aferr: Maybe it's also useful to explain that the sensitivity of the operation is used as a proxy for the sensitivity of the access path defined by this operation.

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Fixed!

src/analysis/souffle/sensitivity_analysis.dl
asejfia: Done.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -1,90 +1,134 @@
...
 
-// helper functions to flatten input lists
+// Helper functions to flatten input/operand lists. 
 .decl intermediateFlatten(list:OperandList, sublist:OperandList, list_element:AccessPath)
 
-intermediateFlatten(list, sublist, list_element) :- isSqlOperation(operation), 
-      operationHasOperandList(operation, list), 
-      list = [list_element, sublist].
-intermediateFlatten(list, sublist, list_element) :- intermediateFlatten(list, [list_element, sublist], _).
+intermediateFlatten(list, sublist, list_element) :- isOperation(operation),
```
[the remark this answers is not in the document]
asejfia: You're right! Thank you.

src/analysis/souffle/sensitivity_analysis.dl
```diff
@@ -1,90 +1,134 @@
...
 
 .decl isSqlValue(value: AccessPath)
 isSqlValue(value) :- isSqlOperation(operation),
       operationHasResult(operation, value).
 
-// keeping track of all initial inputs as they all have sensitivity score 1
-.decl isInitialInput(ingress_point: AccessPath)
+// Keeping track of all initial inputs. 
+.decl isInitialInput (ingress_point: AccessPath)
+isInitialInput(as(ingress_point, AccessPath)) :- isOperation(operation),
```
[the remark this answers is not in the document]
asejfia: It's a symbol and souffle does not convert it automatically.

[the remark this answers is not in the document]
aferr: ah ok thanks

aferr (approved):

Issue #622, opened 2022-07-26
Explore SMT solver as an optional tool in the Raksha toolbox.


shaper:
Towards long term research purposes. Currently considering Z3, CVC4, CVC5.

aferr:
A nice thing about these tools is that they all accept [SMT-LIB](https://smtlib.cs.uiowa.edu/) code, so the decision here can be easily pivoted, or our code can actually make use of all of these solvers.

shaper:
I have CVC5 building externally. I'll put up a pending PR soon to discuss with @markww. There are some choices around library dependencies that affect licensing and a few other small things we need to consider.

PR #624, opened 2022-07-29
Update Bazel rules_foreign_cc from 0.4.0 to 0.8.0.

shaper:
As part of building CVC5 for #622 I found that our `rules_foreign_cc` library seems quite dated. The new version allows specifying `copts` in the `cmake` rule, and there are surely other updates of potential use as well.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -41,9 +41,9 @@ rules_proto_grpc_cpp_repos()
 #--------------------
 http_archive(
     name = "rules_foreign_cc",
-    sha256 = "e14a159c452a68a97a7c59fa458033cc91edb8224516295b047a95555140af5f",
-    strip_prefix = "rules_foreign_cc-0.4.0",
-    url = "https://github.com/bazelbuild/rules_foreign_cc/archive/0.4.0.tar.gz",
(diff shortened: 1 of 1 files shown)


markww (approved):
LGTM, but you need to run `buildifier` to fix internal checks.

harsha-mandadi:
This version change has caused my build to fail on Mac with following error.
```
./lib/findprog-in.c:137:25: error: implicit declaration of function 'eaccess' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                    if (eaccess (progpathname, X_OK) == 0)
                        ^
./lib/findprog-in.c:137:25: note: did you mean 'access'?
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:431:6: note: 'access' declared here
int      access(const char *, int);
         ^
./lib/findprog-in.c:211:21: error: implicit declaration of function 'eaccess' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                if (eaccess (progpathname, X_OK) == 0)
```
Looks like error is related to `eaccess` being called before it is declared. It fails on something not related to raksha. Have you encountered this or is this only for mac users?

markww:
This looks like a warning that has been upgraded to an error. We should probably just disable `-Werror` for third party code; I'm surprised it's on in this build in the first place.

PR #627, opened 2022-08-02
Refactor parse lex

aferr:
This exists only for illustrative purposes, do not review.


rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -25,20 +25,26 @@ grammar AuthLogic;
 //-----------------------------------------------------------------------------
 // Parser
 //-----------------------------------------------------------------------------
+id 
+    : VARIABLE #variable
+    | '"' CONSTANT_BODY '"' #constant
+    | '"""' MULTILINE_CONSTANT_BODY '"""'  #multilineConstant
```
Cypher1, line 31: Not reviewing this as a code review, but out of interest:

When is it useful to have muilti-line identifier names? I can only imagine this causing confusion.

Alternatively, allowing multi-line string literals seems perfectly reasonable (though I see string literal is missing from `pred_arg` and possibly the file)

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
```diff
@@ -139,7 +145,9 @@ ATTRIBUTE: 'attribute';
 
 // Identifiers wrapped in quotes are constants whereas
 // identifiers without quotes are variables.
-ID : ('"')? [_a-zA-Z][_a-zA-Z0-9/.#:]* ('"')?;
```
Cypher1, line 142: Oh, the old version currently allows unmatched `"` quotes in IDs. Is this desired, or should we use something like:
```
ID: RAW_ID | ( '"' RAW_ID '"' );
RAW_ID = [_a-zA-Z][_a-zA-Z0-9/.#:]*;
```

rust/tools/authorization-logic/test_inputs/hyphenIDs
```diff
@@ -0,0 +1,27 @@
...
+ */
+.decl ourFavoritePublicKeyIs(k : Key)
+.decl ourFavoriteTrustedBuilderIs(b : Builder)
+.decl variablesInsteadOfConstantsWithHyhensAreOKToo(v : TestVariable)
+"OakTeam" says {
+    ourFavoritePublicKeyIs(
+"""-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAErnyr7XY6gxXPnqRGYEpv5lDsbBeX
+izVFLckaF/cbgBuzx7kfmf1qi2j+3DDOQYgISQOj9/LYt4c1cxbt+XtG4g==
+-----END PUBLIC KEY-----""").
```
Cypher1, line 24: Okay, that's pretty cool though :D

Cypher1 (commented):
Just reading over PRs and was wondering about expanding the role of `ID` vs introducing a `STRING_LITERAL`

rust/tools/authorization-logic/test_inputs/hyphenIDs
aferr: This is the main use-case for multiline "id"s you were asking about. (Maybe "id" is not the right term for this piece of syntax). :)

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: Maybe "id" isn't the right term. This piece of syntax is really: principal names and predicate arguments. The different forms are either variables (which are similar to predicate arguments in souffle that are not wrapped in quotes) or the two different kinds of constants (which are similar to souffle arguments that *are* wrapped in quotes). I think when you're saying string literal you mean the same thing as what I am calling "constants" here.

You are right, it isn't useful to have miltiline identifiers. This syntax only permits multiple lines in the things that are constants (string literals).

rust/tools/authorization-logic/src/parsing/AuthLogic.g4
aferr: You are right. That old version was a hack, and the new version fixes it. Fortunately, that change was still fixed in #593  which did get merged in :)

aferr: (This PR existed only to save the very last commit which is the only way that this PR differs from #593 )

Issue #634, opened 2022-08-04
Support floats in the IR parser


asejfia:
The epsilon and delta values for dp will be floats (they are usually very small values). The IR parser should support them eventually.

asejfia:
@harsha-mandadi

Cypher1:
I wrote up a quick draft of this in https://github.com/Cypher1/raksha/tree/parse_floats as I've been thinking about float/double parsing a bit (in my own time) lately. If you wanted to use it @harsha-mandadi that'd be fine by me, or I can open a PR for it and see if it meets @asejfia needs :)

harsha-mandadi:
J, Thank you for working on this. Please go ahead and open a PR for the work you have already done. :) This looks similar to what @asejfia needs. (@asejfia to comment on this.)

Cypher1:
I've opened a PR and added handling so that any of:
- a decimal point e.g. `0.`
- an exponent e.g. `0e1`
- or a suffix e.g. `3l` or `3f`

will result in the value being considered a floating point, and therefore a `double` (no `float` support at the moment as I'm not sure it would be helpful).

The syntax is intended to match the `Absl::Atod` and `printf("%g", double_value)` standards, but with added support for the trailing suffix to avoid ambiguity.

e.g.

`3e0` is a `double`.
`3.0` is a `double`.
`3l` is a `double`.
`3` is an `int`.

Happy to rework things as needed.

PR #642, opened 2022-08-09
Parse floats: Early implementation of floating point (double) attribute values

Cypher1:


Cypher1:
Just an FYI CC: @bgogul @markww @aferr

Also: this is in response to #634

src/parser/ir/ir_parser_test.cc
```diff
@@ -93,6 +98,71 @@ INSTANTIATE_TEST_SUITE_P(IrParserTest, IrParserTest,
...
+
+TEST_P(IrParserNormalizingTest, DoublesAreNormalizedInTestOperations) {
+  auto program_text = GetParam();
+  IrProgramParser ir_parser;
+  auto result = ir_parser.ParseProgram(program_text.input);
+  EXPECT_EQ(IRPrinter::ToString(*result.module, *result.ssa_names),
+            program_text.output);
+}
+
+static const ParserInputAndExpectedOutput kNormalizedIR[] = {
```
harsha-mandadi, line 116: Is there a reason why we are using kNormalizedIR? Ideally we would want the input parsed text and outputs to be same. Also, the order of attributes seems to be different after normalization.

Cypher1: Yes :)

Round trip tests are useful, but they only check that normalized output can be parsed.
In my experience at least, parsers should not restrict input unnecessarily.
So, we should allow users to input non-normalized input (to avoid user frustration) and output normalized input.

This is what these tests are for (because it can be hard to work out what a normalized double will be, and similar with the order of attributes).

src/parser/ir/ir_parser.cc
```diff
@@ -75,6 +77,22 @@ IrProgramParser::ConstructOperationResult IrProgramParser::ConstructOperation(
       }
       auto* num_attribute_context =
```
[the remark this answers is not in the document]
Cypher1: Ah, sorry, that was a result of my testing the parser code. I'll put it back sorry.

src/parser/ir/ir_parser_test.cc
harsha-mandadi: I see. Sounds good!

harsha-mandadi (approved):
LGTM

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
```diff
@@ -41,7 +42,9 @@ using DatalogAttributePayload = ir::datalog::AttributePayload;
 
 static DatalogAttributePayload GetPayloadForAttribute(ir::Attribute attr,
                                                       ir::SsaNames &ssa_names) {
-  if (auto int_attr = attr.GetIf<ir::Int64Attribute>()) {
+  if (auto double_attr = attr.GetIf<ir::DoubleAttribute>()) {
+    return DatalogAttribute::Number(double_attr->value());
```
markww, line 46: This should not be `DatalogAttribute::Number`. `Number` is chosen to be parallel to Datalog's `Number` type which is actually its integer type. `Number`'s constructor accepts an `int64_t`. A new type of `DatalogAttribute` called `Float` should be introduced.

I think there must be a gap in testing; this should cause precision to be lost (ie, all the decimal places) which should be fairly easily detectable.

src/parser/ir/ir_parser_test.cc
```diff
@@ -21,17 +21,17 @@
 
 namespace raksha::ir {
 
-class IrParserTest : public testing::TestWithParam<absl::string_view> {};
+class IrParserRoundtripTest : public testing::TestWithParam<absl::string_view> {};
```
markww, line 24: I do worry about relying upon round tripping `double`s. Unlike `int`s, it's relatively easy to write `double`s that cannot be exactly represented and end up with a round trip that does not work. I guess if we find some cases that work and stick to them, this could be ok. Not sure if there's anything to do for this comment, just wanted to note my unease.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
Cypher1: Thanks! I'll add one as desired. I'm interested that the `Number` type doesn't require a conversion but will investigate.

src/parser/ir/ir_parser_test.cc
Cypher1: Yes, there's the normalizing tests for testing values that can't/don't roundtrip, this is just for testing with values that DO roundtrip.

I share your unease, but this was the best way I could think to handle both cases.

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
Cypher1: Ah looking at `souffle/src/include/souffle/TypeAttribute.h` I think it's actually `souffle::TypeAttribute::Signed` that our `Number` type corresponds to.

I'll update the comment to match unless I hear back :)

markww: Are you looking inside the Souffle codebase? I was just thinking about the primitive types exposed by Souffle. Maybe they call it `Signed` internally, but in the Datalog files, [it is called `number`](https://souffle-lang.github.io/types#number-type)

src/ir/attributes/double_attribute.h
```diff
@@ -0,0 +1,60 @@
...
+#include "absl/strings/str_format.h"
+#include "src/common/utils/intrusive_ptr.h"
+#include "src/ir/attributes/attribute.h"
+
+namespace raksha::ir {
+
+class DoubleAttribute;
+bool operator==(const DoubleAttribute& lhs, const DoubleAttribute& rhs);
+
+class DoubleAttribute : public AttributeBase {
```
bgogul, line 28: Not for this PR, but a general observation. It seems like int and double attributes could be templatized.

src/ir/datalog/value_test.cc
```diff
@@ -28,6 +28,27 @@ using testing::Combine;
...
+  std::string datalog_str = double_value.ToDatalogString();
+  double parsed_double = 0;
+  bool conversion_succeeds = absl::SimpleAtod(datalog_str, &parsed_double);
+  ASSERT_TRUE(conversion_succeeds)
+    << "Failed to convert " << datalog_str << " into a double (from " << num << ").";
+  ASSERT_EQ(parsed_double, num)
+    << "Failed to convert " << datalog_str << " into a double equal to " << num << ".";
+}
+
+static double kSampleDoubleValues[] = {0.0, 1.0, 2.0, -0.5, 45.1,
```
bgogul, line 45: formatting off?

Cypher1: Applied ClangFormat

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
Cypher1: Ah, thanks, that surprised me. Fixed them now.

Cypher1: Ah, I've found that there datalog_lowering_visitor_test has no number tests  and so I didn't discover it when adding Float as an analogue. Adding tests for both now.

markww: This still looks to me like it's using DatalogAttribute::Number, which takes a parameter of type int64_t.

src/ir/attributes/double_attribute.h
```diff
@@ -0,0 +1,60 @@
...
+
+  double value() const { return value_; }
+
+  bool IsEqual(const AttributeBase& other) const override {
+    if (this->kind() != other.kind()) return false;
+    return *this == static_cast<const DoubleAttribute&>(other);
+  }
+
+  std::string ToString() const override {
+    return absl::StrFormat("%gl", value_);
```
markww, line 44: I think you want `%lg`. [According to the docs for printf](https://en.cppreference.com/w/cpp/io/c/fprintf) the length modifier is the second-to-last character code in the print format, followed only by the format specifier itself. `g` is the format specifier, meaning "floating point in exponential notation" and `l` is the length specifier, meaning "long" or "double".

src/ir/datalog/value.h
```diff
@@ -37,7 +37,20 @@ class Value {
   virtual ~Value() {}
 };
 
-// Corresponds to Souffle's `number` type.
+// Corresponds to Souffle's `RamFloat` type.
+class Double : public Value {
```
markww, line 41: I'm not sure if this should be called `Double`, as it corresponds to `float` in Souffle. For all of the other types, we have preserved the name used in the Datalog language, only capitalizing it. It is pretty annoying that a type called `Float` would contain a `double`, but `double` does mean "double precision floating point value".

src/parser/ir/Ir.g4
```diff
@@ -57,10 +58,21 @@ ANY: '<<ANY>>';
 BLOCK: 'block';
 MODULE: 'module';
 
-
+SIGN : '+' | '-';
 NUMLITERAL : [0-9]+;
+FRACTIONAL_PART : '.' [0-9]*;
+EXPONENT: ('e'|'E') SIGN? [0-9]+;
+DOUBLE_TAIL
+    : (FRACTIONAL_PART EXPONENT? ('l' | 'f')?)
```
markww, line 66: Does it make sense to have a distinction between `l` and `f`? If we put everything in double precision, we don't really need `f`.

src/ir/attributes/double_attribute.h
markww: Actually, I think you want `%a`. `%g` goes out to a fixed precision, whereas `%a` goes out to the minimum precision needed to exactly represent the value.

Cypher1: Here `l` is a suffix meant to ensure that double values like `3.0` are not accidentally converted to `int` (as %g prints them as just `3`).

I'll test `%lgl` though.

Cypher1:
> Actually, I think you want `%a`. `%g` goes out to a fixed precision, whereas `%a` goes out to the minimum precision needed to exactly represent the value.

Thank you! I thought I had read the whole printf reference but I somehow missed %a. Will get back to it in a bit :)

src/ir/attributes/double_attribute.h
Cypher1: Hmm, looking at the doc I think `%a` is exclusively hexadecimal. So far the best results I've seen come from `%g` and `%lg` (I'll use lg for now)

src/backends/policy_engine/souffle/datalog_lowering_visitor.cc
Cypher1: +1 Sorry for the slow response time, I've been trying to work out why this is building.

Currently, there are no tests for int64 or double attributes lowering to datalog, but the GetIf gives back an ir::FloatAttribute (ptr) and the `value()` call gets a `double`. Additionally the Number constructor is marked as `explicit` so it shouldn't be allowed to cast the double to an int.

Can you illuminate this for me?

markww: I think you might be misunderstanding what `explicit` does for C++ constructors.

If you have a single-argument C++ constructor not marked explicit of the form:

```C++
class Foo {
  Foo(Bar bar) /*...*/
};
```

What this means is that `Foo` can be constructed from `Bar` *even as an implicit cast*. Like many cases where C++ is trying to be helpful, this is generally considered a bad move. Fortunately, `explicit` turns it off.

Now in this case, the problem is that the constructor accepts an `int64_t` but you are passing in a `double`. In languages with strong typing, this is considered Very Bad and causes a compiler error. [C++, however, tries to be helpful](https://en.cppreference.com/w/cpp/language/implicit_conversion):

```
A [prvalue](https://en.cppreference.com/w/cpp/language/value_category#prvalue) of floating-point type can be converted to a prvalue of any integer type. The fractional part is truncated, that is, the fractional part is discarded. If the value cannot fit into the destination type, the behavior is undefined (even when the destination type is unsigned, modulo arithmetic does not apply). If the destination type is bool, this is a boolean conversion (see below)
```

Unfortunately this cannot be turned off without going to extreme lengths (like wrapping all primivite types in classes and defining `operator`s to give you stronger typing, a move that seems like a bridge too far for pretty much everyone).

Cypher1:
> I think you might be misunderstanding what `explicit` does for C++ constructors.
> 
> If you have a single-argument C++ constructor not marked explicit of the form:
> 
> ```C++
> class Foo {
>   Foo(Bar bar) /*...*/
> };
> ```
> 
> What this means is that `Foo` can be constructed from `Bar` *even as an implicit cast*. So the following line of code will pass the compiler:
> 
> ```C++
> void example(Bar bar) {
>   Foo my_foo = bar;
>   /* ... */
> }
> ```
> 
> Like many cases where C++ is trying to be helpful, this is generally considered a bad move. Fortunately, `explicit` turns it off.
> 
> Now in this case, the problem is that the constructor accepts an `int64_t` but you are passing in a `double`. In languages with strong typing, this is considered Very Bad and causes a compiler error. [C++, however, tries to be helpful](https://en.cppreference.com/w/cpp/language/implicit_conversion):
> 
> """
> A [prvalue](https://en.cppreference.com/w/cpp/language/value_category#prvalue) of floating-point type can be converted to a prvalue of any integer type. The fractional part is truncated, that is, the fractional part is discarded. If the value cannot fit into the destination type, the behavior is undefined (even when the destination type is unsigned, modulo arithmetic does not apply). If the destination type is bool, this is a boolean conversion (see below)
> """
> 
> Unfortunately this cannot be turned off without going to extreme lengths (like wrapping all primivite types in classes and defining `operator`s to give you stronger typing, a move that seems like a bridge too far for pretty much everyone).


Thanks Mark,

You're right, I had thought that double to int conversions would also be disallowed by we `explicit`.

It seems the only way to avoid this accident reoccurring is something like the following (which gives nicer, deleted errors when the code would otherwise trigger a loss of precision):

```
template<typename T>
Number(T) = delete;

Number(int value): value_{value} {}
Number(int64_t value): value_{value} {}

```

Could we introduce this to avoid future issues?

Anyway, on my Monday I'll write number and float lowering tests :)

markww:
Unfortunately, that template solution also would not work. When choosing between overloads, C++ populates an overload set for a type and checks to see what matches in priority order. An exact type match is preferred, so `Number(double value)` will beat out `Number(int64_t value)`. But if there is no other option in the overload set, `int64_t` is considered a valid possibility, as there as an implicit conversion from `double` to `int64_t`.

If we wanted to prevent this error from occurring, we could define the following constructor:

```C++
explicit Number(double) {
  LOG(FATAL) << "Attempt to construct a Souffle `Number` from a floating point type!";
}
```

I think that could work.

src/parser/ir/Ir.g4
Cypher1: I've dropped support for 'f'. I had introduced it to attempt to match C++ (with the intention of adding float/double values, but I think the higher accuracy values are better for our purposes anyway).

Cypher1:
> Unfortunately, that template solution also would not work. When choosing between overloads, C++ populates an overload set for a type and checks to see what matches in priority order. An exact type match is preferred, so `Number(double value)` will beat out `Number(int64_t value)`. But if there is no other option in the overload set, `int64_t` is considered a valid possibility, as there as an implicit conversion from `double` to `int64_t`.
> 
> If we wanted to prevent this error from occurring, we could define the following constructor:
> 
> ```c++
> explicit Number(double) {
>   LOG(FATAL) << "Attempt to construct a Souffle `Number` from a floating point type!";
> }
> ```
> 
> I think that could work.

Thanks Mark,
I tested it in https://godbolt.org/z/o7G78oq9s and it does seem to behave as I had hoped, with appropriate static failure (rather than the runtime error you proposed). I think this is safer (than introducing new runtime failures), and does disallow the current (accidental) bad behavior. I'm also open to explicitly deleting the `double` and `float` cases, as they seem very likely to cause errors in future.

Cypher1:
@markww added to an existing test and fixed the ordering as we discussed. Changes are in https://github.com/google-research/raksha/pull/642/commits/b0539667b6607b32b790cb3fe358401e20679873

PR #643, opened 2022-08-09
Visitor for authorization logic AST

aferr:


src/ir/ir_visitor.h
```diff
@@ -31,7 +31,7 @@ class IRVisitor {
  public:
   virtual ~IRVisitor() {}
   virtual Result Visit(CopyConst<IsConst, Module>& module) = 0;
-  virtual Result Visit(CopyConst<IsConst, Block>& operation) = 0;
+  virtual Result Visit(CopyConst<IsConst, Block>& block) = 0;
```
aferr, line 34: FYI This looked like a typo so I deliberately fixed this here.

Cypher1: Thanks! I wish cpp kept us honest about these.

Cypher1 (approved):
Lgtm

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
```diff
@@ -0,0 +1,322 @@
...
+  Result Visit(
+      CopyConst<IsConst, SaysAssertion>& saysAssertion) final override {
+    Result pre_visit_result = PreVisit(saysAssertion);
+    Result principal_result = FoldResult(
+        std::move(pre_visit_result), saysAssertion.principal().Accept(*this));
+    Result fold_result = common::utils::fold(
+        saysAssertion.assertions(), std::move(principal_result),
+        [this](Result acc, CopyConst<IsConst, Assertion> assertion) {
+          return FoldResult(std::move(acc), assertion.Accept(*this));
+        });
```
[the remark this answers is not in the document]
aferr: I started doing this, but because the body of the function you suggested calls `Accept` on element, the template really needs to specify that Element implements this method. One way to do this would be to write an abstract class that defines the methods all AST nodes have and include `Accept` in this abstract class and then have all the AST nodes extend from that. This seems like a more substantial change that belongs in another PR. So I wrote an issue to track this: #648

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
```diff
@@ -0,0 +1,322 @@
...
+    Result variant_visit_result = std::visit(
+        raksha::utils::overloaded{
+            [this](const datalog::Predicate& pred) {
+              return VariantVisit(pred);
+            },
+            [this](const Attribute& attrib) { return VariantVisit(attrib); },
+            [this](const CanActAs& canActAs) {
+              return VariantVisit(canActAs);
+            }},
+        baseFact.GetValue());
```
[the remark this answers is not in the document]
aferr: That doesn't quite work unfortunately... I tried what you suggested tweaking it slightly to the following because it's missing a param and a capture to the following:

```
    Result variant_visit_result = std::visit([*this](const auto &branch) {
      return VariantVisit(branch);
    }, base_fact.GetValue());
```

and auto is not able to deduce the type:
```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/variant:548:12: error: cannot deduce return type 'auto' from returned value of type '<overloaded function type>'
```

along with this other cryptic error I did not dig into:
```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/variant:659:14: error: attempt to use a deleted function
      return _VSTD::__invoke_constexpr(_VSTD::forward<_Visitor>(__visitor),
```

 I remember running into a similar issue when implementing `lowering_ast_datalog`.

I think actually the better way to remove this code would be to remove the use of `variant`s entirely from the implementation of the AST and replace them with sublcasses representing the different forms. This visitor is already implementing the pattern in a way that uses dynamic dispatch and inheritance to call the `Accept` implementation specific to the subclasses of the nodes. If you think that makes sense, I'll open an issue. Otherwise, it's left as-is.

src/ir/auth_logic/ast.h
```diff
@@ -34,6 +35,16 @@ class Principal {
   explicit Principal(std::string name) : name_(std::move(name)) {}
   const std::string& name() const { return name_; }
 
+  template <typename Derived, typename Result>
+  Result Accept(AuthLogicAstVisitor<Derived, Result, false>& visitor) {
```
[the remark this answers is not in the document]

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
```diff
@@ -0,0 +1,322 @@
...
+  Result VariantVisit(Attribute attribute) { return attribute.Accept(*this); }
+  Result VariantVisit(CanActAs canActAs) { return canActAs.Accept(*this); }
+
+  // For AssertionVariantType
+  Result VariantVisit(Fact fact) { return fact.Accept(*this); }
+  Result VariantVisit(ConditionalAssertion conditionalAssertion) {
+    return conditionalAssertion.Accept(*this);
+  }
+
+ private:
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
```diff
@@ -0,0 +1,322 @@
...
+  }
+
+  Result Visit(CopyConst<IsConst, Assertion>& assertion) final override {
+    Result pre_visit_result = PreVisit(assertion);
+    Result variant_visit_result =
+        std::visit(raksha::utils::overloaded{
+                       [this](const Fact& fact) { return VariantVisit(fact); },
+                       [this](const ConditionalAssertion& condAssertion) {
+                         return VariantVisit(condAssertion);
+                       }},
```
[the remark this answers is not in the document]
aferr: see above. resolving this thread.

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
```diff
@@ -0,0 +1,322 @@
...
+    Result fold_result = FoldResult(std::move(pre_visit_result),
+                                    std::move(variant_visit_result));
+    return PostVisit(assertion, std::move(fold_result));
+  }
+
+  Result Visit(
+      CopyConst<IsConst, SaysAssertion>& saysAssertion) final override {
+    Result pre_visit_result = PreVisit(saysAssertion);
+    Result principal_result = FoldResult(
+        std::move(pre_visit_result), saysAssertion.principal().Accept(*this));
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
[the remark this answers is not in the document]
markww: Sounds good, if we're already using subclasses and vtables to subtype, might as well use them everywhere. I think that's a good way forward. Unfortunate that the generic lambda didn't work :(

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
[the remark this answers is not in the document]
markww: Templates don't need to specify anything; they'll just try subbing in the type implied by the arguments into the type parameter and see whether the code compiles.

Although, I do now realize that the order of the `Container` and `Element` type parameters should be swapped, as `Container` is inferrable from arguments and `Element` will probably need to be explicitly specified.

markww (approved):
LGTM modulo the last few (relatively minor) threads here.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -0,0 +1,79 @@
...
+
+namespace raksha::ir::auth_logic {
+namespace {
+
+// A visitor that makes a set of all the names of principals in the program
+class PrincipalNameCollectorVisitor
+    : public AuthLogicAstTraversingVisitor<PrincipalNameCollectorVisitor,
+                                           absl::flat_hash_set<std::string>> {
+ public:
+  PrincipalNameCollectorVisitor() {}
```
bgogul, line 32: Redundant?

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -0,0 +1,79 @@
+//-----------------------------------------------------------------------------
```
bgogul, line 1: I don't think the files are formatted. Can you use clang-format with google3 style to format them?

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -0,0 +1,79 @@
...
+class PrincipalNameCollectorVisitor
+    : public AuthLogicAstTraversingVisitor<PrincipalNameCollectorVisitor,
+                                           absl::flat_hash_set<std::string>> {
+ public:
+  PrincipalNameCollectorVisitor() {}
+
+  absl::flat_hash_set<std::string> GetDefaultValue() override { return {}; }
+
+  absl::flat_hash_set<std::string> CombineResult (
+      absl::flat_hash_set<std::string> acc,
```
bgogul, line 37: Is it correct that these should be passed by value?

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -0,0 +1,79 @@
...
+
+#include "absl/container/flat_hash_set.h"
+#include "src/common/testing/gtest.h"
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
+namespace {
+
+// A visitor that makes a set of all the names of principals in the program
```
bgogul, line 27: Test is not comprehensive enough? It only tests previsit of a `Principal`?

bgogul (commented):
Can you add some comprehensive tests? This is a lot of non-trivial code without sufficient tests.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: should be OK now.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: I made this more comprehensive.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: I left this open in case I am missing something though, which might be the case.

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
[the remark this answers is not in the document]
aferr: Ah I mixed up templates with Generics from Java again. Swapping the order must have been it. It's implemented now.

aferr:
@bgogul I'm not blocked on this yet so I thought I would wait for you to re-review before merging it in. Let me know if it looks OK if you have a moment.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -0,0 +1,279 @@
...
+
+  const std::vector<std::string>& nodes() const { return nodes_; }
+
+ private:
+  bool pre_visits_;
+  bool post_visits_;
+  std::vector<std::string> nodes_;
+};
+
+TEST(AuthLogicAstTraversingVisitorTest, SimpleTraversalTest) {
```
bgogul, line 187: It might be more straightforward with a gmock, but it is fine for now. 

The only question is if there is a way to break this down into smaller tests.

bgogul (approved):
Approving, but left some clean up comments. If you don't address these, can you file issues (e.g., gmock, debugprint as visitor, etc.)

Also, try to keep PRs smaller.

src/ir/auth_logic/ast.h
```diff
@@ -103,6 +175,28 @@ class Fact {
...
+
+  template <typename Derived, typename Result>
+  Result Accept(
+      AuthLogicAstVisitor<Derived, Result, Immutable>& visitor) const {
+    return visitor.Visit(*this);
+  }
+
+  // A potentially ugly print of the state in this class
+  // for debugging/testing only
+  std::string DebugPrint() const {
```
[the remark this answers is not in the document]
aferr: Right, this is just for testing the visitor for now.

src/ir/auth_logic/auth_logic_ast_traversing_visitor.h
```diff
@@ -0,0 +1,317 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+#ifndef SRC_IR_AUTH_LOGIC_AST_TRAVERSING_VISITOR_H_
+#define SRC_IR_AUTH_LOGIC_AST_TRAVERSING_VISITOR_H_
+
+#include <glog/logging.h>
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: I'm assuming it's enough and resolving :)

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: I'm assuming this was OK and resolving.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: Opened an issue just because there is another PR depending on this one (and IDK or don't remember what a gmock is yet) #666

aferr:
> Also, try to keep PRs smaller.

@bgogul Writing code in smaller increments doesn't come so naturally to me. This one seems like a self-contained increment to me. How would I split this into smaller self-contained PRs? Would it be to separately write:
 - just the visitor
 - then the traversing visitor + changes to the AST nodes
 - then add the test
Or what would be a sensible way to split this into smaller PRs?

PR #653, opened 2022-08-12
Check only source

Cypher1:
A few changes to make the default license check ignore:

- `.auth` as well as `.authlogic` files (we're split between the two)
- `testdata` directories (as some of the data is either generated or compared with generated data)
- non-source files

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -34,6 +34,7 @@ SOURCE_FILES_PATTERNS=(
   ':!*.txt'
 
   # Exclude test files.
+  ':!*.auth'
   ':!*.authlogic'
   ':!*.raksha'
   ':!*.ir'
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -37,6 +37,7 @@ SOURCE_FILES_PATTERNS=(
   ':!*.authlogic'
   ':!*.raksha'
   ':!*.ir'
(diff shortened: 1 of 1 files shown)


aferr (approved):
Looks good to me! I think technically, if we wanted to we could add licenses to the auth logic files since they do support comments, but since we are migrating a way from the rust implementation of auth logic, it's not worth the effort to do that now. If we keep the test inputs post migration to C++ then it may be worth adding licenses to them at that point.

Cypher1:
I skipped licenses on the output files to avoid making the code generator to spit out a licence (that might be out of date or invalid some other way).

The input files you're right about, we can add them in a follow-up.

Cypher1:
Requesting merge: @bgogul, @markww  or @aferr

aferr:
> Requesting merge: @bgogul, @markww or @aferr

You don't have permissions to merge? Are you blocked by waiting to merge this? If you are not blocked, I'm inclined to wait until one of either @bgogul or @markww also approve.

markww (approved):
LGTM, will merge for you.

Cypher1:
> > Requesting merge: @bgogul, @markww or @aferr
> 
> You don't have permissions to merge? Are you blocked by waiting to merge this? If you are not blocked, I'm inclined to wait until one of either @bgogul or @markww also approve.

I do not have merge permissions. I can use readytopull labels for g3 code but this is Git only.

PR #654, opened 2022-08-15
Universe relations@aferr

### LINKED ISSUE #636: Add Universe Relation Population to C++ Impl of Auth Logic
This can follow the [equivalent pass from the rust implementation](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/universe_translation.rs)

This would be enhanced by #345

aferr:
This fixes #636  and ports [this rust code](https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/souffle/universe_translation.rs) to raksha.


aferr:
This still needs tests, but I figured I would probably start a PR because of the time zone difference.

aferr:
This still needs more thorough tests, but it is probably in OK enough shape to be reviewed at this point @bgogul @markww

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,450 @@
...
+        type_env_.insert({arg_name, arg_type});
+      } else if (type_env_.find(arg_name)->second != arg_type) {
+        // If there is already a typing for this constant,
+        // but it is different from the one used here,
+        // throw a type error and abort.
+        LOG(FATAL) << "type error for constant: " << arg_name;
+      }
+    }
+
+    // TODO once an AST node for numbers as RVALUEs in numeric
```
aferr, line 222: Note to self to do the implied issue management.

aferr:
@bgogul This is another one that probably should have been more, smaller PRs. To do this, would a good way to split it be to implement and test each of these steps as distinct PRs:
- `// Generate relation declaration environment`
- `// Generate constant type environment`
- `// Generate universe populating facts`
- `// Transform assertions to include conditions that check the universes`

??

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    }
+    return universe_defining_says_assertions;
+  }
+
+  // This visitor populates a mapping (flat_hash_map) of constant (literal)
+  // names to their types by looking at the relations to which the literals
+  // are applied and declarations of those relations (including the types
+  // of the parameters where they are applied). It takes as input
+  // a mapping from relation names to relation declarations.
+  class TypeEnvironmentGenerationVisitor
```
markww, line 187: Usually, in C++, it's a good rule to have only one meaty `class` definition per header. You can sprinkle in a few small supporting `struct`s and `class`es and `enum`s, but another non-trivial `class` should get its own header.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+      if (type_env_.find(arg_name) == type_env_.end()) {
+        // No typing for this constant was found yet,
+        // so add it.
+        type_env_.insert({arg_name, arg_type});
+      } else if (type_env_.find(arg_name)->second != arg_type) {
+        // If there is already a typing for this constant,
+        // but it is different from the one used here,
+        // throw a type error and abort.
+        LOG(FATAL) << "type error for constant: " << arg_name;
+      }
```
markww, line 219: This can be made more concise as follows:

```C++
auto insert_result = type_env.insert({arg_name, arg_type});
CHECK(insert_result.second) << "Type error for constant: " << insert_result.first->first;
```

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+        absl::flat_hash_map<std::string, datalog::RelationDeclaration>;
+
+    TypeEnvironmentGenerationVisitor(
+        absl::flat_hash_map<std::string, datalog::RelationDeclaration> decl_env)
+        : decl_env_(decl_env) {}
+
+    const TypeEnv& type_env() const { return type_env_; }
+
+   private:
+    void AddTyping(std::string arg_name, datalog::ArgumentType arg_type) {
```
markww, line 204: If you can't, or don't want to, use std::move to ensure efficient copy elision or moving of `arg_name`, use an `absl::string_view` to take the parameter in instead.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    // (https://github.com/google-research/raksha/blob/be6ef8e1e1a20735a06637c12db9ed0b87e3d2a2/src/ir/auth_logic/ast_construction.cc#L92)
+    static inline bool PredicateIsNumericOperator(
+        const datalog::Predicate& pred) {
+      if (pred.name() == "<" || pred.name() == "<" || pred.name() == "=" ||
+          pred.name() == "!=" || pred.name() == "<=" || pred.name() == ">=") {
+        return true;
+      } else {
+        return false;
+      }
+    }
```
markww, line 238: Should this just be a member function on `datalog::Predicate`?

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+        // than a numeric operator
+        datalog::RelationDeclaration decl =
+            GetDeclarationOrFatal(decl_env_, pred.name());
+        // The relation declarations give the types for each position in the
+        // predicate. For the xth argument in the predicate, we want to
+        // assign it the type for the xth parameter in the relation declaration.
+        // So to do this, we iterate through the list of arguments in the
+        // predicate and the list of parameters in the declaration using
+        // the indices of the arguments iterator of the predicate.
+        for (unsigned i = 0; i < pred.args().size(); i++) {
```
markww, line 264: It's highly unlikely that this will ever be the case, but `unsigned` isn't long enough to cover all of the range of `pred.args().size()` on a 64 bit system, which is of type `size_t`. [`unsigned` is not preferred to be used anyway](https://google.github.io/styleguide/cppguide.html#Integer_Types). I'd say make `i` of type `size_t` or `uint64_t`.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+          AddTyping(pred.args()[i], decl.arguments()[i].argument_type());
```
markww, line 265: Rather than restating `pred.args()` multiple times, it might be nice to just create a temporary to hold a reference to it, such as:

```C++
const PredicateArgument & /* or whatever the appropriate type is*/ predicate_arguments = pred.args();
```

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+      left.merge(std::move(right));
+      return left;
+    }
+
+    DeclSet Visit(
+        const datalog::RelationDeclaration& relation_declaration) override {
+      DeclSet universe_declarations;
+      for (const auto& argument : relation_declaration.arguments()) {
+        universe_declarations.emplace(
+            TypeToUniverseDeclaration(argument.argument_type()));
```
markww, line 334: In general, just use `insert`. You should only use `emplace` rather than `insert` when forced to by the compiler.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    // The `AddUniverseConditions` methods, including this one,
+    // are not implemented as actual visits because these methods
+    // are implementing a transformation, and the returned values
+    // need to be placed in a part of the syntactic context that
+    // is specific to each syntactic element.
+    SaysAssertion AddUniverseConditions(const SaysAssertion& says_assertion) {
+      std::vector<Assertion> assertion_list;
+      for (const Assertion& assertion : says_assertion.assertions()) {
+        assertion_list.push_back(AddUniverseConditions(assertion));
+      }
```
markww, line 370: This could maybe be made more concise with `map_iter` from `//src/common/utils`.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+      } else {
+        return {};
+      }
+    }
+
+    // Add universe conditions for the non-constant arguments
+    std::vector<BaseFact> Visit(const datalog::Predicate& pred) override {
+      auto decl = GetDeclarationOrFatal(decl_env_, pred.name());
+      std::vector<BaseFact> conditions;
+      for (unsigned i = 0; i < pred.args().size(); i++) {
```
markww, line 433: Don't use `unsigned`, use `size_t` or `uint64_t` instead.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+        return {MakeUniverseMembershipFact(
+            prin.name(), datalog::ArgumentType::MakePrincipalType())};
+      } else {
+        return {};
+      }
+    }
+
+    // Add universe conditions for the non-constant arguments
+    std::vector<BaseFact> Visit(const datalog::Predicate& pred) override {
+      auto decl = GetDeclarationOrFatal(decl_env_, pred.name());
```
markww, line 431: Add explicit type annotation.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
+/*
```
markww, line 1: There is way too much stuff in this header file. While it is often preferable to put the definitions of small methods directly in the header, it should usually be the case that someone can pop open a header and scan it to get an idea of what the class does. Once the details become complex, they should be moved to an associated `.cc` file.

In particular, this file contains a number of `static` methods which do not take advantage of special access to private members of the class. That's a warning sign of bad factoring: maybe those shouldn't be members of the class at all, or should be helpers in an unnamed namespace in the `.cc` file.

Inner classes in a header file are similarly questionable. When they are exposed to clients, it may be right to put them in the header. Same for if they are very small and putting them in the header prevents from having to create a `.cc` file. Otherwise, maybe they should be classes in an anonymous namespace in the `.cc` file or should be broken out into their own header.

src/ir/auth_logic/universe_relation_insertion_test.cc
```diff
@@ -0,0 +1,142 @@
...
+       SimpleProgramDebugStringEquivalenceTest) {
+  const auto& [input_program, expected_program] = GetParam();
+  // declarations
+  EXPECT_THAT(GetDebugVector(input_program.relation_declarations()),
+              UnorderedElementsAreArray(
+                  GetDebugVector(expected_program.relation_declarations())));
+  // says_assertions
+  EXPECT_THAT(GetDebugVector(input_program.says_assertions()),
+              UnorderedElementsAreArray(
+                  GetDebugVector(expected_program.says_assertions())));
```
markww, line 129: We should not use these `std::vector`s of debug prints. See #676 .

src/ir/auth_logic/universe_relation_insertion_test.cc
```diff
@@ -0,0 +1,142 @@
...
+      UnorderedElementsAreArray(GetDebugVector(expected_program.queries())));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    Something, ProgramDebugStringEquivalenceTest,
+    testing::Values(ProgramEquivalenceTestData{
+        UniverseRelationInsertion::InsertUniverseRelations(BuildTestProgram1()),
+        BuildTestProgram1Expected()}));
+
+}  // namespace raksha::ir::auth_logic
```
markww, line 142: Newline at EOF.

src/ir/auth_logic/universe_relation_insertion_test.cc
```diff
@@ -0,0 +1,142 @@
...
+  EXPECT_THAT(
+      GetDebugVector(input_program.queries()),
+      UnorderedElementsAreArray(GetDebugVector(expected_program.queries())));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    Something, ProgramDebugStringEquivalenceTest,
+    testing::Values(ProgramEquivalenceTestData{
+        UniverseRelationInsertion::InsertUniverseRelations(BuildTestProgram1()),
+        BuildTestProgram1Expected()}));
```
markww, line 140: There's probably not much need to make a test parameterized if you only instantiate it with a single argument.

src/ir/auth_logic/universe_relation_insertion.h
aferr: this was done.

src/ir/auth_logic/universe_relation_insertion_test.cc
aferr: OK. This seems like it should be done as a separate PR though so I will fix this separately and keep it as-is with this PR if that's OK.

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion_test.cc
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+class UniverseRelationInsertion {
+ public:
+  // A map from the names of principal / predicate argument literals
+  // (the constants wrapped in quotes) to typings.
+  using TypeEnv = absl::flat_hash_map<std::string, datalog::ArgumentType>;
+  // A map from predicate names to predicate declarations.
+  using DeclEnv =
+      absl::flat_hash_map<std::string, datalog::RelationDeclaration>;
+
+  static Program InsertUniverseRelations(const Program& prog) {
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+namespace raksha::ir::auth_logic {
+
+class UniverseRelationInsertion {
+ public:
+  // A map from the names of principal / predicate argument literals
+  // (the constants wrapped in quotes) to typings.
+  using TypeEnv = absl::flat_hash_map<std::string, datalog::ArgumentType>;
+  // A map from predicate names to predicate declarations.
+  using DeclEnv =
+      absl::flat_hash_map<std::string, datalog::RelationDeclaration>;
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    }
+  }
+
+  static BaseFact MakeUniverseMembershipFact(std::string element_name,
+                                             datalog::ArgumentType arg_type) {
+    return BaseFact(
+        datalog::Predicate(TypeToUniverseName(arg_type),  // universe relation
+                           {element_name},  // the element to add to universe
+                           datalog::Sign::kPositive));
+  }
```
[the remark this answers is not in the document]
aferr: With the current functionality of this compiler pass there is no additional value in distinguishing universe membership facts from other things. I think if we see some actual functional reason to distinguish these from other kinds of facts we can implement that to meet this goal when we see this reason. Otherwise we are adding complexity to the code and spending time to implement this without gaining a benefit.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+
+  static datalog::RelationDeclaration TypeToUniverseDeclaration(
+      datalog::ArgumentType arg_type) {
+    return datalog::RelationDeclaration(TypeToUniverseName(arg_type), false,
+                                        {datalog::Argument("x", arg_type)});
+  }
```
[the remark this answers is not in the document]
aferr: Re universe declaration subclass: similar opinion about the UniverseMembershipFact suclass.

Re static methods: this is fixed.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+ private:
+  static std::string TypeToUniverseName(datalog::ArgumentType arg_type) {
+    switch (arg_type.kind()) {
+      case datalog::ArgumentType::Kind::kNumber:
+        return "isNumber";
+      case datalog::ArgumentType::Kind::kPrincipal:
+        return "isPrincipal";
+      case datalog::ArgumentType::Kind::kCustom:
+        return absl::StrCat("is", arg_type.name());
+    }
```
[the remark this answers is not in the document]
aferr: With switch we have one less vtable lookup, right?

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    // that check the universes
+    UniverseConditionTransformer universe_condition_transformer(decl_env);
+    for (const auto& says_assertion : prog.says_assertions()) {
+      new_says_assertions.push_back(
+          universe_condition_transformer.AddUniverseConditions(says_assertion));
+    }
+
+    // Add declarations for universe relations.
+    auto universe_declarations = GetUniverseDeclarations(prog);
+    auto new_declarations = prog.relation_declarations();
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+                                        {datalog::Argument("x", arg_type)});
+  }
+
+  // Determine if an identifier (which may be an argument of a predicate
+  // or a principal name) is a constant or variable.
+  // TODO(#673 aferr) it would be better to refactor the AST so that
+  // identifiers are a separate AST node with constant and non-constant
+  // children determined at parse time rather than here since
+  // this information is available at parse time.
+  static bool IsNameConstant(const std::string& id) {
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+  }
+
+  static datalog::ArgumentType GetTypingOrFatal(
+      const TypeEnv& type_env, const std::string& argument_name) {
+    if (type_env.find(argument_name) == type_env.end()) {
+      LOG(FATAL) << "could not find typing for argument" << argument_name;
+    } else {
+      return type_env.find(argument_name)->second;
+    }
+  }
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+  // identifiers are a separate AST node with constant and non-constant
+  // children determined at parse time rather than here since
+  // this information is available at parse time.
+  static bool IsNameConstant(const std::string& id) {
+    if (id[0] == '"') {
+      return true;
+    } else {
+      // Numeric literals are constants and this is
+      // a numeric literal if all the characters are numbers.
+      for (const char& c : id) {
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    }
+  }
+
+  static datalog::ArgumentType GetTypingOrFatal(
+      const TypeEnv& type_env, const std::string& argument_name) {
+    if (type_env.find(argument_name) == type_env.end()) {
+      LOG(FATAL) << "could not find typing for argument" << argument_name;
+    } else {
+      return type_env.find(argument_name)->second;
+    }
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+  }
+
+  static datalog::RelationDeclaration GetDeclarationOrFatal(
+      const DeclEnv& decl_env, const std::string& predicate_name) {
+    if (decl_env.find(predicate_name) == decl_env.end()) {
+      LOG(FATAL) << "could not find declaration for relation" << predicate_name;
+    } else {
+      return decl_env.find(predicate_name)->second;
+    }
+  }
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,449 @@
...
+    for (const auto& prin : constant_principals) {
+      for (const auto& typing : type_env) {
+        universe_defining_says_assertions.push_back(SaysAssertion(
+            prin,  // speaker
+            {Assertion(Fact({}, MakeUniverseMembershipFact(typing.first,
+                                                           typing.second)))}));
+      }
+    }
+    return universe_defining_says_assertions;
+  }
```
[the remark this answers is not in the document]
aferr: I instead made this a function in the anonymous namespace of the `.cc` file. My rationale is that this function seems like an implementation detail of the compiler pass, whereas `TypeEnv` could plausibly be re-used in other passes (which is why `TypeEnv` is still in the header at all).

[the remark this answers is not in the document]
aferr: I left this thread open. I'll do it the way you suggested if you still think that's better after the above reply :)

src/ir/auth_logic/universe_relation_insertion.h
aferr: I think this may now have 0 meaty classes in the header and so this is fixed now. I'll leave this one open to see if you agree with that.

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed.

aferr:
@markww just FYI I am only part-way through your comments before stepping away from the keyboard, but I will come back some other time :)

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+// this information is available at parse time.
+bool IsNameConstant(absl::string_view id) {
+  if (id[0] == '"') {
+    return true;
+  } else {
+    // Numeric literals are constants and this is
+    // a numeric literal if all the characters are numbers.
+    for (char c : id) {
+      if (!isdigit(c)) return false;
+    }
```
markww, line 36: Could you save yourself some custom code by using something like [`absl::SimpleAtoi`](https://github.com/abseil/abseil-cpp/blob/master/absl/strings/numbers.h#L62)?

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+// comparisons is added, this visitor should also visit
+// the numeric RVALUEs and add them to the type environment
+// with type ArgumentType(kNumber, "Number")
+// This is a workaround that makes use of the fact that
+// numeric comparisons are represented as predicates
+// with a name that matches the operator:
+// (https://github.com/google-research/raksha/blob/be6ef8e1e1a20735a06637c12db9ed0b87e3d2a2/src/ir/auth_logic/ast_construction.cc#L92)
+static inline bool PredicateIsNumericOperator(const datalog::Predicate& pred) {
+  if (pred.name() == "<" || pred.name() == "<" || pred.name() == "=" ||
+      pred.name() == "!=" || pred.name() == "<=" || pred.name() == ">=") {
```
markww, line 51: In general, it's best to not repeat a method call over and over when you expect it to always return the same thing. Just call it once and assign it to a temporary.

Also, `if (predicate) return true; else return false;` is equivalent to `return predicate`.

Also, I wonder whether it might be more readable to just create a set of the options and look this up in the set like so:

```C++
static inline bool PredicateIsNumericOperator(const datalog::Predicate &pred) {
  const absl::flat_hash_set numeric_operators = {
   "<", ...
  };
  return numeric_operators.contains(pred.name());
}
```

Also, I think you may have accidentally repeated `<` rather than put in `>`.

src/ir/auth_logic/universe_relation_insertion.h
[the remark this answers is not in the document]
markww: Yeah, but that's the kind of thing you should only care about if you're getting complaints about code being too slow and your profiler points to this happening over and over again.

A single vtable lookup doesn't matter for time complexity and at a high level, time complexity is the only thing that matters (well, space complexity too, but that's usually harder to make too much of a mess with). Looking a tiny bit more closely, you can help performance out without much additional effort or readability impact with some well known patterns, like "use offline sorts instead of online sorts when possible" or "use densely allocated structures rather than pointer-heavy structures when possible" or "reserve your space in your vector in advance when possible instead of leaning on `push_back`". But even those matter only sometimes.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+// constants of all types, and then generates a SaysAssertion that
+// adds the constant to the universe for each speaker.
+std::vector<SaysAssertion> GetUniverseDefiningFacts(
+    const TypeEnvironment& type_env) {
+  std::vector<Principal> constant_principals;
+  for (const auto& typing : type_env.inner_map()) {
+    if (typing.second.kind() == datalog::ArgumentType::Kind::kPrincipal) {
+      constant_principals.push_back(Principal(typing.first));
+    }
+  }
```
markww, line 95: Nothing for you to do on this comment, but seeing this code makes me want to put a functional-style `filter` in our `//src/common/utils`.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+      universe_defining_says_assertions.push_back(SaysAssertion(
+          prin,  // speaker
+          {Assertion(Fact(
+              {}, MakeUniverseMembershipFact(typing.first, typing.second)))}));
+    }
+  }
+  return universe_defining_says_assertions;
+}
+
+// This visitor is used only to implmenet GetUniverseDeclarations
```
markww, line 109: Nit: typo: iplmenet

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+    } else {
+      return Assertion(fact);
+    }
+  }
+
+  Assertion AddUniverseConditions(const ConditionalAssertion& cond_assertion) {
+    std::vector<BaseFact> conditions =
+        GetUniverseConditions(cond_assertion.lhs().base_fact());
+    absl::c_copy(cond_assertion.rhs(), std::back_inserter(conditions));
+    return Assertion(ConditionalAssertion(cond_assertion.lhs(), conditions));
```
markww, line 204: Might want to break `cond_assertion.lhs()` into a temporary, just to make things a bit more concise.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  // This method, given a BaseFact appearing in the LHS of an assertion,
+  // returns a vector of conditions that should be added to the RHS of that
+  // assertion. This method is where this class is actually used as a visitor.
+  // The visitor works only on subtrees of the AST that are
+  // these BaseFacts on the LHS and returns a set of BaseFacts with
+  // the universe conditions.
+  inline std::vector<BaseFact> GetUniverseConditions(
```
markww, line 213: I'm not sure if `inline` matters in a `.cc` file.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+      return {};
+    }
+  }
+
+  // Add universe conditions for the non-constant arguments
+  std::vector<BaseFact> Visit(const datalog::Predicate& pred) override {
+    datalog::RelationDeclaration decl =
+        decl_env_.GetDeclarationOrFatal(pred.name());
+    std::vector<BaseFact> conditions;
+    for (unsigned i = 0; i < pred.args().size(); i++) {
```
markww, line 239: Use `size_t` rather than `unsigned`, as on a 64 bit system, `unsigned` isn't wide enough to cover the range of potential indices in `pred.args().size()`.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+      auto arg = pred.args()[i];
```
markww, line 240: What is the type of `arg` here? Also, might want to break `pred.args()` into a local.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+  prog.Accept(rel_visitor);
+}
+
+datalog::RelationDeclaration DeclarationEnvironment::GetDeclarationOrFatal(
+    absl::string_view relation_name) {
+  if (inner_map_.find(relation_name) == inner_map_.end()) {
+    LOG(FATAL) << "could not find declaration for relation" << relation_name;
+  } else {
+    return inner_map_.find(relation_name)->second;
+  }
```
markww, line 265: Equivalently, but more concisely:

```C++
auto find_result = inner_map_.find(relation_name);
CHECK(find_result != inner_map_.end()) << "could not find declaration for relation " << relation_name;
return find_result->second;
```

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+    // than a numeric operator
+    datalog::RelationDeclaration decl =
+        decl_env_.GetDeclarationOrFatal(pred.name());
+    // The relation declarations give the types for each position in the
+    // predicate. For the xth argument in the predicate, we want to
+    // assign it the type for the xth parameter in the relation declaration.
+    // So to do this, we iterate through the list of arguments in the
+    // predicate and the list of parameters in the declaration using
+    // the indices of the arguments iterator of the predicate.
+    for (unsigned i = 0; i < pred.args().size(); i++) {
```
markww, line 321: `size_t` rather than `unsigned`.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,357 @@
...
+      GetUniverseDeclarations(prog);
+  std::vector<datalog::RelationDeclaration> new_declarations =
+      prog.relation_declarations();
+  new_declarations.insert(new_declarations.end(), universe_declarations.begin(),
+                          universe_declarations.end());
+
+  return Program(new_declarations, new_says_assertions, prog.queries());
+}
+
+}  // namespace raksha::ir::auth_logic
```
markww, line 357: Newline at eof

src/ir/auth_logic/universe_relation_insertion.h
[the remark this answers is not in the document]
markww: That's a reasonable and fine call to make, SGTM.

src/ir/auth_logic/universe_relation_insertion.h
aferr: speculating that this is resolved and resolving...

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
aferr: I think it's about the same to do it with the for loop and would rather leave it as is if that's ok?

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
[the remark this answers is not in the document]
aferr: On second thought, I think this is best left in the universe relation pass since at the time of writing the concept of a universe lives exclusively in this pass, so I think it is best to also keep this function about universes there? we can also always revise if universes become more of a first-class citizen for some reason. Feel free to re-open if you object, of course.

src/ir/auth_logic/universe_relation_insertion_test.cc
aferr: fixed by adding 2 more tests.

src/ir/auth_logic/ast.h
```diff
@@ -400,9 +400,9 @@ class Program {
       query_strings.push_back(query.DebugPrint());
     }
     return absl::StrCat("Program(\n",
-                        absl::StrJoin(relation_decl_strings, "\n"),
-                        absl::StrJoin(says_assertion_strings, "\n"),
-                        absl::StrJoin(query_strings, "\n"), ")");
+                        absl::StrJoin(relation_decl_strings, "\n"), "\n",
+                        absl::StrJoin(says_assertion_strings, "\n"), "\n",
+                        absl::StrJoin(query_strings, "\n"), "\n)");
```
markww, line 405: Rather than `StrCat`, you might want to just join these things with another `StrJoin` with a delimiter of `\n`.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+
+// Generate a new version of the program transformed to include
+// the universe declarations, universe populating rules,
+// and extensions of the rules to include conditions that check
+// these universes.
+Program InsertUniverseRelations(const Program& prog);
+
+};  // namespace raksha::ir::auth_logic
+
+#endif  // SRC_IR_AUTH_LOGIC_UNIVERSE_RELATION_INSERTION_H_
```
markww, line 139: Newline at EOF.

src/ir/auth_logic/universe_relation_insertion.h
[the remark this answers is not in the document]
markww: OK, that's reasonable.

src/ir/auth_logic/universe_relation_insertion.h
markww: Sure, that's fine.

src/ir/auth_logic/universe_relation_insertion.cc
markww: It's not about performance, it's more about preventing a paranoid reader from having to look up the definition to ensure that the function does not perform mutation if a bug pops up.

src/ir/auth_logic/ast.h

src/ir/auth_logic/universe_relation_insertion.h
aferr: fixed.

aferr:
@markww @bgogul re-review when you have a moment? :)

markww (approved):
LGTM

aferr:
@markww @bgogul Do you have any preference with how to sequence this PR and #700 ? As in, is it OK if I merge this one into main and then merge 700 into main shortly after? Or would you rather I merge 700 into this PR before merging this PR into main? (I slightly prefer the first way, but will do it however you want).

(I am also likely to stall on finishing 700 since I am busy with another deadline).

aferr:
@bgogul @markww This was just force-pushed because it was rebased on top of #700 so that [the dependence on DebutString/ToString can be removed](https://github.com/google-research/raksha/pull/654/commits/99d0ed732b46bc645ff295ba78478b4c09d2af16)

src/ir/datalog/program.h
```diff
@@ -149,11 +179,11 @@ class RelationDeclaration {
   absl::string_view relation_name() const { return relation_name_; }
   bool is_attribute() const { return is_attribute_; }
   const std::vector<Argument>& arguments() const { return arguments_; }
-  
+
   bool operator==(const RelationDeclaration& otherDeclaration) const {
     return this->relation_name_ == otherDeclaration.relation_name_ &&
-    this->is_attribute_ == otherDeclaration.is_attribute_ &&
-    this->arguments_ == otherDeclaration.arguments_;
+           this->is_attribute_ == otherDeclaration.is_attribute_ &&
```
aferr, line 185: this is just linting

src/ir/datalog/program.h
```diff
@@ -210,10 +245,10 @@ class Program {
   const std::vector<std::string>& outputs() const { return outputs_; }
 
   bool operator==(const Program& otherProgram) const {
-    return this->relation_declarations_ == 
-      otherProgram.relation_declarations_ &&
-      this->rules_ == otherProgram.rules_ &&
-      this->outputs_ == otherProgram.outputs_;
+    return this->relation_declarations_ ==
```
aferr, line 248: linting

src/ir/datalog/program.h
```diff
@@ -130,7 +155,12 @@ class Argument {
 
   bool operator==(const Argument& otherArgument) const {
     return this->argument_name_ == otherArgument.argument_name_ &&
-    this->argument_type_ == otherArgument.argument_type_;
```
aferr, line 133: linting

src/ir/datalog/program.h
```diff
@@ -101,12 +116,22 @@ class ArgumentType {
...
+    return H::combine(std::move(h), typ.name(), typ.kind());
+  }
+  // Equality is needed to use a RelationDeclaration in a flat_hash_set
   bool operator==(const ArgumentType& otherType) const {
     // The name field is only compared for arguments of kind kCustom
     // because the name field is only used for kCustom types (and
     // not for kPrincipal or kNumber).
-    return this->kind_ == otherType.kind_ &&
-      this->kind_ == Kind::kCustom ? this->name_ == otherType.name_ : true;
+    return this->kind_ == otherType.kind_ && this->kind_ == Kind::kCustom
```
aferr, line 128: linting

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/auth_logic/auth_logic_ast_traversing_visitor.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
+
+// A map from relation names to relation declarations.
+class DeclarationEnvironment {
+ public:
+  DeclarationEnvironment(const Program& prog);
```
bgogul, line 57: Do we have unit tests for `DeclarationEnvironment` class?

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+    }
+    DeclarationEnvironment& enclosing_env_;
+  };
+
+  absl::flat_hash_map<std::string, datalog::RelationDeclaration> inner_map_;
+};
+
+// A map from the names of principal / predicate argument literals
+// (the constants wrapped in quotes) to typings.
+class TypeEnvironment {
```
bgogul, line 92: Do we have unit tests for `TypeEnvironment` class?

bgogul (commented):
Andrew, I haven't been able to review the actual logic of `DeclarationEnvironment`, `TypeEnvironment` and `UniverseDeclarationGenerationVisitor`. 

I left some comments to refactor the code into separate files (not PRs) and some tests. Can you address them first? It will make reviewing the code somewhat easier.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,337 @@
...
+    // the indices of the arguments iterator of the predicate.
+    for (size_t i = 0; i < pred.args().size(); i++) {
+      enclosing_env_.AddTyping(pred.args()[i],
+                               decl.arguments()[i].argument_type());
+    }
+    return Unit();
+  }
+}
+
+Program InsertUniverseRelations(const Program& prog) {
```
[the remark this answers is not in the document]
aferr: I think keeping the phases of a compiler singular in purpose and self-contained is useful. Doing so makes the code easier to understand and test. As a result, I think it is best to have an AST that corresponds identically to the grammar, and for parsing to do nothing but convert the text into an AST that represents the input program exactly without any other transformations. Since this code is describing a program transformation, I think it should happen as a separate step after constructing the initial program (parsing).

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+*/
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/container/flat_hash_set.h"
+#include "absl/strings/str_cat.h"
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/auth_logic/auth_logic_ast_traversing_visitor.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::auth_logic {
```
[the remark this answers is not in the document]
aferr: Because the definitions of TypeEnvironment and DeclarationEnvironment are small, I put each one in a separate header file but did not split the implementations into cc files. Is that fine or are cc files needed even with classes this small?

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+
+// A map from relation names to relation declarations.
+class DeclarationEnvironment {
```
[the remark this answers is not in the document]
aferr: OK thanks for the example of how to split things into smaller PRs.

src/ir/datalog/program.h
```diff
@@ -101,12 +116,22 @@ class ArgumentType {
...
+    return H::combine(std::move(h), typ.name_, typ.kind_);
+  }
+  // Equality is needed to use a RelationDeclaration in a flat_hash_set
   bool operator==(const ArgumentType& otherType) const {
-    // The name field is only compared for arguments of kind kCustom
-    // because the name field is only used for kCustom types (and
-    // not for kPrincipal or kNumber).
-    return this->kind_ == otherType.kind_ &&
-      this->kind_ == Kind::kCustom ? this->name_ == otherType.name_ : true;
+    return this->kind_ == otherType.kind_ && this->name_ == otherType.name_;
```
aferr, line 130: I needed to fix this to change this to make hash tests pass. Now that there are factories for ArgumentTypes, the names of all the Number and Principal types will be the same, so the problem the old == operator was trying to fix is gone.

src/ir/datalog/program.h
```diff
@@ -101,12 +116,22 @@ class ArgumentType {
 
   std::string ToString() const { return absl::StrCat(kind_, name_); }
 
+  template <typename H>
```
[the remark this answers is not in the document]
aferr: fixed.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  absl::flat_hash_map<std::string, datalog::ArgumentType> inner_map_;
+};
+
+// Generate a new version of the program transformed to include
+// the universe declarations, universe populating rules,
+// and extensions of the rules to include conditions that check
+// these universes.
+Program InsertUniverseRelations(const Program& prog);
```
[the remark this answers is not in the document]
aferr: This should be called just after the AST is generated from parsing and before the lowing_ast_to_datalog pass. This isn't called yet because that should probably come with integration tests and be a separate PR.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+
+namespace raksha::ir::auth_logic {
+
+// A map from relation names to relation declarations.
+class DeclarationEnvironment {
+ public:
+  DeclarationEnvironment(const Program& prog);
+
+  void AddDeclaration(const datalog::RelationDeclaration& rel_decl) {
+    inner_map_.insert({std::string{rel_decl.relation_name()}, rel_decl});
```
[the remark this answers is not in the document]
aferr: At the time that you left this comment, the second declaration would have had no effect. This is probably not the right way to handle it.

Now it throws an error if two relations with the same name are found. We can consider handling it differently though. For example, we can permit multiple declarations of the same relation if they are identical, but not if there are different relations with the same name. I went with the simplest answer first.

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+  absl::flat_hash_map<std::string, datalog::RelationDeclaration> inner_map_;
+};
+
+// A map from the names of principal / predicate argument literals
+// (the constants wrapped in quotes) to typings.
+class TypeEnvironment {
+ public:
+  TypeEnvironment(DeclarationEnvironment decl_env, const Program& prog);
+
+  datalog::ArgumentType GetTypingOrFatal(absl::string_view argument_name);
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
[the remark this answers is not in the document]
aferr: Actually there were implementations of the environments in `universe_relation_insertion.cc` that I forgot about, so I moved those into separate .cc files for the environments

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+    RelationDeclarationEnvironmentVisitor(DeclarationEnvironment& enclosing_env)
+        : enclosing_env_(enclosing_env) {}
+
+   private:
+    Unit Visit(const datalog::RelationDeclaration& rel_decl) {
+      enclosing_env_.AddDeclaration(rel_decl);
+      return Unit();
+    }
+    DeclarationEnvironment& enclosing_env_;
+  };
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+
+   private:
+    Unit PreVisit(const Principal& principal);
+    Unit Visit(const datalog::Predicate& pred);
+    DeclarationEnvironment decl_env_;
+    TypeEnvironment& enclosing_env_;
+  };
+
+  absl::flat_hash_map<std::string, datalog::ArgumentType> inner_map_;
+};
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.h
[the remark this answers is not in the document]
aferr: This is also now checked in the test

src/ir/auth_logic/universe_relation_insertion.h
aferr: now there are

src/ir/auth_logic/universe_relation_insertion.h
aferr: now there are

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,139 @@
...
+class DeclarationEnvironment {
+ public:
+  DeclarationEnvironment(const Program& prog);
+
+  void AddDeclaration(const datalog::RelationDeclaration& rel_decl) {
+    inner_map_.insert({std::string{rel_decl.relation_name()}, rel_decl});
+  }
+
+  datalog::RelationDeclaration GetDeclarationOrFatal(
+      absl::string_view relation_name);
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/declaration_environment_test.cc
```diff
@@ -0,0 +1,47 @@
...
+  DeclarationEnvironment decl_env(prog);
+  EXPECT_DEATH(decl_env.GetDeclarationOrFatal("relation_not_here"), "");
+}
+
+TEST(DeclarationEnvironmentTests, FatalAddTest) {
+  datalog::Argument arg1("arg1", datalog::ArgumentType::MakePrincipalType());
+  datalog::RelationDeclaration decl1("relname", false, {arg1});
+  Program prog({decl1}, {}, {});
+  DeclarationEnvironment decl_env(prog);
+  EXPECT_DEATH(decl_env.AddDeclaration(decl1), "");
```
bgogul, line 44: It might be better to make `AddDeclaration` private? If so, we can change the death test as follows:
```suggestion
  datalog::RelationDeclaration decl1("relname", false, {arg1});
  datalog::RelationDeclaration decl2("relname", false, {arg1});
  Program prog({decl1, decl2}, {}, {});
  DeclarationEnvironment decl_env(prog);
  EXPECT_DEATH(decl_env.AddDeclaration(decl1), "whatever error message");
```

src/ir/auth_logic/declaration_environment.h
```diff
@@ -0,0 +1,45 @@
...
+namespace raksha::ir::auth_logic {
+
+// A map from relation names to relation declarations.
+class DeclarationEnvironment {
+ public:
+  DeclarationEnvironment(const Program& prog);
+
+  // Adds a relation to the environment if there is not already
+  // a relation with that name in the environment, or errors otherwise.
+  void AddDeclaration(const datalog::RelationDeclaration& rel_decl);
```
bgogul, line 34: Move this API to the visitor class?

src/ir/auth_logic/declaration_environment.cc
```diff
@@ -0,0 +1,68 @@
...
+ public:
+  RelationDeclarationEnvironmentVisitor(DeclarationEnvironment& enclosing_env)
+      : enclosing_env_(enclosing_env) {}
+
+ private:
+  Unit Visit(const datalog::RelationDeclaration& rel_decl) {
+    enclosing_env_.AddDeclaration(rel_decl);
+    return Unit();
+  }
+  DeclarationEnvironment& enclosing_env_;
```
bgogul, line 39: Why not construct `absl::flat_hash_map<std::string, datalog::RelationDeclaration> inner_map_;` in the visitor and return it?

src/ir/auth_logic/is_name_constant.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef SRC_IR_AUTH_LOGIC_IS_NAME_CONSTANT_H_
+#define SRC_IR_AUTH_LOGIC_IS_NAME_CONSTANT_H_
+
+// Determine if an identifier (which may be an argument of a predicate
+// or a principal name) is a constant or variable.
+// TODO(#673 aferr) it would be better to refactor the AST so that
+// identifiers are a separate AST node with constant and non-constant
+// children determined at parse time rather than here since
+// this information is available at parse time.
+bool IsNameConstant(absl::string_view id) {
```
bgogul, line 29: If this is only used in type_environment.cc, simply move it there? 

Otherwise, if you need keep this in the header file, you should add some tests.

src/ir/auth_logic/type_environment.h
```diff
@@ -0,0 +1,57 @@
...
+
+  // Insert a typing for constant value (e.g. "SomeArgument" in quotes
+  // applied to a relation, or a constant principal in any context where
+  // principals may be applied.
+  // - If no typing for this constant already exists, the typing is added
+  // - If a typing for this constant exists but it is the same type as the new
+  // one, nothing happens
+  // - If a typing for this constant exists, and it is a different type from
+  // the new one, an error is thrown.
+  void AddTyping(absl::string_view arg_name, datalog::ArgumentType arg_type);
```
bgogul, line 49: Just as with `DecalrationEnvironment`, move this method to the visitor.

src/ir/auth_logic/type_environment.h
```diff
@@ -0,0 +1,57 @@
...
+
+// A map from the names of principal / predicate argument literals
+// (the constants wrapped in quotes) to typings.
+class TypeEnvironment {
+ public:
+  TypeEnvironment(DeclarationEnvironment decl_env, const Program& prog);
+
+  datalog::ArgumentType GetTypingOrFatal(absl::string_view argument_name);
+
+  const absl::flat_hash_map<std::string, datalog::ArgumentType>& inner_map()
```
bgogul, line 36: more meaningful name than `inner_map`?

src/ir/auth_logic/type_environment.cc
```diff
@@ -0,0 +1,108 @@
...
+ public:
+  TypeEnvironmentGenerationVisitor(DeclarationEnvironment decl_env,
+                                   TypeEnvironment& enclosing_env)
+      : decl_env_(decl_env), enclosing_env_(enclosing_env) {}
+
+ private:
+  Unit PreVisit(const Principal& principal);
+  Unit Visit(const datalog::Predicate& pred);
+  DeclarationEnvironment decl_env_;
+  TypeEnvironment& enclosing_env_;
```
bgogul, line 44: Instead of `TypeEnvironment`, keep around   `absl::flat_hash_map<std::string, datalog::ArgumentType> inner_map_;` and return it after the visit?

src/ir/datalog/program.h
```diff
@@ -66,6 +67,20 @@ class Predicate {
...
 
+  // (TODO: #672) once an AST node for numbers as RVALUEs in numeric
+  // comparisons is added, this visitor should also visit
+  // the numeric RVALUEs and add them to the type environment
+  // with type ArgumentType(kNumber, "Number")
+  // This is a workaround that makes use of the fact that
+  // numeric comparisons are represented as predicates
+  // with a name that matches the operator:
+  // (https://github.com/google-research/raksha/blob/be6ef8e1e1a20735a06637c12db9ed0b87e3d2a2/src/ir/auth_logic/ast_construction.cc#L92)
+  bool IsNumericOperator() const {
```
bgogul, line 78: No tests for this?

src/ir/datalog/program.h
```diff
@@ -66,6 +67,20 @@ class Predicate {
...
+    const absl::flat_hash_set<std::string> numeric_operators = {
+        "<", ">", "=", "!=", "<=", ">="};
+    return numeric_operators.contains(name_);
```
bgogul, line 81: Make it static. You will need to use a pointer to avoid data races during destruction:
```suggestion
    static const auto* const numeric_operators = new absl::flat_hash_set<std::string>({
        "<", ">", "=", "!=", "<=", ">="});
    return numeric_operators->contains(name_);
```

src/ir/datalog/program.h
bgogul: Ideally a separate CL in future.

src/ir/datalog/program.h
```diff
@@ -101,12 +116,22 @@ class ArgumentType {
...
+  // Equality is needed to use a RelationDeclaration in a flat_hash_set
   bool operator==(const ArgumentType& otherType) const {
-    // The name field is only compared for arguments of kind kCustom
-    // because the name field is only used for kCustom types (and
-    // not for kPrincipal or kNumber).
-    return this->kind_ == otherType.kind_ &&
-      this->kind_ == Kind::kCustom ? this->name_ == otherType.name_ : true;
+    return this->kind_ == otherType.kind_ && this->name_ == otherType.name_;
+  }
+  // This one is just for convenience
```
bgogul, line 132: The comment doesn't seem right. This is still useful?

src/ir/auth_logic/declaration_environment_test.cc
```diff
@@ -0,0 +1,47 @@
...
+  Program prog({decl1, decl2}, {}, {});
+  DeclarationEnvironment decl_env(prog);
+  EXPECT_EQ(decl_env.GetDeclarationOrFatal("relname"), decl1);
+  EXPECT_EQ(decl_env.GetDeclarationOrFatal("relname2"), decl2);
+}
+
+TEST(DeclarationEnvironmentTests, FatalLookupTest) {
+  Program prog({}, {}, {});
+  DeclarationEnvironment decl_env(prog);
+  EXPECT_DEATH(decl_env.GetDeclarationOrFatal("relation_not_here"), "");
```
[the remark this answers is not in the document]
aferr: I think the error goes to stderr or a log rather than stdout, so the expected value for `EXPECT_DEATH` is `""` it seems?

src/ir/auth_logic/declaration_environment_test.cc
aferr: The reason `AddDeclaration` is not private is so it can be called by the `RelationDeclarationEnvironmentVisitor` in `declaration_environment.cc`.

src/ir/auth_logic/declaration_environment.cc
aferr: That seems also fine, but then we need to either make the `inner_map_` of the `DeclarationEnvironment` public or give it a public setter method. This seems fine but I don't necessarily see why that is any better or worse than the way it works now.

src/ir/auth_logic/is_name_constant.h
aferr: It's also used in `universe_relation_insertion.cc`, so I needed to put it somewhere that it's accessible for both. It's both an implementation detail I'd rather hide and something I would really rather get rid of eventually. Is there another way to make this available to both `type_environment.cc` and `universe_relation_insertion.cc` and not put it in a header? I could duplicate it in both cc files but that seems worse.

src/ir/auth_logic/type_environment.h
aferr: This is currently public for the same reason as with the `DeclarationEnvironment`, so that it can be called by the visitor. Similarly the visitor could return the data structure it builds, but then this requires either making the data structure in `TypeEnvironment` public or giving it a public setter method. Both the way this is currently implemented and making a public setter seem equally fine to me (and leaving it as is requires no additional work). Happy to do it either way though if you have a preference!

src/ir/auth_logic/type_environment.h
aferr: fixed

src/ir/datalog/program.h
aferr: fixed!

src/ir/datalog/program.h
aferr: There is a comment for `==` to explain why it is needed, but `!=` isn't *strictly* needed for anything. It is just for convenience given that there is already a `==` operator. Since this comment is more confusing that helpful, I'll delete it.

src/ir/datalog/program.h
aferr: fixed

src/ir/auth_logic/declaration_environment_test.cc
aferr: On second thought, I see what you mean this is fixed.

src/ir/auth_logic/declaration_environment.cc
aferr: On second thought, fixed this.

src/ir/auth_logic/declaration_environment.h
aferr: fixed

src/ir/auth_logic/is_name_constant.h
aferr: added the tests

aferr:
@bgogul this one is ready for re-review btw

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -213,27 +213,25 @@ TEST(AuthLogicAstTraversingVisitorTest, SimpleTraversalTest) {
...
-          program1.ToString(), saysAssertion1.ToString(),
-          prinA.ToString(), assertion1.ToString(), fact1.ToString(),
-          baseFact1.ToString(), saysAssertion2.ToString(),
-          prinC.ToString(), assertion1.ToString(), fact1.ToString(),
-          baseFact1.ToString(), assertion2.ToString(), fact2.ToString(),
-          prinB.ToString(), baseFact2.ToString(), query1.ToString(),
-          prinA.ToString(), fact1.ToString(), baseFact1.ToString(),
-          query2.ToString(), prinB.ToString(), fact2.ToString(),
-          prinB.ToString(), baseFact2.ToString()));
+          program1.ToString(), saysAssertion1.ToString(), prinA.ToString(),
```
bgogul, line 216: You might want to sync to HEAD.

src/ir/auth_logic/is_name_constant.h
```diff
@@ -0,0 +1,44 @@
...
+namespace raksha::ir::auth_logic {
+
+// Determine if an identifier (which may be an argument of a predicate
+// or a principal name) is a constant or variable.
+// TODO(#673 aferr) it would be better to refactor the AST so that
+// identifiers are a separate AST node with constant and non-constant
+// children determined at parse time rather than here since
+// this information is available at parse time.
+bool IsNameConstant(absl::string_view id) {
+  if (id[0] == '"' && id[id.size()-1] == '"') {
```
bgogul, line 32: You can use `absl::StartsWith` and `absl::EndsWith`  to make this more readable.

src/ir/auth_logic/is_name_constant_test.cc
```diff
@@ -0,0 +1,29 @@
...
+
+#include "src/ir/auth_logic/is_name_constant.h"
+
+#include "src/common/testing/gtest.h"
+
+namespace raksha::ir::auth_logic {
+
+TEST(IsNameConstantTest, BasicTest) {
+  EXPECT_EQ(IsNameConstant("\"hello\""), true);
+  EXPECT_EQ(IsNameConstant("\"hello"), false);
```
bgogul, line 25: Add an test for numbers?

src/ir/auth_logic/type_environment.cc
```diff
@@ -0,0 +1,134 @@
...
+// This visitor aids in the construction of the TypeEnvironment.
+// It populates a mapping (flat_hash_map) of constant (literal)
+// names to their types by looking at the relations to which the literals
+// are applied and declarations of those relations (including the types
+// of the parameters where they are applied). It takes as input
+// a DeclarationEnvironment.
+// This is a nested class because it needs to access the private
+// fields of the TypeEnvironment in the constructor implementation.
+
+using LiteralTypeMapType =
```
bgogul, line 34: Not a huge problem as this is in an anonymous namespace, but would be good to move it inside the class?

src/ir/auth_logic/type_environment.cc
```diff
@@ -0,0 +1,134 @@
...
+    CHECK(find_result->second == arg_type)
+        << "Type error for constant: " << arg_name;
+  }
+
+  /*
+   auto insert_result =
+       literal_type_map_.insert({std::string(arg_name), arg_type});
+   CHECK(insert_result.second)
+       << "Type error for constant: " << insert_result.first->first;
+   */
```
bgogul, line 75: stale comment.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+    case datalog::ArgumentType::Kind::kNumber:
+      return "isNumber";
+    case datalog::ArgumentType::Kind::kPrincipal:
+      return "isPrincipal";
+    case datalog::ArgumentType::Kind::kCustom:
+      return absl::StrCat("is", arg_type.name());
+  }
+}
+
+BaseFact MakeUniverseMembershipFact(std::string element_name,
```
bgogul, line 36: `absl::string_view element_name`

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+
+#include "src/ir/auth_logic/universe_relation_insertion.h"
+
+#include "src/ir/auth_logic/is_name_constant.h"
+
+namespace raksha::ir::auth_logic {
+
+namespace {
+
+std::string TypeToUniverseName(datalog::ArgumentType arg_type) {
```
bgogul, line 25: `const datalog::ArgumentType arg_type&` (We are not making a copy of arg_type here.)

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+      return "isNumber";
+    case datalog::ArgumentType::Kind::kPrincipal:
+      return "isPrincipal";
+    case datalog::ArgumentType::Kind::kCustom:
+      return absl::StrCat("is", arg_type.name());
+  }
+}
+
+BaseFact MakeUniverseMembershipFact(std::string element_name,
+                                    datalog::ArgumentType arg_type) {
```
bgogul, line 37: `const datalog::ArgumentType& arg_type` as we are not making a copy of `arg_type` here.

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+  return BaseFact(
+      datalog::Predicate(TypeToUniverseName(arg_type),  // universe relation
+                         {element_name},  // the element to add to universe
+                         datalog::Sign::kPositive));
+}
+
+datalog::RelationDeclaration TypeToUniverseDeclaration(
+    datalog::ArgumentType arg_type) {
```
bgogul, line 45: `const datalog::ArgumentType& arg_type`

src/ir/auth_logic/universe_relation_insertion.h
```diff
@@ -0,0 +1,64 @@
...
+`ast.h`. It should come before the translation in `lowering_ast_datalog.h`.
+*/
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/container/flat_hash_set.h"
+#include "absl/strings/str_cat.h"
+#include "src/ir/auth_logic/ast.h"
+#include "src/ir/auth_logic/auth_logic_ast_traversing_visitor.h"
+#include "src/ir/auth_logic/declaration_environment.h"
+#include "src/ir/auth_logic/type_environment.h"
```
bgogul, line 51: These headers are not needed in the header file. Move it to the .cc file?

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+  // referenced in relation declarations in the front-end
+  absl::flat_hash_set<datalog::RelationDeclaration> builtin_universes = {
+      TypeToUniverseDeclaration(datalog::ArgumentType(
+          datalog::ArgumentType::Kind::kNumber, "Number")),
+      TypeToUniverseDeclaration(datalog::ArgumentType(
+          datalog::ArgumentType::Kind::kPrincipal, "Principal")),
+  };
+  ret.merge(builtin_universes);
+  return ret;
+}
```
bgogul, line 116: Move this to the visitor as a static method?

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+// This class transforms SaysAssertions so that if
+// there are any potentially ungrounded variables on the LHS of an assertion,
+// conditions are added to the RHS that check that this variable
+// is in a universe. As a way of safely overapproximating the ungrounded
+// variables, this just assumes any *non-constant* arguments in the RHS
+// are ungrounded, and adds conditions for them to the LHS.
+// It implements just one public method,
+// `AddUniverseConditions(SaysAssertion says_assertion)`
+// that returns a transformed version of the says assertion.
+// Internally, it extends a visitor for a part of this translation
```
bgogul, line 130: nit: Format properly?

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+          datalog::ArgumentType::Kind::kPrincipal, "Principal")),
+  };
+  ret.merge(builtin_universes);
+  return ret;
+}
+
+//----------------------------------------------------------------------------
+// Grounding Condition Transformation
+//----------------------------------------------------------------------------
+// This class transforms SaysAssertions so that if
```
bgogul, line 121: Add an example transformation in the comment?

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  // Add universe conditions for the non-constant arguments
+  std::vector<BaseFact> Visit(const datalog::Predicate& pred) override {
+    datalog::RelationDeclaration decl =
+        decl_env_.GetDeclarationOrFatal(pred.name());
+    std::vector<BaseFact> conditions;
+    std::vector<std::string> args = pred.args();
+    for (size_t i = 0; i < args.size(); i++) {
+      std::string arg = args[i];
```
bgogul, line 207: `absl::string_view arg = args[i];`

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -0,0 +1,249 @@
...
+// there are any potentially ungrounded variables on the LHS of an assertion,
+// conditions are added to the RHS that check that this variable
+// is in a universe. As a way of safely overapproximating the ungrounded
+// variables, this just assumes any *non-constant* arguments in the RHS
+// are ungrounded, and adds conditions for them to the LHS.
+// It implements just one public method,
+// `AddUniverseConditions(SaysAssertion says_assertion)`
+// that returns a transformed version of the says assertion.
+// Internally, it extends a visitor for a part of this translation
+class UniverseConditionTransformer
```
bgogul, line 131: nit: a more appropriate name? e.g., `SaysAssertionTransformer`

src/ir/auth_logic/universe_relation_insertion.cc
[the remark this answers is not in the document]
bgogul: OK. Let us leave this for now, but I think this can be cleaned up a bit.

bgogul (approved):
Looks good modulo some suggestions.

I think there is possibility of cleaning up insert_universe_relation, but let us do it in a follow up.

src/ir/auth_logic/type_environment.cc
aferr: fixed

src/ir/auth_logic/type_environment.cc
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed. included one to show overconservativeness

src/ir/auth_logic/universe_relation_insertion.cc
aferr: Fixed I thin? Ran the linter on it anyway.

src/ir/auth_logic/universe_relation_insertion.cc
aferr: OK I called it `GroundingConditionTransformer`. ?

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed

PR #669, opened 2022-08-19
Add --proto to check_policy_compliance

cromwellian:
Add --proto to check_policy_compliance
Adds ProtoToIR::Convert(context, textproto)
    
    Allows running a policy check against a text proto
    instead of parsing IR.

### DIFF
--- a/src/backends/policy_engine/souffle/check_policy_compliance.cc
+++ b/src/backends/policy_engine/souffle/check_policy_compliance.cc
@@ -29,12 +29,14 @@
 #include "src/backends/policy_engine/souffle/souffle_policy_checker.h"
 #include "src/backends/policy_engine/sql_policy_rule_policy.h"
 #include "src/common/logging/logging.h"
+#include "src/ir/proto_to_ir.h"
 #include "src/parser/ir/ir_parser.h"
 
-ABSL_FLAG(std::string, ir, "", "the IR file");
-ABSL_FLAG(std::optional<std::string>, sql_policy_rules, "",
+ABSL_FLAG(std::optional<std::string>, ir, std::nullopt, "the IR file");
+ABSL_FLAG(std::optional<std::string>, sql_policy_rules, std::nullopt,
           "file containing the SQL policy rules");
-ABSL_FLAG(std::optional<std::string>, policy_engine, "",
+ABSL_FLAG(std::optional<std::string>, proto, std::nullopt, "the proto file");
+ABSL_FLAG(std::optional<std::string>, policy_engine, std::nullopt,
           "name of the policy engine");
 
 constexpr char kUsageMessage[] =
@@ -61,6 +63,52 @@ absl::StatusOr<std::string> ReadFileContents(std::filesystem::path file_path) {
   return string_stream.str();
 }
 
+enum class ReturnCode : uint8_t { PASS = 0, FAIL = 1 , ERROR = 2};
+
+int UnwrapExitCode(ReturnCode code) { return static_cast<int>(code); }
+
+struct IrGraphComponents {
+  std::unique_ptr<raksha::ir::Module> ir_module;
+  std::unique_ptr<raksha::ir::SsaNames> ir_ssa_names;
+  std::unique_ptr<raksha::ir::IRContext> ir_context;
+};
+
+absl::StatusOr<IrGraphComponents> GetIrGraphComponentsFromIrPath(
+    absl::string_view ir_path) {
+  raksha::ir::IrProgramParser ir_parser;
+  absl::StatusOr<std::string> ir_string = ReadFileContents(ir_path);
+  if (ir_string.ok()) {
+    auto [context, module, ssa_names] = ir_parser.ParseProgram(*ir_string);
+    return IrGraphComponents{.ir_module = std::move(module),
+                             .ir_ssa_names = std::move(ssa_names),
+                             .ir_context = std::move(context)};
+  } else {
+    return ir_string.status();
+  }
+}
+
+absl::StatusOr<IrGraphComponents> GetIrGraphComponentsFromProtoPath(
+    absl::string_view proto_path) {
+  absl::StatusOr<std::string> proto_string = ReadFileContents(proto_path);
+  if (proto_string.ok()) {
+    std::unique_ptr<raksha::ir::IRContext> context =
+        std::make_unique<raksha::ir::IRContext>(); /*...*/
+    absl::StatusOr<raksha::ir::ProtoToIR::Result> ir_result =
+        raksha::ir::ProtoToIR::Convert(*context, *proto_string);
+    if (ir_result.ok()) {
+      return IrGraphComponents{
+          .ir_module = std::move(ir_result.value().module),
+          .ir_ssa_names = std::move(ir_result.value().ssa_names),
+          .ir_context = std::move(context)};
+    } else {
+      return ir_result.status();
+    }
+  } else {
+    LOG(ERROR) << "Error reading proto file: " << proto_string.status();
+    return proto_string.status();
+  }
+}
+
 }  // namespace
 
 using raksha::backends::policy_engine::AuthLogicPolicy;
@@ -72,36 +120,48 @@ int main(int argc, char* argv[]) {
   google::InitGoogleLogging("check_policy_compliance");
   absl::SetProgramUsageMessage(kUsageMessage);
   absl::ParseCommandLine(argc, argv);
-  // Parse the given IR file.
-  absl::StatusOr<std::string> ir_string =
-      ReadFileContents(absl::GetFlag(FLAGS_ir));
-  if (!ir_string.ok()) {
-    LOG(ERROR) << "Error reading IR file: " << ir_string.status();
-    return 2;
-  }
 
   // Read the sql policy rules file.
   absl::StatusOr<std::string> sql_policy_rules =
       ReadFileContents(absl::GetFlag(FLAGS_sql_policy_rules).value());
-  if (absl::GetFlag(FLAGS_sql_policy_rules).has_value() &&
-      !sql_policy_rules.ok()) {
+  if (absl::GetFlag(FLAGS_sql_policy_rules).has_value() && !sql_policy_rules.ok()) {
     LOG(ERROR) << "Error reading sql policy rules file: "
                << sql_policy_rules.status();
-    return 2;
+    return UnwrapExitCode(ReturnCode::ERROR);
+  }
+
+  const std::optional<std::string> ir_path = absl::GetFlag(FLAGS_ir);
+  const std::optional<std::string> proto_path = absl::GetFlag(FLAGS_proto);
+
+  if (ir_path.has_value() == proto_path.has_value()) {
+    if (ir_path.has_value()) {
+      LOG(ERROR) << "Both --ir and --proto cannot be used simultaneously.";
+    } else {
+      LOG(ERROR) << "One of --ir or --proto must be specified.";
+    }
+
+    return UnwrapExitCode(ReturnCode::ERROR);
+  }
+
+  // Parse either an IR file or a Proto file
+  const absl::StatusOr<IrGraphComponents> components =
+      ir_path.has_value() ? GetIrGraphComponentsFromIrPath(*ir_path)
+                          : GetIrGraphComponentsFromProtoPath(*proto_path);
+  if (!components.ok()) {
+    LOG(ERROR) << "Error during parsing graph " << components.status();
+    return UnwrapExitCode(ReturnCode::ERROR);
   }
 
   // Invoke policy checker and return result.
-  IrProgramParser irParser;
   bool policyCheckSucceeded = false;
-  auto [context, module, ssa_names] = irParser.ParseProgram(*ir_string);
   if (absl::GetFlag(FLAGS_policy_engine).has_value()) {
     AuthLogicPolicy policy(absl::GetFlag(FLAGS_policy_engine).value());
     policyCheckSucceeded =
-        SoufflePolicyChecker().IsModulePolicyCompliant(*module, policy);
+        SoufflePolicyChecker().IsModulePolicyCompliant(*components.value().ir_module, policy);
   } else if (absl::GetFlag(FLAGS_sql_policy_rules).has_value()) {
     SqlPolicyRulePolicy policy(*sql_policy_rules);
     policyCheckSucceeded =
-        SoufflePolicyChecker().IsModulePolicyCompliant(*module, policy);
+        SoufflePolicyChecker().IsModulePolicyCompliant(*components.value().ir_module, policy);
   } else {
     LOG(ERROR) << "Both sql_policy_rules and authorzation logic generated "
                   "datalog policy engine are undefined";
@@ -110,9 +170,9 @@ int main(int argc, char* argv[]) {
 
   if (policyCheckSucceeded) {
     LOG(ERROR) << "Policy check succeeded!";
-    return 0;
+    return UnwrapExitCode(ReturnCode::PASS);
   } else {
     LOG(ERROR) << "Policy check failed!";
-    return 1;
+    return UnwrapExitCode(ReturnCode::FAIL);
   }
 }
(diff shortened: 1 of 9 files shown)


src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -69,29 +72,60 @@ int main(int argc, char* argv[]) {
...
+      if (!ir_string.ok()) {
+          LOG(ERROR) << "Error reading IR file: " << ir_string.status();
+          return 2;
+      }
+
+      IrProgramParser irParser;
+      auto [context, module, ssa_names] = irParser.ParseProgram(*ir_string);
+      ir_module = std::move(module);
+      ir_ssa_names = std::move(ssa_names);
+      ir_context = std::move(context);
```
cromwellian, line 109: I'm making these moves here because if I don't move ssa_names and context, it'll crash with bad_alloc, even though they aren't used. My guess is when the if{} scope disappears, the local unique_ptr destructs, but there's still some way it's being referenced from module? my C++ fu is too stale.

src/ir/proto_to_ir.h
```diff
@@ -33,6 +34,20 @@ class ProtoToIR {
...
   };
+
+  static Result Convert(IRContext& context,
+                        const std::string irTranslationUnitJsonProto) {
+      proto::IrTranslationUnit irTranslationUnit;
+      auto status = google::protobuf::util::JsonStringToMessage(irTranslationUnitJsonProto,
+                                                                &irTranslationUnit);
+      if (!status.ok()) {
+          LOG(ERROR) << "Json IR proto is corrupt " << status;
+          // TODO(cromwellian): handle this
```
cromwellian, line 45: J, maybe I return optional<Result> or add a status field to Result

WORKSPACE
```diff
@@ -143,9 +143,9 @@ http_archive(
 
 http_archive(
     name = "absl",
-    sha256 = "dcf71b9cba8dc0ca9940c4b316a0c796be8fab42b070bb6b7cab62b48f0e66c4",
-    strip_prefix = "abseil-cpp-20211102.0",
-    urls = ["https://github.com/abseil/abseil-cpp/archive/refs/tags/20211102.0.tar.gz"],
+    sha256 = "4208129b49006089ba1d6710845a45e31c59b0ab6bff9e5788a87f55c5abd602",
```
cromwellian, line 146: necessary for ABSL_FLAG with std::optional

cromwellian (commented):
TODO: tests
TODO: fix TODOs

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -27,11 +27,14 @@
 #include "src/backends/policy_engine/policy.h"
 #include "src/backends/policy_engine/souffle/souffle_policy_checker.h"
 #include "src/common/logging/logging.h"
+#include "src/ir/proto_to_ir.h"
 #include "src/parser/ir/ir_parser.h"
 
-ABSL_FLAG(std::string, ir, "", "the IR file");
+// TODO(cromwellian): why doesn't std::optional<std::string> (or absl::optional) work?
```
Cypher1, line 33: What error do you have when using `std::optional<std::string>`?

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -69,29 +72,60 @@ int main(int argc, char* argv[]) {
...
+  // Read the sql policy rules file.
+  absl::StatusOr<std::string> sql_policy_rules =
+          ReadFileContents(absl::GetFlag(FLAGS_sql_policy_rules));
+  if (!sql_policy_rules.ok()) {
+      LOG(ERROR) << "Error reading sql policy rules file: "
+                 << sql_policy_rules.status();
+      return 2;
+  }
+
+  // TODO(cromwellian): these flags are mutually exclusive, add a check
```
Cypher1, line 84: Missing check?
Maybe something like
```
CHECK(absl::GetFlag(FLAGS_ir).has_value() != absl::GetFlag(FLAGS_proto).has_value()) << "Error message";
```

src/backends/policy_engine/souffle/check_policy_compliance.cc
cromwellian: I fixed it, see the discussion in the Raksha chat. This was caused by an old version ABSL, I upgraded our workspace to pickup the newest version that supports optional flags. 

(The error you'll get is the same you'll get for any custom type: Can't find AbslUnparseFlags(TYPE). This is because marshalling.h didn't provide an unparser for std::optional out of the box in older versions)

src/backends/policy_engine/souffle/check_policy_compliance.cc
Cypher1: Ah actually just a stale comment

src/backends/policy_engine/souffle/check_policy_compliance.cc
Cypher1: Right, if you don't move them, the ownership is still 'in' the if, so they get cleaned up. Tbh we shouldn't be able to assign to a unique_ptr without a move... but it seems C++ is not as strict as it should be here (*cough* unlike rust *cough*).

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -69,29 +72,60 @@ int main(int argc, char* argv[]) {
...
+      ir_ssa_names = std::move(ssa_names);
   }
 
   // Invoke policy checker and return result.
   SoufflePolicyChecker checker;
-  IrProgramParser irParser;
-  auto [context, module, ssa_names] = irParser.ParseProgram(*ir_string);
   Policy policy(*sql_policy_rules);
-  if (checker.IsModulePolicyCompliant(*module, policy)) {
+
```
Cypher1, line 127: Probably should check that the `ir_*` `unique_ptrs` are all initialized here to avoid a potential seg fault with no error message.

src/backends/policy_engine/souffle/check_policy_compliance.cc
cromwellian: Yeah, I almost believed the compiler's "auto [x,y,z]" destructuring assignment would do it for us automagically. :(

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -27,11 +27,14 @@
...
 #include "src/common/logging/logging.h"
+#include "src/ir/proto_to_ir.h"
 #include "src/parser/ir/ir_parser.h"
 
-ABSL_FLAG(std::string, ir, "", "the IR file");
+// TODO(cromwellian): why doesn't std::optional<std::string> (or absl::optional) work?
+ABSL_FLAG(std::optional<std::string>, ir, std::nullopt, "the IR file");
 ABSL_FLAG(std::string, sql_policy_rules, "",
           "file containing the SQL policy rules");
+ABSL_FLAG(std::optional<std::string>, proto, std::nullopt, "the proto file");
```
Cypher1, line 37: I think this would be better as a command (i.e. always take 1 file and then 0 or 1 commands either 'proto' or 'ir').

If I understand correctrly, currently your flags will accept a null `FLAG_ir` and a null `FLAG_proto` and then segfault as the `ir_*` unique_ptrs are not initialized.

WORKSPACE
markww: lol, see #674 .

src/backends/policy_engine/souffle/check_policy_compliance.cc
markww: SG, please remove the TODOs.

src/ir/proto_to_ir.h
markww: Returning an `absl::StatusOr<Result>` is probably the best way forward.

src/backends/policy_engine/souffle/check_policy_compliance.cc
cromwellian: I'll fix the segfault, but I think we should discuss how to arrange the CLI. I kind of feel that the ideal thing would be just to detect the file type (proto or IR format) and work no matter what.  Either that, or build two different CLI binaries, one that handles each format. For now, just leave as two flags (with fixes to remove foot guns)

src/ir/proto_to_ir.h
```diff
@@ -16,6 +16,7 @@
 #ifndef SRC_IR_PROTO_TO_IR_H_
 #define SRC_IR_PROTO_TO_IR_H_
 
+#include <google/protobuf/util/json_util.h>
```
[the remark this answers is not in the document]
cromwellian: Done (this is an artifact of CLion's auto-insertion of #includes BTW)

src/ir/proto_to_ir.h
```diff
@@ -33,6 +34,20 @@ class ProtoToIR {
     std::unique_ptr<Module> module;
     std::unique_ptr<SsaNames> ssa_names;
   };
+
+  static Result Convert(IRContext& context,
+                        const std::string irTranslationUnitJsonProto) {
```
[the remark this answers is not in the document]
cromwellian: Thanks for the tip. My habit in other languages is towards immutability for everything, since mutability has usually been a source of bugs, I didn't even realize this was inhibiting compiler opts.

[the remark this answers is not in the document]
markww: Very sympathetic to that, I used to make all of my C++ locals `const`. Unfortunately, when C++ introduced move constructors, they couldn't do what Rust did and enforce that the name was no longer used after the move, as it was incompatible with past approaches, so move constructors are allowed to modify the original object and leave it in a "valid but unspecified state". Which is basically saying "it's no longer live, but we aren't going to stop you from touching it". Very annoying, but the additional freedom to steal resources from expiring objects is quite handy.

src/backends/policy_engine/souffle/check_policy_compliance.cc
cromwellian: Cool, great suggestions. I had thought of using the ? operator to avoid creating the scoping issue, but for some reason convinced myself the destructuring assignment might not work. :(

cromwellian:
Sorry if you're getting spammed with inprogress updates. I'm using the In-IDE resolution of coments and it seems to submit to github every time instead of queuing them up until I'm about to commit.

cromwellian (commented):
PTAL

Cypher1 (approved):
LGTM

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -61,6 +63,51 @@ absl::StatusOr<std::string> ReadFileContents(std::filesystem::path file_path) {
   return string_stream.str();
 }
 
+enum class ReturnCode : uint8_t { PASS = 0, FAIL, ERROR };
```
markww, line 66: For the sake of exit code stability, you may want to make every enumeration have an explicit value. Imagine if we added a case like this:

```C++
enum class ReturnCode : uint8_t { PASS = 0, PARTIAL_SUCCESS, FAIL, ERROR };
```

Now the `FAIL` and `ERROR` cases would silently change their value, which could cause issues if we had hardcoded them into tests outside of this file (a pretty reasonable thing to do for exit codes).

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -61,6 +63,51 @@ absl::StatusOr<std::string> ReadFileContents(std::filesystem::path file_path) {
...
+          .ir_module = std::move(ir_result.value().module),
+          .ir_ssa_names = std::move(ir_result.value().ssa_names),
+          .ir_context = std::move(context)};
+    } else {
+      return ir_result.status();
+    }
+  } else {
+    LOG(ERROR) << "Error reading proto file: " << proto_string.status();
+    return proto_string.status();
+  }
```
markww, line 108: What I see here is some code that basically wants to turn one `StatusOr` into another concisely. [Rust provides a `map` method for this purpose](https://doc.rust-lang.org/std/result/enum.Result.html#method.map), and there's no reason why we can't do that as well. We have a small little functional-programming-in-C++ library in `//src/common/utils`; perhaps we could add a `statusor_map` function there for handling this case more concisely.

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -72,36 +119,48 @@ int main(int argc, char* argv[]) {
...
-      !sql_policy_rules.ok()) {
+  if (absl::GetFlag(FLAGS_sql_policy_rules).has_value() && !sql_policy_rules.ok()) {
     LOG(ERROR) << "Error reading sql policy rules file: "
                << sql_policy_rules.status();
-    return 2;
+    return UnwrapExitCode(ReturnCode::ERROR);
+  }
+
+  auto ir_path = absl::GetFlag(FLAGS_ir);
+  auto proto_path = absl::GetFlag(FLAGS_proto);
```
markww, line 133: [`auto` should only be used when the type is obvious or when there's a substantial readability benefit.](https://google.github.io/styleguide/cppguide.html#Type_deduction)

src/backends/policy_engine/souffle/check_policy_compliance.cc
```diff
@@ -72,36 +119,48 @@ int main(int argc, char* argv[]) {
...
+      LOG(ERROR) << "Both --ir and --proto cannot be used simultaneously.";
+    } else {
+      LOG(ERROR) << "One of --ir or --proto must be specified.";
+    }
+
+    return UnwrapExitCode(ReturnCode::ERROR);
+  }
+
+  // Parse either an IR file or a Proto file
+  auto components = ir_path.has_value()
```
markww, line 146: Unclear `auto`, see other comment.

src/backends/policy_engine/souffle/check_policy_compliance_errors.sh
```diff
@@ -0,0 +1,34 @@
...
+SQL_POLICY_RULES_FILE_ARG=$2
+
+# A simple script to test the `check_policy_compliance` command line tool.
+ROOT_DIR=${TEST_SRCDIR}/${TEST_WORKSPACE}
+CMD=${ROOT_DIR}/${CMD_ARG}
+SQL_POLICY_RULES_FILE=${ROOT_DIR}/${SQL_POLICY_RULES_FILE_ARG}
+
+$CMD --ir=$FILE --proto=$FILE --sql_policy_rules=$SQL_POLICY_RULES_FILE
+CMD_RESULT=$?
+if [ ${CMD_RESULT} -eq 2 ]; then
```
markww, line 30: Maybe assign `2` to a local variable like `ERROR_EXIT_CODE` to attach a bit of documentation about what it means?

src/backends/policy_engine/souffle/check_policy_compliance_test.sh
```diff
@@ -22,19 +22,20 @@ FAIL_RESULT="fail"
 function printUsageAndExit() {
   echo "Usage: "
   echo "  check_policy_compliance_test.sh \ "
-  echo "     <cmd> ${PASS_RESULT}|${FAIL_RESULT} <ir_file> <sql_policy_rules_file> <policy_engine>"
+  echo "     <cmd> ${PASS_RESULT}|${FAIL_RESULT} <ir|proto> <file> <sql_policy_rules_file> <policy_engine>"
   exit 1
 }
 
-if [ $# -ne 5 || $# -ne 4]; then
+if [ $# -ne 6 || $# -ne 5]; then
```
markww, line 29: What do `6` and `5` mean here?

cromwellian: min # args and max # args, I added vars for them

src/backends/policy_engine/souffle/check_policy_compliance.cc
cromwellian: I think that's a great idea, maybe I should do it in a follow up PR? I feel like if I'm adding some useful utility function for everyone it should be a separate commit + tests?

markww: Totally, follow on PR is probably best to keep this PR focused on one aspect.

PR #675, opened 2022-08-19
Visitor for authorization logic AST

aferr:
This is exactly the same as  #643 which I did not bring into google3 correctly.


aferr:
Nope there is a better way!

PR #679, opened 2022-08-23
Fix for use of parser without exception support

Cypher1:
Fixes build breakage from 7b3011c `Connect the IR translator to the DP analysis.`


markww (approved):
I'm very confused that my changes caused this to occur, but I can see how it might happen. `souffle_policy_checker_test` exercises the `souffle_policy_checker`, which is allowed to throw exceptions. It's a bit unnerving that the continuous integration didn't catch this before it showed up on your machine; perhaps we need to exercise it with some more platforms.

Regardless, LGTM.

PR #680, opened 2022-08-23
Fix AuthLogic Visitor to not depend on evaluation order of arguments

bgogul:
If the visitor has side effects, this was causing different test results depending upon the compiler.


bgogul:
Will close this PR and create one from google repo.

Issue #676, opened 2022-08-19
Make tests comparing AST auth logic trees use `operator==` rather than equality of debug prints.


markww:
In #643 , we added a visitor for the authorization logic AST. The test for this compared a large number of debug printed strings from the AST nodes, which were introduced for this purpose alone. This isn't great, as if the test fails, a reader must parse a lot of text in their heads to diagnose how the failure occurred. In addition, debug printing usually means printing for debugging, which usually does not imply any particular stability requirements for tests, making these tests 

The relevant AST nodes should be refactored to just have an `operator==` and for those tests to compared via those `operator==`. If the test fails, someone can always insert print statements or go to the debugger to discover why the test is misbehaving.

bgogul:
I believe this is done?

PR #700, opened 2022-09-01
auth_logic AST operator==

aferr:


aferr:
Fixes #676

aferr:
As a side-effect this also moves the DebugPrint() functions that are longer than 2 lines to a .cc file.

src/ir/auth_logic/ast_equality_test.cc
```diff
@@ -61,4 +61,206 @@ TEST(AstEqualityTestSuite, Test7) {
   EXPECT_TRUE(!(p1 == p2));
 }
 
+TEST(AstEqualityTestSuite, PrinTest) {
```
bgogul, line 64: There is a better idiom for writing equality tests that is quite exhaustive.  Can we change to that pattern?

e.g., 

https://github.com/google-research/raksha/blob/258b6d96db68a8ab79e04d6817270bf5a3167f8d/src/ir/value_test.cc#L162

src/ir/auth_logic/ast.cc
```diff
@@ -0,0 +1,90 @@
...
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#include "src/ir/auth_logic/ast.h"
+
+namespace raksha::ir::auth_logic {
+
+// A potentially ugly print of the state in this class
+// for debugging only
+std::string BaseFact::DebugPrint() const {
```
bgogul, line 23: When we have the equality operators, will we need these debug print for tests at all?

src/ir/auth_logic/ast.h
```diff
@@ -225,15 +243,15 @@ class ConditionalAssertion {
...
-      rhs_strings.push_back(base_fact.DebugPrint());
-    }
-    return absl::StrCat(lhs_.DebugPrint(), ":-",
-                        absl::StrJoin(rhs_strings, ", "));
+  // for debugging only
+  std::string DebugPrint() const;
+
+  bool operator==(const ConditionalAssertion& rhs) const {
+    return lhs_ == rhs.lhs() && rhs_ == rhs.rhs();
+  }
```
bgogul, line 251: Might be a better idea to refer to the fields directly: 
```suggestion
  bool operator==(const ConditionalAssertion& other) const {
    return lhs_ == other.lhs_ && rhs_ == other.rhs_;
  }
```


Can you change the other places also?

src/ir/auth_logic/ast.h
```diff
@@ -265,14 +283,15 @@ class Assertion {
...
-  std::string DebugPrint() const {
-    return absl::StrCat(
-        "Assertion(",
-        std::visit([](auto& obj) { return obj.DebugPrint(); }, this->value_),
-        ")");
+  // for debugging only
+  std::string DebugPrint() const;
+
+  bool operator==(const Assertion& rhs) const {
+    return value_ == rhs.GetValue();
```
bgogul, line 290: value_ == other.value_

src/ir/auth_logic/ast.h
```diff
@@ -337,12 +354,19 @@ class Query {
...
   // A potentially ugly print of the state in this class
-  // for debugging/testing only
+  // for debugging only
   std::string DebugPrint() const {
     return absl::StrCat("Query(", name_, principal_.DebugPrint(),
                         fact_.DebugPrint(), ")");
   }
 
+  bool operator==(const Query& rhs) const {
+    return name_ == rhs.name() && principal_ == rhs.principal() &&
```
bgogul, line 364: refer to fields directly to make the fieldwise comparison obvious. (Please change other implementations too.)

src/ir/auth_logic/ast.h
```diff
@@ -382,32 +406,17 @@ class Program {
...
-        std::vector<std::string>({"Program(",
-                                  absl::StrJoin(relation_decl_strings, "\n"),
-                                  absl::StrJoin(says_assertion_strings, "\n"),
-                                  absl::StrJoin(query_strings, "\n")}),
-        "\n");
+  // for debugging only
+  std::string DebugPrint() const;
+
+  bool operator==(const Program& rhs) const {
+    return relation_declarations_ == rhs.relation_declarations() &&
```
bgogul, line 413: refer to fields

bgogul (commented):
Thank you for doing this. I have left A few readability and testing related comments.

src/ir/auth_logic/ast.cc
aferr: No the `DebugPrint` isn't needed for anything -- it will be technically dead code but it could be used for debugging. Would you rather I delete all the DebugPrints?

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -209,21 +210,64 @@ TEST(AuthLogicAstTraversingVisitorTest, SimpleTraversalTest) {
...
-          baseFact1.DebugPrint(), saysAssertion2.DebugPrint(),
-          prinC.DebugPrint(), assertion1.DebugPrint(), fact1.DebugPrint(),
-          baseFact1.DebugPrint(), assertion2.DebugPrint(), fact2.DebugPrint(),
-          prinB.DebugPrint(), baseFact2.DebugPrint(), query1.DebugPrint(),
-          prinA.DebugPrint(), fact1.DebugPrint(), baseFact1.DebugPrint(),
-          query2.DebugPrint(), prinB.DebugPrint(), fact2.DebugPrint(),
-          prinB.DebugPrint(), baseFact2.DebugPrint()));
-
+          std::addressof(program1),
+          saysAssertion1ptr,
```
aferr, line 246: @bgogul I am trying to fix this test to work without `DebugPrint`. This is an attempt at collecting the addresses of the ast nodes in the visitor rather than collecting `DebugPrint` strings (see line 185), but with the way `auth_logic::ast` is implemented, I am not sure there is a way to actually get the right addresses. And in any case this looks like it would be awful to update if the test needs to be updated.

Is there a better way to write this test?

The next thing I was thinking to try is to write a class `ASTNode` that has `==operator` as a pure virtual method and having both the `ast.h` and `program.h` classes extend from this. Then I think this visitor could use `std::vector<ASTNode` as the type of `node_` as long as `EXPECT_THAT` would then use the `==` operator in that case. Do you happen to know if that will/won't work or is there a better way to do this?

aferr: @bgogul OK I tried using the abstract class idea instead, but it looks like this isn't compatible with `EXPECT_THAT` for some reason?

I get:
```
external/com_google_googletest/googletest/include/gtest/gtest-matchers.h:880:40: error: no matching function for call to 'Eq'
```

Any idea if I am just implementing this idea wrong or if this fundamentally won't work?

aferr: Perhaps @markww might also know why `ASTNode` is not enough to make `auth_logic_ast_traversing_visitor_test` work with `EXPECT_THAT` and a `std::vector<ASTNode>`

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -86,102 +87,102 @@ class TraversalOrderVisitor
...
   }
 
-  const std::vector<std::string>& nodes() const { return nodes_; }
+  const std::vector<const void*>& nodes() const { return nodes_; }
 
  private:
   bool pre_visits_;
   bool post_visits_;
-  std::vector<std::string> nodes_;
+  std::vector<const void*> nodes_;
```
markww: `void*` should basically never be used. This should be either some base class of the possibilities or, if you don't care about dereferencing the pointer and only want it as a number to be compared/hased/etc, an `intptr_t` or `uintptr_t`.

src/ir/ast_node.h
```diff
@@ -0,0 +1,28 @@
...
+namespace raksha::ir {
+// This is the supertype for all nodes in
+// `raksha::ir::datalog::program.h` and 
+// `raksha::ir::auth_logic::ast.h`
+class ASTNode {
+    virtual bool operator==(const ASTNode& otherNode ) const = 0;
+};
+
+}; // raksha::ir
+#endif // SRC_IR_AST_NODE_H_
```
markww, line 28: Nit: newline at EOF.

src/ir/auth_logic/ast.h
```diff
@@ -75,24 +86,34 @@ class Attribute {
...
+    return principal_ == rhs.principal() && predicate_ == rhs.predicate();
   }
 
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Attribute & rhsAttribute = dynamic_cast<const Attribute&>(rhs);
+      return *this == rhsAttribute;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 99: Rather than having this `try..catch` structure, just `dynamic_cast` to `const Attribute *` like so:

```C++
const Attribute *rhs_attribute = dynamic_cast<const Attribute *>(&rhs);
```

If the cast fails, `rhs` will contain `nullptr`, and you can return false in that case, like so:

```C++
if (const Attribute *rhs_attribute = dynamic_cast<const Attribute *>(&rhs)) {
  return *this == *rhs_attribute;
}
return false;
```

src/ir/auth_logic/ast.h
```diff
@@ -107,13 +128,22 @@ class CanActAs {
...
+           right_principal_ == rhs.right_principal();
   }
 
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const CanActAs & rhsCanActAs = dynamic_cast<const CanActAs&>(rhs);
+      return *this == rhsCanActAs;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 143: Don't use a `try...catch` for this, as mentioned in another comment.

src/ir/auth_logic/ast.h
```diff
@@ -148,26 +179,31 @@ class BaseFact {
...
+  bool operator==(const BaseFact& rhs) const {
+    return value_ == rhs.GetValue();
   }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const BaseFact & rhsBaseFact = dynamic_cast<const BaseFact&>(rhs);
+      return *this == rhsBaseFact;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 191: Don't use a `try...catch` for this, as mentioned in another comment.

src/ir/auth_logic/ast.h
```diff
@@ -186,17 +222,21 @@ class Fact {
...
+    return delegation_chain_ == rhs.delegation_chain() &&
+           base_fact_ == rhs.base_fact();
+  }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Fact & rhsFact = dynamic_cast<const Fact&>(rhs);
+      return *this == rhsFact;
+    } catch(const std::bad_cast& e) {
+      return false;
     }
```
markww, line 235: Don't use a `try...catch` for this.

src/ir/auth_logic/ast.h
```diff
@@ -223,16 +264,20 @@ class ConditionalAssertion {
...
+  bool operator==(const ConditionalAssertion& rhs) const {
+    return lhs_ == rhs.lhs() && rhs_ == rhs.rhs();
+  }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const ConditionalAssertion & rhsCondAssertion = dynamic_cast<const ConditionalAssertion&>(rhs);
+      return *this == rhsCondAssertion;
+    } catch(const std::bad_cast& e) {
+      return false;
     }
```
markww, line 276: Don't use a try catch for this.

src/ir/auth_logic/ast.h
```diff
@@ -263,24 +309,30 @@ class Assertion {
...
+  bool operator==(const Assertion& rhs) const {
+    return value_ == rhs.GetValue();
   }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Assertion & rhsAssertion = dynamic_cast<const Assertion&>(rhs);
+      return *this == rhsAssertion;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 321: Don't use a try catch for this.

src/ir/auth_logic/ast.h
```diff
@@ -295,31 +347,34 @@ class SaysAssertion {
...
-      assertion_strings.push_back(assertion.DebugPrint());
+  bool operator==(const SaysAssertion& rhs) const {
+    return principal_ == rhs.principal() && assertions_ == rhs.assertions();
+  }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const SaysAssertion & rhsSays = dynamic_cast<const SaysAssertion&>(rhs);
+      return *this == rhsSays;
+    } catch(const std::bad_cast& e) {
+      return false;
```
markww, line 358: Don't use a try catch for  this.

src/ir/auth_logic/ast.h
```diff
@@ -335,28 +390,38 @@ class Query {
...
+           fact_ == rhs.fact();
+  }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Query & rhsQuery = dynamic_cast<const Query&>(rhs);
+      return *this == rhsQuery;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
   }
```
markww, line 405: Don't use a try-catch for this.

src/ir/auth_logic/ast.h
```diff
@@ -380,31 +445,23 @@ class Program {
...
+           says_assertions_ == rhs.says_assertions() &&
+           queries_ == rhs.queries();
+  }
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Program & rhsProgram = dynamic_cast<const Program&>(rhs);
+      return *this == rhsProgram;
+    } catch(const std::bad_cast& e) {
+      return false;
     }
```
markww, line 460: Don't use a try catch for this.

src/ir/ast_node.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef SRC_IR_AST_NODE_H_
+#define SRC_IR_AST_NODE_H_
+
+namespace raksha::ir {
+// This is the supertype for all nodes in
+// `raksha::ir::datalog::program.h` and 
+// `raksha::ir::auth_logic::ast.h`
+class ASTNode {
+    virtual bool operator==(const ASTNode& otherNode ) const = 0;
+};
```
markww, line 25: Because `ASTNode` is intended only to be an interface, make it so that it can only be constructed by its subclasses, like so:

```
class ASTNode {
  public:
  virtual bool operator==(const ASTNode& otherNode ) const = 0;
  protected:
  ASTNode() = default;
};
```

We'll see why in the `std::vector<ASTNode>` that you create later. Also, style guide suggests this should be `AstNode`.

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
```diff
@@ -86,102 +86,102 @@ class TraversalOrderVisitor
...
   }
 
-  const std::vector<std::string>& nodes() const { return nodes_; }
+  const std::vector<ASTNode>& nodes() const { return nodes_; }
 
  private:
   bool pre_visits_;
   bool post_visits_;
-  std::vector<std::string> nodes_;
+  std::vector<ASTNode> nodes_;
```
markww, line 184: This will not do what you want. This is saying "`nodes_` is a `std::vector` containing directly-embedded instances of `ASTNode`". When you look at what an `ASTNode` is, it's underwhelming:

```C++
class ASTNode {
    virtual bool operator==(const ASTNode& otherNode ) const = 0;
};
```

An `ASTNode` contains no members and has no method definitions. It is useless on its own. It is only useful as an interface through which to access other classes.

An annoying thing coming to C++ from languages which emphasize true sum types or which hide their pointers is that the sum-type-ness of using an abstract class in C++ only "activates" if you dereference a pointer. So calling a virtual method on a value of type `ASTNode *` may cause a subclass's method to be called, but calling a virtual method on a value of type `ASTNode` will use the version in `ASTNode` itself to be used only.

This seam comes from both the history and the mechanical implementation of derived classes in C++. It's not far off to say "is a base type of" in C++ is the same as "is a prefix of" the type in its memory layout. If you're writing C in the pre-C++ days, but you find yourself wanting to write derived classes, you end up having some type be the prefix of your family of derived classes, pass around a pointer to that prefix, and put some kind of hint to the full type in that prefix. Then, when you need the full type, you use the hint to perform the cast and widen the pointer's element type to the full type.

C++ builds off of this C pattern by making this cast occur automatically upon the pointer dereference and, rather than forcing you to put your own hint in the prefix, uses the vtable pointer itself as the hint. Giant improvement over the C version!

But, just like in the C version, if you end up copying that prefix, you've essentially "beheaded" the structure. If the interesting stuff was in the tail, you've left it behind, and potentially opened yourself up to a memory error by looking for a tail that isn't there.

To help prevent yourself from making this error, I suggest making the constructor for `ASTNode` `protected`, thus preventing you from accidentally constructing objects of that type directly (as you do here). This should change to be something like `std::vector<const ASTNode *>` or `std::vector<std::unique_ptr<ASTNode>>`, depending on ownership story.

src/ir/datalog/program.h
```diff
@@ -58,6 +61,15 @@ class Predicate {
            this->args() == otherPredicate.args();
   }
 
+  bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Predicate& rhsPredicate= dynamic_cast<const Predicate&>(rhs);
+      return *this == rhsPredicate;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 70: Don't use try-catch for this.

src/ir/datalog/program.h
```diff
@@ -82,64 +88,89 @@ class Predicate {
...
+  }
+
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const ArgumentType& rhsArgType= dynamic_cast<const ArgumentType&>(rhs);
+      return *this == rhsArgType;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
+  }
```
markww, line 112: Don't use try-catch for this.

src/ir/datalog/program.h
```diff
@@ -82,64 +88,89 @@ class Predicate {
...
+    this->argument_type_ == otherArgument.argument_type();
   }
 
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Argument& rhsArgument= dynamic_cast<const Argument&>(rhs);
+      return *this == rhsArgument;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 139: Don't use try-catch for this.

src/ir/datalog/program.h
```diff
@@ -82,64 +88,89 @@ class Predicate {
...
-    arg_strings.reserve(arguments_.size());
-    for (const Argument& arg : arguments_) {
-      arg_strings.push_back(arg.DebugPrint());
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const RelationDeclaration& rhsRelDecl = 
+        dynamic_cast<const RelationDeclaration&>(rhs);
+      return *this == rhsRelDecl;
+    } catch(const std::bad_cast& e) {
+      return false;
```
markww, line 171: Don't use try-catch for this.

src/ir/datalog/program.h
```diff
@@ -150,32 +181,62 @@ class RelationDeclaration {
...
+    this->rhs_ == otherRule.rhs();
+  }
+
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Rule& rhsRule = dynamic_cast<const Rule&>(rhs);
+      return *this == rhsRule;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 204: Don't use try-catch for this.

src/ir/datalog/program.h
```diff
@@ -150,32 +181,62 @@ class RelationDeclaration {
...
+      this->outputs_ == otherProgram.outputs();
+  }
+
+  virtual bool operator==(const ASTNode& rhs) const override {
+    try {
+      const Program& rhsProgram= dynamic_cast<const Program&>(rhs);
+      return *this == rhsProgram;
+    } catch(const std::bad_cast& e) {
+      return false;
+    }
```
markww, line 238: Don't use try-catch for this.

src/ir/auth_logic/ast_equality_test.cc
aferr: fixed

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h

src/ir/ast_node.h
aferr: resolbed by this file not extisting anymore

src/ir/auth_logic/auth_logic_ast_traversing_visitor_test.cc
aferr: resolved

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h

src/ir/auth_logic/ast.h
aferr: @markww Just to save yourself some typing in the future, you could try writing comments to express a general pattern in one place and requesting me to do this "throughout the code". Of course you can continue to keep doing it this way if you prefer.

src/ir/ast_node.h
aferr: fixed in a way

src/ir/auth_logic/ast_equality_test.cc
```diff
@@ -61,4 +61,278 @@ TEST(AstEqualityTestSuite, Test7) {
...
+  auto &[obj_and_eq_class1, obj_and_eq_class2] = GetParam();
+  auto &[obj1, eq_class1] = obj_and_eq_class1;
+  auto &[obj2, eq_class2] = obj_and_eq_class2;
+
+  EXPECT_EQ(obj1 == obj2, eq_class1 == eq_class2);
+  EXPECT_EQ(obj1 != obj2, eq_class1 != eq_class2);
+}
+
+static datalog::Predicate pred1("foo", {"bar", "baz"}, datalog::kPositive);
+static datalog::Predicate pred2("foooo", {"bar", "baz"}, datalog::kPositive);
```
bgogul, line 101: Do we need a case for kNegative also?

src/ir/datalog/program.h
```diff
@@ -92,6 +92,11 @@ class ArgumentType {
 
   std::string DebugPrint() const { return absl::StrCat(kind_, name_); }
 
+  bool operator==(const ArgumentType& otherType) const {
+    return this->kind_ == otherType.kind() &&
```
bgogul, line 96: Can you add a comment as to why we ignore the name?

src/ir/datalog/program.h
```diff
@@ -127,6 +137,12 @@ class RelationDeclaration {
   absl::string_view relation_name() const { return relation_name_; }
   bool is_attribute() const { return is_attribute_; }
   const std::vector<Argument>& arguments() const { return arguments_; }
+  
+  bool operator==(const RelationDeclaration& otherDeclaration) const {
+    return this->relation_name_ == otherDeclaration.relation_name() &&
+    this->is_attribute_ == otherDeclaration.is_attribute() &&
+    this->arguments_ == otherDeclaration.arguments();
```
bgogul, line 144: Use the fields directly?
```suggestion
    return this->relation_name_ == otherDeclaration.relation_name_ &&
    this->is_attribute_ == otherDeclaration.is_attribute_ &&
    this->arguments_ == otherDeclaration.arguments_;
```

src/ir/datalog/program.h
```diff
@@ -158,6 +174,11 @@ class Rule {
   const Predicate& lhs() const { return lhs_; }
   const std::vector<Predicate>& rhs() const { return rhs_; }
 
+  bool operator==(const Rule& otherRule) const {
+    return this->lhs_ == otherRule.lhs() &&
+    this->rhs_ == otherRule.rhs();
```
bgogul, line 179: Use fields directly:
```suggestion
    return this->lhs_ == otherRule.lhs_ &&
    this->rhs_ == otherRule.rhs_;
```

src/ir/datalog/program.h
```diff
@@ -176,6 +197,13 @@ class Program {
   const std::vector<Rule>& rules() const { return rules_; }
   const std::vector<std::string>& outputs() const { return outputs_; }
 
+  bool operator==(const Program& otherProgram) const {
+    return this->relation_declarations_ == 
+      otherProgram.relation_declarations() &&
+      this->rules_ == otherProgram.rules() &&
+      this->outputs_ == otherProgram.outputs();
```
bgogul, line 204: Use fields directly. (Can you make sure you change other places also?)
```suggestion
    return this->relation_declarations_ == 
      otherProgram.relation_declarations_ &&
      this->rules_ == otherProgram.rules_ &&
      this->outputs_ == otherProgram.outputs_;
```

src/ir/datalog/program.h
```diff
@@ -111,6 +116,11 @@ class Argument {
     return absl::StrCat(argument_name_, " : ", argument_type_.DebugPrint());
   }
 
+  bool operator==(const Argument& otherArgument) const {
+    return this->argument_name_ == otherArgument.argument_name() &&
+    this->argument_type_ == otherArgument.argument_type();
```
bgogul, line 121: Use fields directly: 
```suggestion
    return this->argument_name_ == otherArgument.argument_name_ &&
    this->argument_type_ == otherArgument.argument_type_;
```

src/ir/datalog/program.h
```diff
@@ -92,6 +92,11 @@ class ArgumentType {
 
   std::string DebugPrint() const { return absl::StrCat(kind_, name_); }
 
+  bool operator==(const ArgumentType& otherType) const {
```
bgogul, line 95: Refer to the fields directly from `otherType`.

bgogul (approved):
Looks good. Just some minor nits here and there.

src/ir/auth_logic/ast_equality_test.cc
aferr: fixed

src/ir/datalog/program.h
aferr: fixed

src/ir/auth_logic/ast.h
```diff
@@ -77,10 +81,16 @@ class Attribute {
 
   // A potentially ugly print of the state in this class
   // for debugging/testing only
-  std::string DebugPrint() const {
-    return absl::StrCat(principal_.name(), predicate_.DebugPrint());
+  std::string ToString() const {
```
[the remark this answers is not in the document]

src/ir/auth_logic/ast.h
```diff
@@ -405,6 +458,14 @@ class Program {
                         absl::StrJoin(query_strings, "\n"), ")");
   }
 
+  // for debugging/testing only
```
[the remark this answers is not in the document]

src/ir/datalog/program.h
aferr: fixed

src/ir/datalog/program.h
aferr: fixed

src/ir/datalog/program.h
aferr: fixed

src/ir/datalog/program.h
aferr: fixed

PR #714, opened 2022-09-09
Add targets for building Souffle in 64-bit mode.

markww:
This change adds targets for building Souffle in 64-bit mode, as opposed to the default 32-bit mode. This does not affect the target architecture, just the size of words that Souffle uses in its own data structures.


bgogul (approved):
Is this needed in addition to the changes that were done internally and exported as https://github.com/google-research/raksha/pull/713

markww:
Yes, this is needed to complete that change.

PR #723, opened 2022-09-15
Rename DebugPrint to ToString

aferr:


bgogul (approved):
Can you please add tests in a follow up PR?

PR #726, opened 2022-09-20
Convenience Factories for datalog::ArgumentType

aferr:
This was originally done as part of #654 but has
been split into a separate PR in the interest
of keeping PRs more singular.

Note that because #654 uses these methods, #654 depends on this PR.

### DIFF
--- a/src/ir/datalog/argument_type_factory_test.cc
+++ b/src/ir/datalog/argument_type_factory_test.cc
@@ -1,0 +1,37 @@
+//-----------------------------------------------------------------------------
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//----------------------------------------------------------------------------
+
+#include "src/common/testing/gtest.h"
+#include "src/ir/datalog/program.h"
+
+namespace raksha::ir::datalog {
+
+TEST(ArgumentTypeFactoryTest, SimpleFactoryTest) {
+  ArgumentType number = ArgumentType::MakeNumberType();
+  ArgumentType principal = ArgumentType::MakePrincipalType();
+  ArgumentType custom1 = ArgumentType::MakeCustomType("Custom1");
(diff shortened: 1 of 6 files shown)


src/ir/datalog/program.h
```diff
@@ -89,6 +89,15 @@ class ArgumentType {
       : kind_(kind), name_(name) {}
   Kind kind() const { return kind_; }
   absl::string_view name() const { return name_; }
+  static ArgumentType MakePrincipalType() {
```
bgogul, line 92: Since we are making this part of the API, adding some tests would be good. 

Also, should you make the constructor private if you only expect this API to be used to create objects?

src/ir/datalog/program.h
```diff
@@ -89,6 +89,15 @@ class ArgumentType {
...
+    return ArgumentType(Kind::kPrincipal, "Principal");
+  }
+  static ArgumentType MakeNumberType() {
+    return ArgumentType(Kind::kNumber, "Number");
+  }
+  static ArgumentType MakeCustomType(absl::string_view name) {
+    return ArgumentType(Kind::kCustom, std::move(name));
```
bgogul, line 99: I don't `std::move` needed here.

bgogul (commented):
Add simple tests?

src/ir/datalog/program.h
aferr: done.

src/ir/datalog/program.h
aferr: done

aferr: done

src/ir/datalog/argument_type_factory_test.cc
```diff
@@ -0,0 +1,37 @@
...
+  ArgumentType custom2 = ArgumentType::MakeCustomType("Custom2");
+  EXPECT_EQ(number.name(), "Number");
+  EXPECT_EQ(principal.name(), "Principal");
+  EXPECT_EQ(custom1.name(), "Custom1");
+  EXPECT_EQ(custom2.name(), "Custom2");
+  EXPECT_EQ(number.kind(), ArgumentType::Kind::kNumber);
+  EXPECT_EQ(principal.kind(), ArgumentType::Kind::kPrincipal);
+  EXPECT_EQ(custom1.kind(), ArgumentType::Kind::kCustom);
+  EXPECT_EQ(custom2.kind(), ArgumentType::Kind::kCustom);
+}
```
bgogul, line 35: Split this into three different tests: `MakeNumberType`, `MakePrincipalType`, and `MakeCustomType`.

bgogul (approved):
Looks good modulo a minor refactoring in the tests.

src/ir/datalog/argument_type_factory_test.cc
aferr: fixed

Issue #720, opened 2022-09-14
Write Tests for the ToString methods in auth_logic/ast.h and datalog/program.h


aferr:
(Note that these ToString methods were previously called DebugPrint)

It turns out that DebugPrint was actually used to implement `auth_logic_ast_traversing_visitor_test.cc` and between me and Gogul we were not able to think of a way to re-implement this test without using strings within a reasonable timeframe (see also: [this code](https://github.com/google-research/raksha/tree/ast-equality3%40aferr)). 

Instead of deprecating this method, we should write tests for `ToString`. Then it's OK for tests to depend on it.

aferr:
Consider also extending https://github.com/google-research/raksha/blob/main/src/ir/auth_logic/ast_string.h so that it can be used instead of these methods.

PR #727, opened 2022-09-20
Small improvements to Auth Logic ToString

aferr:
These were originally made as part of #654, but
are split into this separate PR in order to
keep PRs more singular in purpose.

Tests of ToString will still be added later in a
separate PR in order to satisfy #720.


bgogul (approved):
Ideally, we should have tests, but will defer for now due to #720.

aferr:
@bgogul do you know why this one wasn't merged by copybara?

PR #731, opened 2022-09-22
ToString Tests

aferr:
This adds tests to `ast_string.h` which we apparently already had and already used for other tests in the repository. However, this code didn't have any tests itself. So rather than have two ways of making a string from the AST, I updated `ast_string.h` so that it is good enough for the uses of the ToString methods in the AST and added some tests for `ast_string.h`. In a follow-up PR, I will deprecate the ToString methods in the AST class and use just `ast_string.h` for the remaining tests thereby satisfying #720.


aferr:
I am not deprecating the `ToString` methods of `ast.h` in this PR and will only do that as a separate PR in interest of keeping the PRs simpler.

src/ir/auth_logic/ast_construction_says_assertion_test.cc
```diff
@@ -61,47 +61,52 @@ INSTANTIATE_TEST_SUITE_P(
         SaysAssertionTestData({R"("UserA" says { currTimeIs(2022).
                                mayA(NotifyA) :- currTimeIs > 2021. })",
                                R"("UserA")",
-                               {"currTimeIs(2022)",
-                                "mayA(NotifyA) :- currTimeIs > 2021"}}),
+                               {"Fact(BaseFact(currTimeIs(2022))).",
```
bgogul, line 64: It looks like the existing `ToString` was meant to be human readable. With this PR, we are making it expose the AST hierarchy. Is that desirable?

IIRC, one (or the only?) reason we exposed the AST nodes like `Fact` and `BaseFact` was to test the visitor. With https://github.com/google-research/raksha/commit/9706fa433bf0281ee4f2e4829bb19c9139f1f911, we do not need `ToString` for testing the visitor, right? Did we decide that it is not possible to fix the ast visitor tests without `ToString`?

src/ir/auth_logic/ast_string.h
```diff
@@ -48,58 +50,123 @@ std::string ToString(AstNodeVariantType Node) {
                 predicate.name(), "(", absl::StrJoin(predicate.args(), ", "),
                 ")");
           },
+          [](datalog::ArgumentType argumentType) {
```
bgogul, line 53: Should `ToString` be implemented using the AST visitor? (Let us discuss and agree what the right thing is. Don't make the change yet.)

bgogul (commented):
I have some questions before I do a deeper review. Can you take a look, please?

src/ir/auth_logic/ast_construction_says_assertion_test.cc
aferr: This `ToString` function that I did not write and that was already in our repository and used for several tests that I also did not write is used for debugging and testing, and for both of these use-cases distinguishing between the AST nodes seems useful. For both of these use cases exposing the AST is important.

The souffle emitter is a way of converting the AST into strings that are souffle code. That emitter is distinct from this ToString and has a different purpose. It does not expose the AST.

As I tried to explain in the comment at the top of this PR, this `ToString` function defined in `ast_string.h` is distinct from the `ToString` methods inside `ast.h` [like this one](https://github.com/google-research/raksha/blob/main/src/ir/auth_logic/ast.h#L53). The whole point of this PR is to make the `ToString` function in `ast_string.h` sufficient for the uses of the method in `ast.h` so I can delete the methods in `ast.h` and use only `ast_string.h` later to remove this redundancy.

https://github.com/google-research/raksha/commit/9706fa433bf0281ee4f2e4829bb19c9139f1f911 did not remove the need to have a ToString method for the visitor test. You and I sitting together for a full hour were not able to think of a way to test the visitor with equality because in C++ there is no way to declare a vector of a class with a pure virtual method. So we still need it for that test. Also as mentioned in this comment, there were already tests in the code (that I did not write) that used ToString for testing.

src/ir/auth_logic/ast_string.h
aferr: I don't think that is better because emitting strings is a function on the AST where the structure of the AST and the function performed cannot be usefully decoupled because emitting a string is a function that requires a very specific order (which may not be exactly a pre-order or post-order traversal). (I think an example of a function that would make better use of the visitor pattern would be something like computing the set of subexpressions in a program with a particular property.)

src/ir/auth_logic/ast_construction_says_assertion_test.cc
bgogul: I understood that you were consolidating the `ToString` methods. I was actually wondering if the `ToString` in `ast_string.h` is meant to be human readable and if this change is removing this functionality. 

BTW, the use of `std::variant` in this PR gave me an idea to remove the dependency on `ToString` for ast visitor tests. (See  https://github.com/google-research/raksha/pull/734/files and the internal CL.)

bgogul: While I was at it, I went ahead and removed the `ToString` methods from Ast as well. (See https://github.com/google-research/raksha/pull/736 and the internal CL)

aferr: > BTW, the use of std::variant in this PR gave me an idea to remove the dependency on ToString for ast visitor tests. (See https://github.com/google-research/raksha/pull/734/files and the internal CL.)

Ah nice that we have a way to fix it. I can't see the change though. When I look at that link it has 0 commits and 0 files changed... hmm...

src/ir/auth_logic/ast_string.h
bgogul: SG. We can leave this as is.

src/ir/auth_logic/ast_construction_says_assertion_test.cc
aferr: I thought that outputting the details of the AST was a good thing because this `ToSTring` is meant to help with debugging. If you strongly feel it should not output those details, I can change it the other way so that we can make forward progress though!

aferr: I guess this print is also used in tests, so changing this print means the tests also have to change... OK I can make this toString output fewer details of the AST then

aferr: Ah weirdly I have an e-mail with a reply from you in this thread, but it isn't showing up in this thread on the github webpage!

aferr:
@bgogul ready for re-review

src/ir/auth_logic/ast_string_test.cc
```diff
@@ -0,0 +1,154 @@
...
+}
+
+static Query query1("query1", principal1, fact1);
+static Query query2("query2", principal2, fact1);
+TEST(QueryToStringTest, SimpleQueryTest) {
+  EXPECT_EQ(ToString(query1),
+            "Query(query1 = principal1 says predicate1(arg1, "
+            "arg2)?)");
+  EXPECT_EQ(ToString(query2),
+            "Query(query2 = principal2 says predicate1(arg1, "
```
bgogul, line 141: Do we need `Query` in the output? If this is for human consumption, `?` at the end is sufficient to mark this as a query?

src/ir/auth_logic/ast_string.h
```diff
@@ -48,58 +50,116 @@ std::string ToString(AstNodeVariantType Node) {
...
+            for (const SaysAssertion& says_assertion :
+                 program.says_assertions()) {
+              says_assertion_strings.push_back(ToString(says_assertion));
+            }
+            std::vector<std::string> query_strings;
+            query_strings.reserve(program.queries().size());
+            for (const Query& query : program.queries()) {
+              query_strings.push_back(ToString(query));
+            }
+            return absl::StrCat("Program(\n",
```
bgogul, line 159: Is `Program(` needed  at all?

src/ir/auth_logic/ast_string.h
```diff
@@ -48,58 +50,116 @@ std::string ToString(AstNodeVariantType Node) {
...
+            assertion_strings.reserve(saysAssertion.assertions().size());
+            for (const Assertion& assertion : saysAssertion.assertions()) {
+              assertion_strings.push_back(ToString(assertion));
+            }
+            return absl::StrCat(ToString(saysAssertion.principal()),
+                                " says {\n",
+                                absl::StrJoin(assertion_strings, "\n"), "}");
+          },
+          [](Query query) {
+            return absl::StrCat("Query(", query.name(), " = ",
```
bgogul, line 136: is `Query(` needed? Just omit it?

src/ir/auth_logic/ast_string.h
```diff
@@ -48,58 +50,116 @@ std::string ToString(AstNodeVariantType Node) {
...
+            says_assertion_strings.reserve(program.says_assertions().size());
+            for (const SaysAssertion& says_assertion :
+                 program.says_assertions()) {
+              says_assertion_strings.push_back(ToString(says_assertion));
+            }
+            std::vector<std::string> query_strings;
+            query_strings.reserve(program.queries().size());
+            for (const Query& query : program.queries()) {
+              query_strings.push_back(ToString(query));
+            }
```
bgogul, line 158: We can replace these loops with `utils::MapIter`?

src/ir/auth_logic/ast_string.h
```diff
@@ -48,58 +50,116 @@ std::string ToString(AstNodeVariantType Node) {
...
+            return absl::StrCat(argument.argument_name(), " : ",
+                                ToString(argument.argument_type()));
+          },
+          [](datalog::RelationDeclaration relationDeclaration) {
+            std::vector<std::string> arg_strings;
+            arg_strings.reserve(relationDeclaration.arguments().size());
+            for (const datalog::Argument& arg :
+                 relationDeclaration.arguments()) {
+              arg_strings.push_back(ToString(arg));
+            }
```
bgogul, line 73: use `utils::MapIter`?

src/ir/auth_logic/ast_string.h
```diff
@@ -48,58 +50,116 @@ std::string ToString(AstNodeVariantType Node) {
                 predicate.name(), "(", absl::StrJoin(predicate.args(), ", "),
                 ")");
           },
+          [](datalog::ArgumentType argumentType) {
+            switch (argumentType.kind()) {
+              case datalog::ArgumentType::Kind::kCustom:
+                return absl::StrCat(argumentType.name(), "::CustomType");
```
bgogul, line 56: Why `Type` suffix? Just `Custom`, `Number` and `Principal`?

bgogul (approved):
Thank you, Andrew.  I have some clean up suggestions. Looks good otherwise.

src/ir/auth_logic/ast_string.h
aferr: fixed

src/ir/auth_logic/ast_string.h
aferr: fixed

src/ir/auth_logic/ast_string.h
aferr: fixed

src/ir/auth_logic/ast_string_test.cc
```diff
@@ -0,0 +1,154 @@
...
+            "arg2)?)");
+}
+
+static Program program1({relationDeclaration1}, {saysAssertion1}, {query1});
+TEST(ProgramToStringTest, SimpleProgramTest) {
+  EXPECT_EQ(
+      ToString(program1),
+      "Program(\n.decl rel1(arg1 : NumberType, arg2 : PrincipalType)principal1 "
```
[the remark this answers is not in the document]
aferr: fixed

src/ir/auth_logic/ast_string.h
aferr: fixed

src/ir/auth_logic/ast_string_test.cc
aferr: fixed

PR #741, opened 2022-09-27
Add duplicate `http_archive`s for absl and re2 with official names.

markww:
Although Bazel gives you the option of naming your http_archive repositories whatever you want, it apparently would really prefer if everyone chose the same name everywhere. This becomes apparent where fuzztest really wants the top level `WORKSPACE` to call absl `com_google_absl`, and Bazel will not allow me to use the shorter name `absl`. This is especially annoying because both `WORKSPACE`s describe `absl` using the same URL and sha256 hash, so you'd hope it would identify them as the same dependency? Discussions online do not give me great hope in flexibility from Bazel devs on this point:
https://github.com/bazelbuild/bazel/issues/3219

In any case, for now I am duplicating the dependency in two different `http_archive`s. I will remove the shorter, more readable `absl` and `re2` in the future once I have cut over to the new dependency.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -148,6 +148,13 @@ http_archive(
     urls = ["https://github.com/abseil/abseil-cpp/archive/refs/tags/20220623.0.tar.gz"],
 )
 
+http_archive(
+    name = "com_google_absl",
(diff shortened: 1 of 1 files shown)


markww:
This is needed for #742

bgogul (approved):

PR #743, opened 2022-09-27
Turn off the unknown warning warning.

markww:
This is to work around my encouter of
https://github.com/google/fuzztest/issues/34 .

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,6 +20,12 @@ build --host_cxxopt='-std=c++17'
 # warnings from our third party packages.
 # build --host_cxxopt='-Werror' --host_cxxopt='-Wall'
 build --host_cxxopt='-Wno-deprecated-declarations'
+
+# Annoyingly, in the fuzztest library, their attempt to turn off the `-Wpsabi`
+# warning in Clang causes Clang to complain about an unknown warning flag. This
+# is our attempt to work around that.
+build --cxxopt='-Wno-unknown-warning-option'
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -20,6 +20,12 @@ build --host_cxxopt='-std=c++17'
 # warnings from our third party packages.
 # build --host_cxxopt='-Werror' --host_cxxopt='-Wall'
 build --host_cxxopt='-Wno-deprecated-declarations'
+
+# Annoyingly, in the fuzztest library, their attempt to turn off the `-Wpsabi`
+# warning in Clang causes Clang to complain about an unknown warning flag. This
+# is our attempt to work around that.
+build --cxxopt='-Wno-unknown-warning-option'
```
bgogul, line 27: Is it OK to turn this off? Seems like we might miss some valid problems?

markww: That's a possibility. On the other hand, this inhibits older versions of clang from compiling code written for newer versions of clang that have more warning options. I think this is OK to turn off because in order for it to hide a real bug, two things must be true:

1. The targeted warning option was really a misspelled warning, not just one that is unavailable in our compiler.
2. The targeted warning actually would have found a real issue in our code rather than an FP.

That strikes me as decreasing the utility of this warning significantly, and adds less graceful handling of code crafted with different versions of clang in mind.

bgogul (approved):

PR #749, opened 2022-09-30
Disable the `-Wmacro-redefined` warning.

markww:
Currently, the fuzztest library has an issue where two macros with colliding names are leaked from the headers of `absl` and `re2` into one of fuzztest's `.cc` files
( https://github.com/google/fuzztest/issues/41 ). Once this issue is fixed, we should re-enable this warning.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,6 +5,7 @@ try-import %workspace%/gcb/rbe/remote.bazelrc
 build --cxxopt='-std=c++17'
 build --cxxopt='-Werror' --cxxopt='-Wall'
 build --cxxopt='-Wno-deprecated-declarations'
+
 # Why are we doing this when Souffle-generated C++ clearly uses exceptions?
 # Well, Google famously does not like C++ exceptions in its internal codebase,
 # so we will need to explicitly override that in Raksha files everywhere we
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -26,6 +27,12 @@ build --host_cxxopt='-Wno-deprecated-declarations'
 # is our attempt to work around that.
 build --cxxopt='-Wno-unknown-warning-option'
 build --host_cxxopt='-Wno-unknown-warning-option'
+# The fuzztest library also has a conflict in one of its `.cc` files in
+# a macro. When https://github.com/google/fuzztest/issues/41 is fixed, we
+# should re-enable this warning.
+build --cxxopt='-Wno-macro-redefined'
+build --host_cxxopt='no-macro-redefined'
```
bgogul, line 34: typo `-Wno-macro-defined`

bgogul (approved):
LGTM modulo typo.

.bazelrc
markww: No, I think this is right. The warning is complaining about a macro being defined twice with the contents of the definition differing. From the build error:

```
external/com_google_absl/absl/base/internal/thread_annotations.h:220:9: error: 'NO_THREAD_SAFETY_ANALYSIS' macro redefined [-Werror,-Wmacro-redefined]
```

markww: Oh, but I do need a -W there, that's probably what you meant.

markww: Fixed, good catch, thanks!

PR #757, opened 2022-10-17
Deprecate relations to not declare

aferr:
In the very early days of Raksha when we were connecting the rust implementation of authorizaiton logic to a datalog backend of Arcs, we needed to skip some relation declarations as a hack to get this to work. We have implemented a similar feature in the C++ implementation of authorization logic, but we do not need this. Also, once we have integration tests for the C++ implementation of auth logic, skipping some of these declarations will now cause the compiler not to work. To that end, this PR deprecates that small bit of code.

No additional tests are added, but notably when this (dead) code is deleted, the existing test that run on PRs should pass.

### DIFF
--- a/src/ir/auth_logic/souffle_emitter.h
+++ b/src/ir/auth_logic/souffle_emitter.h
@@ -40,20 +40,6 @@ class SouffleEmitter {
                         std::move(outputs));
   }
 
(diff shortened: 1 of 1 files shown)


bgogul (approved):
Thank you for cleaning this up.

PR #758, opened 2022-10-17
Call universe relations pass from CPP logic bin

aferr:
This PR calls the universe relation populating pass from the binary that implements authorization logic in CPP. Other changes needed to get this working are also made.

There are no tests as of this PR, but:
 - we can start porting the integration tests from the rust implementation in followup PRs, and
 - this does not break existing tests


src/ir/auth_logic/BUILD
```diff
@@ -79,13 +79,20 @@ cc_test(
     ],
 )
 
+cc_library(
```
bgogul, line 82: why this change? 

[nit: not super particular as the other chagnes are small, but this could be a separate PR.]

src/backends/policy_engine/souffle/raksha_policy_datalog_emitter.cc
```diff
@@ -79,6 +80,7 @@ int main(int argc, char* argv[]) {
   // Invoke the parser for auth_logic and return datalog
   AuthLogicProgram program =
       raksha::ir::auth_logic::ParseProgram(policy_rules.value());
+  program = raksha::ir::auth_logic::InsertUniverseRelations(program);
```
bgogul, line 83: No tests for this change?

bgogul (commented):
Tests seem to be failing?

src/ir/auth_logic/BUILD
aferr: This is a part of this PR because if I don't make this change, but DO add the universe relation pass as a dependency for the policy_emitter, it causes the build to fail because the implementation of the function is linked twice.

OK this could be a separate PR. Based on your parenthetical, I'll assume you don't insist I split it, but if you do actually insist, say so.

aferr:
OK Now that I have a workaround to be able to build on one of my machines again, I can see that calling the universe pass breaks the old interfaces that the rust version of auth logic had at the time that we tied it into raksha.

In particular, with the universe relation pass, every relation mentioned in policies needs to be declared so that the pass can populate universes by using types. If any are left out, an error will be thrown. Declarations in the frontend language also result in declarations being emitted in the souffle backend, and unfortunately some of these would conflict with relations we already have declared in `taint.dl`. 

I think I no longer have enough context to know what parts are using the rust implementation of auth logic (and that can be deprecated) and what parts are using the CPP version.

I think we can solve this a few ways:
 - re-implement something like the "relations not to declare" in the CPP side of auth logic, and skip declaring the parts we know are built into the raksha backend
 - bake something into the 

It would be useful to discuss:
 - what parts we can deprecate instead of maintaining (if any)
 - how we want to properly fix interfacing relation declarations in auth logic and the raksha backend.

aferr:
Closing this one. Since there were other in-flight PRs that conflicted with this and I split this into another smaller PR, it was easier to start over.

PR #762, opened 2022-10-21
Add default to switch statement for universe pass

aferr:
This was a needed step to make Raksha buildable on my cloudtop.

### DIFF
--- a/src/ir/auth_logic/universe_relation_insertion.cc
+++ b/src/ir/auth_logic/universe_relation_insertion.cc
@@ -36,7 +36,7 @@ std::string TypeToUniverseName(const datalog::ArgumentType& arg_type) {
     case datalog::ArgumentType::Kind::kCustom:
       return absl::StrCat("is", arg_type.name());
     default:
-      return "";
+      CHECK(false) << "Unexpected variant of Value from IR";
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -35,6 +35,8 @@ std::string TypeToUniverseName(const datalog::ArgumentType& arg_type) {
       return "isPrincipal";
     case datalog::ArgumentType::Kind::kCustom:
       return absl::StrCat("is", arg_type.name());
+    default:
+      return "";
```
bgogul, line 39: Instead of `return ""`, shouldn't we do `CHECK(false)`? e.g.,

https://github.com/google-research/raksha/blob/56203028becf8fb196ea641cc8a663c059ee20b7/src/ir/proto_to_ir.cc#L76

aferr: fixed

src/ir/auth_logic/universe_relation_insertion.cc
```diff
@@ -35,6 +35,8 @@ std::string TypeToUniverseName(const datalog::ArgumentType& arg_type) {
       return "isPrincipal";
     case datalog::ArgumentType::Kind::kCustom:
       return absl::StrCat("is", arg_type.name());
+    default:
+      CHECK(false) << "Unexpected variant of Value from IR";
```
bgogul, line 39: Fix the error message to match the case here?

bgogul (approved):
LGTM, but error message should be fixed.

src/ir/auth_logic/universe_relation_insertion.cc
aferr: fixed

bgogul (approved):

PR #765, opened 2022-10-24
Wno-error=sign-compare 4 fuzztest on GLIBC>2.34-8

aferr:
This is part of a broader quest to make Raksha buildable on my cloudtop again

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -10,11 +10,6 @@ build --cxxopt='-Werror' --cxxopt='-Wall'
 build --features=external_include_paths
 build --cxxopt='-Wno-deprecated-declarations'
 
-# These are needed on systems with GLIBC > 2.34-8 to build google-fuzztest
-# which has these errors.
-build --cxxopt='-Wno-error=sign-compare'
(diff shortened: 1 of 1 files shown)


aferr:
The second commit should help make this work even if google-fuzztest folks don't agree with my issue or take my change:
https://github.com/google/fuzztest/issues/64
https://github.com/google/fuzztest/pull/65

aferr:
@bgogul Is this one OK ?

.bazelrc
```diff
@@ -10,6 +10,11 @@ build --cxxopt='-Werror' --cxxopt='-Wall'
 build --features=external_include_paths
 build --cxxopt='-Wno-deprecated-declarations'
 
+# These are needed on systems with GLIBC > 2.34-8 to build google-fuzztest
```
bgogul, line 13: You might want to move it below where there is already some exceptions related to fuzztest?

You will also need to add an equivalent `--host_cxxopt`

aferr: fixed

bgogul (approved):

PR #779, opened 2022-11-02
setup zetasql

harsha-mandadi:
Experiments with zetasql


bgogul:
I presume this is no longer needed?

harsha-mandadi:
Yes. We were able to resolve the issue by pointing to zetaSQL latest commit (instead of latest release) and downgrading re2 version. We now have zetaSQL integrated with Raksha without failures on github side. I am closing this as PR has served its purpose.

PR #781, opened 2022-11-03
    Call universe relation pass

aferr:
Call universe relation pass.
    
    This requires fixing a number of declarations and interfaces since:
     - the universe relation pass assumes that there are declarations
        (of non-universe relations) in the frontend of authorization logic
     - the universe relation pass creates declarations of universe relations
      as an output (and the pass that follows this one prepends these with
      "says")


src/backends/policy_engine/souffle/raksha_policy_datalog_emitter.cc
```diff
@@ -86,6 +87,7 @@ int main(int argc, char* argv[]) {
   // Invoke the parser for auth_logic and return datalog
   AuthLogicProgram program =
       raksha::ir::auth_logic::ParseProgram(policy_rules.value());
+  program = raksha::ir::auth_logic::InsertUniverseRelations(program);
```
aferr, line 90: This whole PR is just about this line. The test for this PR is that there are no regressions of the existing tests. The followup work that will test this is to port the regression tests that the rust implementation had.

aferr:
@bgogul Even though I am busy this week, it would help if you could review this by EOD Friday so that I can iterate when I am back on Monday.

bgogul:
> @bgogul Even though I am busy this week, it would help if you could review this by EOD Friday so that I can iterate when I am back on Monday.

Thank you for flagging, Andrew. Will take a look.

src/ir/auth_logic/souffle_emitter_test.cc
```diff
@@ -64,7 +64,6 @@ TEST(EmitterTestSuite, EmptyAuthLogicTest) {
 .type Number <: symbol
 .decl grounded_dummy(dummy_param : DummyType)
 .decl says_canActAs(speaker : Principal, p1 : Principal, p2 : Principal)
-.decl says_isNumber(speaker : Principal, x : Number)
```
bgogul, line 67: Why is this not there anymore?

bgogul (approved):

src/ir/auth_logic/souffle_emitter_test.cc
aferr: this is generated by the universe relation compiler pass now.

Cypher1:
Out of interest, what does this change do and for what purpose?

I do not have context enough to decipher the PR name.

aferr:
The point of this PR is actually just this one line change in
src/backends/policy_engine/souffle/raksha_policy_datalog_emitter.cc


  // Invoke the parser for auth_logic and return datalog
  AuthLogicProgram program =
      raksha::ir::auth_logic::ParseProgram(policy_rules.value());
+  program = raksha::ir::auth_logic::InsertUniverseRelations(program);
  DatalogProgram datalog_program =
      raksha::ir::auth_logic::LoweringToDatalogPass::Lower(program);
  std::string datalog_policy =

The rest of the PR is about making the tests compatible with this change.

This one line change uses the `InsertUniverseRelations` pass to transform
the program from the AST (which is generated from the parser) into a form
that has: 1) declarations of the universe relations for all types used in
the program (this list of declarations comes from the relation
declarations) scoped to a speaker, 2) transformations of the program  to
populate these relations where constants are used, 3) transformations of
the program to check these relations where non-constants are used (and
where stratification might not be possible otherwise).

A longer comment describes the pass here
<https://github.com/google-research/raksha/blob/main/src/ir/auth_logic/universe_relation_insertion.h>
and
the cc file implements it.

Cypher1:
Thanks @aferr I really appreciate your explanation.

A suggestion: I think a short summary of your explanation would greatly enhance your commit messages.
e.g. Something like

"Invoke conversion of AuthLogic program to Universe Relations for datalog"

PR #788, opened 2022-11-04
Show ZetaSQL in Raksha distinguishing between valid and invalid SQL

markww:


.bazelrc
```diff
@@ -42,11 +42,16 @@ build --host_cxxopt='-Wno-unknown-warning-option'
 # should re-enable this warning.
 build --cxxopt='-Wno-macro-redefined'
 build --host_cxxopt='-Wno-macro-redefined'
+build --cxxopt='-Wno-error'
```
harsha-mandadi, line 45: This was temporary fix I added to avoid Werrors. Should I make this specific to each warning or keep this as is and remove other changes here?

markww: I'd say leave it alone. I mostly put this PR up to have the work visible and passing on a machine other than my own. The delicate dance between Google's internal repo and GitHub will mean I'll probably have to carve this PR up into a few different, smaller changes to actually land it.

harsha-mandadi: ok sounds good.

harsha-mandadi: ok. Sounds good.

markww:
This PR was refactored into #790 and a few other warning-disabling PRs such as #801 and #802. This can now be closed.

PR #789, opened 2022-11-04
Import ZetaSQL into Raksha's `WORKSPACE`.

markww:
This commit imports ZetaSQL into Raksha's `WORKSPACE`, laying the groundwork for it to be used as a SQL frontend. This also required downgrading re2 to a previous version to keep in sync with ZetaSQL.


markww:
See #788 to see a more complete picture of what using this dependency will look like.

harsha-mandadi (approved):
LGTM.

PR #794, opened 2022-11-04
Remove -Wno-deprecated-declarations

bgogul:
Now that we don't need to import glog, we can enable `deprecated-declarations` warning.


markww (approved):
LGTM

PR #798, opened 2022-11-05
Turn off a warning about int to float conversion in host config.

markww:
ZetaSQL's ICU dependency has an implicit float to int conversion, which triggers a warning. This is definitely fixed in a later version of icu, so we should be able to remove this when ZetaSQL gets around to upgrading.


harsha-mandadi (approved):

markww:
I'm closing this because it is insufficient, on its own, to get ZetaSQL compiling with Raksha. I'm going to make a bigger PR turning off many more warning flags.

PR #801, opened 2022-11-07
Turn off warnings to get ZetaSQL to build.

markww:


.bazelrc
```diff
@@ -44,7 +49,27 @@ build --cxxopt='-Wno-final-dtor-non-final-class'
...
+build --cxxopt='-Wno-enum-compare-switch'
+build --host_cxxopt='-Wno-enum-compare-switch'
+build --cxxopt='-Wno-char-subscripts'
+build --host_cxxopt='-Wno-char-subscripts'
+# This warning has a false positive on ZetaSQL.
+build --cxxopt='-Wno-thread-safety-precise'
+build --host_cxxopt='-Wno-thread-safety-precise'
 
+# Need this for ZetaSQL's icu dependency as long as it remains at version 65.
+# Should remove when it gets upgraded.
```
bgogul, line 70: File a bug?

.bazelrc
```diff
@@ -44,7 +49,27 @@ build --cxxopt='-Wno-final-dtor-non-final-class'
 build --host_cxxopt='-Wno-final-dtor-non-final-class'
 build --cxxopt='-Wno-unused-const-variable'
 build --host_cxxopt='-Wno-unused-const-variable'
+build --cxxopt='-Wno-unused-variable'
+build --host_cxxopt='-Wno-unused-variable'
+build --cxxopt='-Wno-unused-function'
+build --host_cxxopt='-Wno-unused-function'
```
bgogul, line 55: Are these for ZetaSQL as well? If so, move it after the comment on line 56?

bgogul (approved):

.bazelrc
markww: There are two sections here: warnings we are turning off for ZetaSQL for which I have proposed fixes commits to the ZetaSQL folks that should address the warnings, and warnings that we are turning off for ZetaSQL where I don't know if a fix will be published soon. These are in the former category, and there's already a header addressing them.

.bazelrc
markww: https://github.com/google/zetasql/issues/122 .

PR #804, opened 2022-11-08
uncomment src/analysis/souffle/examples/ in gcb

harsha-mandadi:


markww:
Looks like the presubmit check may need to be updated. Did the `examples` in Souffle get removed or moved?

PR #812, opened 2022-11-16
[DEMO ONLY DO NOT CHECK IN] How to get ZetaSQL AST Node from string

aferr:
Quick demo of how to get Zeta AST node from a string using their public APIs.


aferr:
@bgogul @markww FYI I have figured out at least this much

Do not do code review on this, it should not be checked in :)

PR #816, opened 2022-11-18
Make the ZetaSQL frontend grab only analyzer dependencies.

markww:


WORKSPACE
```diff
@@ -307,7 +307,15 @@ zetasql_deps_step_1()
 
 load("@com_google_zetasql//bazel:zetasql_deps_step_2.bzl", "zetasql_deps_step_2")
 
-zetasql_deps_step_2()
+# To use ZetaSQL as a frontend, we only need the analyzer. Don't grab other
+# dependencies unnecessarily.
+zetasql_deps_step_2(
```
harsha-mandadi, line 312: Are we still using zetaSQL?

markww: For the work we're doing immediately, no. We expect, however, that we will use it in the future. So long as this is going on ice for a bit, we might as well not build unnecessary dependencies every time.

bgogul (approved):

PR #828, opened 2022-11-30
remove arcs directory

harsha-mandadi:


src/ir/proto/raksha_ir.proto
```diff
@@ -126,3 +126,163 @@ enum AnalysisResultStatus {
 message AnalysisResult {
   AnalysisResultStatus result = 1;
 }
+
+message TypeProto {
```
bgogul, line 130: Can we split this off to a separate proto file  (e.g., type.proto) instead of adding this to raksha_ir.proto?

src/ir/proto/raksha_ir.proto
```diff
@@ -126,3 +126,163 @@ enum AnalysisResultStatus {
...
+// An Annotation parameter.
+message AnnotationParamProto {
+  string name = 1;
+  oneof value {
+    string str_value = 2;
+    int32 num_value = 3;
+    bool bool_value = 4;
+  }
+}
+
```
bgogul, line 288: Remove this as it is not needed.

src/analysis/souffle/tests/arcs_manifest_tests_todo/BUILD
```diff
@@ -14,8 +14,6 @@
 # limitations under the License.
 #
 #-------------------------------------------------------------------------------
-load("//build_defs:arcs.bzl", "arcs_manifest_proto")
```
markww, line 17: This whole `arcs_manifest_tests_todo` directory can be deleted. It contains arcs manifests that we eventually wanted to turn into tests. But since the arcs frontend is no longer being supported, there's no reason to keep them around.

src/ir/proto/raksha_type.proto
```diff
@@ -0,0 +1,100 @@
...
+
+// This proto file contains the proto representation of the Raksha IR. This
+// allows frontends to construct the Raksha IR in any language and send the IR
+// over the wire or across an FFI layer for analysis.
+
+syntax = "proto3";
+
+package raksha.type.proto;
+
+message TypeProto {
```
markww, line 25: I'm not sure if we should even keep around this `TypeProto` file. While we are making some minimal use of encoding and decoding these protos in Raksha types, we aren't actually using types for anything in Raksha at the moment. It might be best to just have a clean break from the Arcs conception of types and build our own types as we need them. Otherwise, we risk sleepwalking into following the Arcs path, which might not always work for us.

For instance, does it make sense to think of a `BIGINT` as a primitive type? Or a `DURATION`? This is one of those things where the answer is "maybe, maybe not". In both the Souffle and potential custom solver case, the only primitive types (in this case meaning those that map directly to a simple type in the underlying runtime) would be `string`, `int64_t`, `uint64_t`, and `double`.

src/ir/types/proto/entity_type.h
```diff
@@ -16,21 +16,24 @@
...
 #define SRC_IR_TYPES_PROTO_ENTITY_TYPE_H_
 
+#include "src/ir/proto/raksha_type.pb.h"
 #include "src/ir/types/entity_type.h"
 #include "src/ir/types/type_factory.h"
-#include "third_party/arcs/proto/manifest.pb.h"
 
 namespace raksha::ir::types::proto {
 
+using type::proto::EntityTypeProto;
```
markww, line 25: I think we might want to get rid of `EntityType`s and `Schema`s altogether. We aren't using them in any significant way, and we might want to build something different for Raksha's purposes.

bgogul: Let us keep them around for now. This would become useful/needed when we start talking about SQL tables and their schemas.

src/ir/proto/raksha_type.proto
bgogul: I think it might be a good idea to simplify the proto to just the things we use currently.

bgogul (commented):
Marked some parts of the proto for removal. PTAL.

src/ir/types/proto/entity_type.h
harsha-mandadi: Acknowledge

bgogul (approved):
Thank you!

markww (approved):
LGTM modulo getting rid of the last node dependencies.

markww:
Also, be sure to set `ready to pull` on this to verify that there's no internal footprint for this change before merging. This has the potential to be really painful to undo, so best to be careful.

gcb/presubmit.sh
```diff
@@ -94,13 +94,8 @@ check_build_files_for_licenses_rule
...
 
 # Verifies that the following targets build fine:
-#  - Arcs parser and proto works.
-#  - Arcs manifest tests that we do not yet handle parse correctly.
 #  - Arcs manifest examples.
 bazel_run build \
-  //third_party/arcs/examples:consume \
-  //third_party/arcs/proto:manifest_cc_proto \
-  //src/analysis/souffle/tests/arcs_manifest_tests_todo/... \
   //src/analysis/souffle/examples/...
```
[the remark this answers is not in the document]
harsha-mandadi: //src/analysis/souffle/examples/... are not dependent on arcs in any way. We removed the rust/arcs dependency and use raksha_policy_verifier instead. Why do you think they should go?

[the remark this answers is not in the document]
markww: OK, if they don't depend on Arcs, we should keep them.

PR #831, opened 2022-11-30
remove rust dir

harsha-mandadi:


bgogul:
> @bgogul @harsha-mandadi Do we still want to re-create the old rust integration tests in the C++ implementation before we delete the rust directory? The universe relation pass is implemented in C++, so it should be possible to do this now.

It would be ideal to port the tests if we haven't already done so.

bgogul:
> > @bgogul @harsha-mandadi Do we still want to re-create the old rust integration tests in the C++ implementation before we delete the rust directory? The universe relation pass is implemented in C++, so it should be possible to do this now.
> 
> It would be ideal to port the tests if we haven't already done so.

What does it involve? Is it simple as moving the test files to a c++ directory and invoking a script?

harsha-mandadi:
Oh sorry. For some reason I thought the integration tests are already imported to c++ 😅. I will pause on this till that is done.

aferr:
> What does it involve? Is it simple as moving the test files to a c++ directory and invoking a script?

Most of the tests work by running the compiler on some input authorization logic program and then checking against some expected values of the queries. Some of the tests run on an input program and then have assertions that are not necessarily checking queries. Porting the tests involves:
 - Directly copying the directory with the input authorization logic programs to a location that makes sense for the C++ version: https://github.com/google-research/raksha/tree/main/rust/tools/authorization-logic/test_inputs
 - Porting the code that runs these tests from Rust to C++. The rust integration test code lives here: https://github.com/google-research/raksha/tree/main/rust/tools/authorization-logic/src/test
     - A good first file to look at there is https://github.com/google-research/raksha/blob/main/rust/tools/authorization-logic/src/test/test_queries.rs because several of the tests depend on datastructures and functions there.

> Oh sorry. For some reason I thought the integration tests are already imported to c++ 😅. I will pause on this till that is done.

Just to be clear, my understanding is that this task is on @harsha-mandadi 's plate as a part of deprecating the rust implementation unless someone tells me otherwise (so I won't personally work on this unless someone tells me I should do that :)

harsha-mandadi:
oh I thought it was on your plate Andrew. In one of the team standup meetings, you mentioned you would like to work on the tests as you were looking for something to work on. Did I misunderstood what you meant. I can work on it, no worries!

bgogul (approved):
It looks like the rust tests are being migrated

PR #859, opened 2023-01-06
Add an Antlr grammar for SQLite to dependencies and build.

markww:
This commit adds an Antlr grammar describing the SQLite SQL dialect to Raksha's `WORKSPACE` dependencies. This opens us up to use this to build a SQLite frontend for Raksha in the future.


bgogul (approved):

harsha-mandadi (approved):

aferr:
Oh I actually already wrote this same code essentially but didn't check it in because I wanted to get the grammar in first... darn wasted / duplicated work.

